PR #21, opened 2021-09-03
Loading Screen + Tests

### LINKED ISSUE #18: Loading Screen Scaffold and navigation
Create a loading screen that serves as a buffer to load the keychain and figure out the actual first screen of the Application: (Welcome and Onboarding or Home ScreenA)

pacu:
closes #18 

image

### DIFF
--- a/secant/App Errors/AppError.swift
+++ b/secant/App Errors/AppError.swift
@@ -1,0 +1,12 @@
+//
+//  AppError.swift
+//  secant-testnet
+//
+//  Created by Francisco Gindre on 9/3/21.
+//
+
+import Foundation
+
+enum AppError: Error {
+    case failedToInitialize(underlyingError: Error)
+}
--- a/secant/App Errors/AppError.swift
+++ b/secant/App Errors/AppError.swift
@@ -8,5 +8,5 @@
 import Foundation
 
 enum AppError: Error {
-    case failedToInitialize(underlyingError: Error)
+    case failedToInitialize(Error)
 }
--- a/secant/Mocked Dependencies/KeyStoring.swift
+++ b/secant/Mocked Dependencies/KeyStoring.swift
@@ -12,7 +12,7 @@ protocol KeyStoring {
     func exportBirthday() throws -> BlockHeight
     func importPhrase(bip39 phrase: String) throws
     func exportPhrase() throws -> String
-
+    func areKeysPresent() throws -> Bool
     /**
     Use carefully: Deletes the seed phrase from the keychain
     */
@@ -34,4 +34,5 @@ protocol KeyStoring {
 enum KeyStoringError: Error {
     case alreadyImported
     case uninitializedWallet
+    case storageError(underlyingError: Error)
 }
--- a/secant/Mocked Dependencies/KeyStoring.swift
+++ b/secant/Mocked Dependencies/KeyStoring.swift
@@ -34,5 +34,5 @@ protocol KeyStoring {
 enum KeyStoringError: Error {
     case alreadyImported
     case uninitializedWallet
-    case storageError(underlyingError: Error)
+    case storageError(Error)
 }
--- a/secant/Routers/AppRouter.swift
+++ b/secant/Routers/AppRouter.swift
@@ -12,6 +12,7 @@ enum AppRouterScreen {
     case appLoading
     case createRestoreWallet
     case home
+    case loadingFailed
 }
 
 class AppRouter: Router {
@@ -39,7 +40,11 @@ class AppRouter: Router {
     }
 
     @ViewBuilder func loadingScreen() -> some View {
-        Text("Loading")
+        LoadingScreen(router: self, viewModel: LoadingScreenViewModel(services: self.services))
+    }
+    
+    @ViewBuilder func loadingFailedScreen() -> some View {
+        Text("loading failed")
     }
 }
 
@@ -48,22 +53,29 @@ struct AppRouterView: View {
 
     var body: some View {
         viewForScreen(router.screen)
-            .onAppear {
-                DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
-                    if router.services.keyStorage.keysPresent {
-                        router.screen = .home
-                    } else {
-                        router.screen = .createRestoreWallet
-                    }
-                }
-            }
     }
-
+    
     @ViewBuilder func viewForScreen(_ screen: AppRouterScreen) -> some View {
         switch router.screen {
         case .appLoading:           router.loadingScreen()
         case .createRestoreWallet:  router.createNew()
         case .home:                 router.home()
+        case .loadingFailed:        router.loadingFailedScreen()
         }
+    }
+}
+
+extension AppRouter: LoadingScreenRouter {
+    func proceedToWelcome() {
+        self.screen = .createRestoreWallet
+    }
+
+    func proceedToHome() {
+        self.screen = .home
+    }
+    
+    // TODO: handle Errors
+    func failWithError() {
+        self.screen = .loadingFailed
     }
 }
--- a/secant/Routers/AppRouter.swift
+++ b/secant/Routers/AppRouter.swift
@@ -40,7 +40,10 @@ class AppRouter: Router {
     }
 
     @ViewBuilder func loadingScreen() -> some View {
-        LoadingScreen(router: self, viewModel: LoadingScreenViewModel(services: self.services))
+        LoadingScreen(
+            viewModel: LoadingScreenViewModel(services: self.services),
+            router: self
+        )
     }
     
     @ViewBuilder func loadingFailedScreen() -> some View {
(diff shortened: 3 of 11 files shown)


secant/App Errors/AppError.swift
```diff
@@ -0,0 +1,12 @@
...
+//  AppError.swift
+//  secant-testnet
+//
+//  Created by Francisco Gindre on 9/3/21.
+//
+
+import Foundation
+
+enum AppError: Error {
+    case failedToInitialize(underlyingError: Error)
```
adamstener, line 11: Is there a benefit here to naming the associated value?

secant/Mocked Dependencies/KeyStoring.swift
```diff
@@ -34,4 +34,5 @@ protocol KeyStoring {
 enum KeyStoringError: Error {
     case alreadyImported
     case uninitializedWallet
+    case storageError(underlyingError: Error)
```
adamstener, line 37: Same question here. Since there is only a single value and it's an error, are we providing any benefit by adding an argument label?

secant/Routers/AppRouter.swift
```diff
@@ -48,22 +53,29 @@ struct AppRouterView: View {
...
-                    if router.services.keyStorage.keysPresent {
-                        router.screen = .home
-                    } else {
-                        router.screen = .createRestoreWallet
-                    }
-                }
-            }
     }
-
+
```
adamstener, line 57: Looks like an incidental tab of white space made it in here

secant/Screens/Loading/LoadingScreen.swift
```diff
@@ -0,0 +1,58 @@
...
+protocol LoadingScreenRouter: AnyObject {
+    func proceedToHome()
+    func failWithError()
+    func proceedToWelcome()
+}
+
+struct LoadingScreen: View {
+    @State var router: LoadingScreenRouter?
+    
+    @StateObject var viewModel: LoadingScreenViewModel
```
adamstener, line 19: What do you think has higher precedence, State or StateObject?  Im inclined to say StateObject because it's meant to be the single source of truth for that object through all of it's ancestors if I understand correctly.  Which if true, would mean that @StateObject belongs on top of @State variables.  I think that's how the template is setup now too.

secant/Screens/Loading/LoadingScreen.swift
```diff
@@ -0,0 +1,58 @@
...
+    
+    var body: some View {
+        Text("Loading")
+            .onReceive(viewModel.$loadingResult, perform: { result in
```
adamstener, line 23: What do you think of this formatting to keep things cleanly indented and easy to read?  This is in line with what I've fixed in the Swfitlint PR.

```
.onReceive(
    viewModel.$loadingResult,
    perform: { result in
        guard
            let loadingResult = result,
            let router = self.router
        else { return }

        viewModel.callRouter(router, with: loadingResult)
     }
)
.onAppear {
    DispatchQueue.main.asyncAfter(deadline: .now() + 4) {
        viewModel.loadAsync()
    }
}
```

secant/Stubs/MockServices.swift
```diff
@@ -65,7 +57,69 @@ class MockMnemonicPhraseHandling: MnemonicSeedPhraseHandling {
     func isValid(mnemonic: String) throws {}
 }
 
+class KeysPresentStub: KeyStoring {
+    init(returnBlock: @escaping () throws -> Bool) {
+        self.returnBlock = returnBlock
+    }
+    var returnBlock: () throws -> Bool
```
adamstener, line 64: Do you think we can reorganize this class to follow the Type structure that we put in the StyleGuide?

secant/Stubs/MockServices.swift
```diff
@@ -65,7 +57,69 @@ class MockMnemonicPhraseHandling: MnemonicSeedPhraseHandling {
...
+        self.birthday = nil
+    }
+    
+    func nukeWallet() {
+        nukePhrase()
+        nukeBirthday()
+    }
+}
 class MockKeyStoring: KeyStoring {
+    func areKeysPresent() throws -> Bool {
```
adamstener, line 119: This function should come after the properties.  The keysPresent computed property could be moved up too to come just after these properties here.

secantTests/LoadingScreenTests.swift
```diff
@@ -0,0 +1,239 @@
...
+                case .none:
+                    XCTFail("found None when expected a value")
+                }
+            }
+            .store(in: &cancellables)
+        loadingViewModel.loadAsync()
+        wait(for: [testExpectation], timeout: 0.1)
+    }
+
+    func testLoadingPublishesNavigatesToOnboardingWhenNoCredentialsFound() throws {
```
adamstener, line 48: This test reads to me like we're waiting for the view model to publish newWallet, which may end up having something to do with navigating but should the test description just call out what we are asserting?  This applies to most of these tests here I believe.

secantTests/LoadingScreenTests.swift
```diff
@@ -0,0 +1,239 @@
...
+@testable import secant_testnet
+import Foundation
+import Combine
+class LoadingScreenTests: XCTestCase {
+    var cancellables: [AnyCancellable] = []
+    
+    // MARK: LoadingScreenViewModel Tests
+    
+    func testLoadingPublishedHomeScreenWhenCredentialsArePresent() throws {
+        let mockServices = MockServices()
```
adamstener, line 18: It looks like we go through alot of the same boilerplate setup code here.  Would it be good to factor this out into a helper that just takes the returnBlock as input so you don't have to write this same code for every test?

adamstener (changes requested):

secant/Mocked Dependencies/KeyStoring.swift
pacu: good point

secant/Screens/Loading/LoadingScreen.swift
pacu: Yes this is much better

secantTests/LoadingScreenTests.swift
pacu: Thanks for all the suggestions!

I don't think I'm quite following you here. Probably due to the code highlighting.

adamstener: It looks like every test runs though this code:
```
 let mockServices = MockServices()
         let stub = KeysPresentStub(returnBlock: {
             throw KeyStoringError.alreadyImported
         })
         mockServices.keyStorage = stub
         let loadingViewModel = LoadingScreenViewModel(services: mockServices)
          let testExpectation = XCTestExpectation(description: "LoadingViewModel Publishes .credentialsFound when credentials are present and there's no failure")
         let expected = LoadingScreenViewModel.LoadingResult.credentialsFound
 ```
I was just wondering if it would be worth it to move this to a setup function so you don't have to repeat yourself so much is all.  There are only very little changes between the different tests, but this is just something to consider so that as you add more tests you don't need to keep adding the same boilerplate in the future.

File: secant/Stubs/MockServices.swift (hunk 6 above)
Comment (adamstener) [reply secant/Stubs/MockServices.swift:64]: I saw this was marked as resolved but I do think even stub files should follow our normal file structure guidelines.  Mixing the organization of inits, functions and properties only makes the code harder to reason about.  I see the point of having some local reasoning with variables near their functions, but overall the same standard file structure with well named variables and functions will be the easiest thing for someone else to read and understand.

File: secantTests/LoadingScreenTests.swift [hunk 10]
```diff
@@ -0,0 +1,239 @@
...
+        let spyRouter = LoadingScreenRouterSpy(fulfillment: {
+        })
+        
+        loadingViewModel.callRouter(spyRouter, with: loadingViewModel.load())
+        XCTAssertTrue(spyRouter.failWithErrorCalled)
+    }
+}
+
+class LoadingScreenRouterSpy: LoadingScreenRouter {
+    init(fulfillment: @escaping () -> Void) {
```
adamstener, line 216: init functions should come after the property declarations for the structure we've defined for classes.

adamstener (commented):
This looks really close, just a few questions or comments left!

secantTests/LoadingScreenTests.swift
pacu: I'm going to make a post it with our own rules LOL

secantTests/LoadingScreenTests.swift
pacu: The problem of extracting this is that the stub is lost and we would need to return a tuble with the stub with the spy boolean and the view model. I've tried it and I don't see much added value on the reusability.

adamstener: Sounds good!

pacu:
````
enum LoadingScreenViewModelHelper {
    static func loadingViewModelWith(keysPresentStubBlock: @escaping () throws -> Bool) -> LoadingScreenViewModel {
        let mockServices = MockServices()
        let stub = KeysPresentStub(returnBlock: keysPresentStubBlock)
        mockServices.keyStorage = stub
        return LoadingScreenViewModel(services: mockServices)
    }
}
````

This can be used when the Spy is not needed

PR #113, opened 2021-11-04
Onboarding: Skip Functionality

adamstener:
This PR adds the Skip capability to Onboarding.

The user can go forward and backward through the onboarding steps but can at any time press Skip, which puts the user to the end of onboarding.  When you are on the last screen, there is still a back button.  The behavior of this PR makes it so that the previously known onboarding step is preserved if someone presses skip, such that if you are on the first page and skip to the end, but press Back from there, the user is landed back on the first page.  The assumption here is that pressing the skip button may be incidental, and the user should be taken back to their last known position.

This PR includes the state changes in the SMA store, updates to the current onboarding example/testing screen, and unit tests.  It also updates the onboarding step data to reflect the current verbiage in the design better.

Skip

This code review checklist is intended to serve as a starting point for the author and reviewer. However, it may not be appropriate for all types of changes (e.g., fixing a spelling typo in the documentation).  For a more in-depth discussion of how we think about code review, please see [Code Review Guidelines](../blob/main/CODE_REVIEW_GUIDELINES.md).

# Author

- [x] Rebase and squash: Did you pull in the latest changes from the main branch and squash your commits before assigning a reviewer? Having your code up to date and squashed will make it easier for others to review. Use your best judgment when squashing commits, as some changes (such as refactoring) might be easier to review as a separate commit.

# Reviewer

- [ ] Ad hoc review: Did you perform an ad hoc review?  _In addition to a first pass using the code review guidelines, do a second pass using your best judgment and experience, which may identify additional questions or comments. Research shows that code review is most effective when done in multiple passes, where reviewers look for different things through each pass._
- [ ] How is Code Coverage affected by this PR? _We encourage you to compare coverage before and after your changes and, when possible, leave it in a better place. [Learn More...](../blob/main/docs/testing/local_coverage.md)_
- [ ] Run the app: Did you run the app and try the changes? While the CI server runs the app to look for build failures or crashes, humans are more likely to notice unexpected log messages, UI inconsistencies, or bad output data.

### DIFF
--- a/secant/Features/Onboarding/OnboardingStore.swift
+++ b/secant/Features/Onboarding/OnboardingStore.swift
@@ -11,38 +11,54 @@ import ComposableArchitecture
 
 struct OnboardingStep: Equatable, Identifiable {
     let id: UUID
+    let title: String
     let description: String
-    let imageName: String
 }
 
 struct OnboardingState: Equatable {
     var steps: IdentifiedArrayOf<OnboardingStep> = Self.onboardingSteps
     var index = 0
     var offset: CGFloat = .zero
+    var skippedAtindex: Int?
     
     var currentStep: OnboardingStep { steps[index] }
-    var nextButtonDisabled: Bool { steps.count == index + 1 }
+    var skipButtonDisabled: Bool { steps.count == index + 1 }
     var backButtonDisabled: Bool { index == 0 }
-    var progress: Int {
-        ((index + 1) * 100) / (steps.count)
-    }
+    var progress: Int { ((index + 1) * 100) / (steps.count) }
 }
 
 enum OnboardingAction: Equatable {
-    case nextPressed
-    case backPressed
+    case next
+    case back
+    case skip
+    case createNewWallet
 }
 
 let onboardingReducer = Reducer<OnboardingState, OnboardingAction, Void> { state, action, _ in
     switch action {
-    case .backPressed:
-        state.index -= 1
-        state.offset += 20.0
+    case .back:
+        guard state.index > 0 else { return .none }
+        if let skippedFrom = state.skippedAtindex {
+            state.index = skippedFrom
+            state.skippedAtindex = nil
+        } else {
+            state.index -= 1
+            state.offset += 20.0
+        }
         return .none
         
-    case .nextPressed:
+    case .next:
+        guard state.index < state.steps.count - 1 else { return .none }
         state.index += 1
         state.offset -= 20.0
+        return .none
+        
+    case .skip:
+        state.skippedAtindex = state.index
+        state.index = state.steps.count - 1
+        return .none
+        
+    case .createNewWallet:
         return .none
     }
 }
--- a/secant/Features/Onboarding/Views/Onboarding.swift
+++ b/secant/Features/Onboarding/Views/Onboarding.swift
@@ -15,26 +15,20 @@ struct OnboardingView: View {
         WithViewStore(self.store) { viewStore in
             VStack(spacing: 50) {
                 HStack(spacing: 50) {
-                    Button(
-                        action: { viewStore.send(.backPressed) },
-                        label: { Text("Previous") }
-                    )
-                    .disabled(viewStore.backButtonDisabled)
+                    Button("Back") { viewStore.send(.back) }
+                        .disabled(viewStore.backButtonDisabled)
                     
                     Spacer()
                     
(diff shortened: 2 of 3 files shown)


secant/Features/Onboarding/OnboardingStore.swift
```diff
@@ -11,38 +11,54 @@ import ComposableArchitecture
...
+    case createNewWallet
 }
 
 let onboardingReducer = Reducer<OnboardingState, OnboardingAction, Void> { state, action, _ in
     switch action {
-    case .backPressed:
-        state.index -= 1
-        state.offset += 20.0
+    case .back:
+        guard state.index > 0 else { return .none }
```
pacu, line 40: This is more of a question / reflection than a review.
I've found that code like this in the reducer could be a source of bugs due to side-effects introduced by changes to the State. 

Should the reducer be unit tested for these scenarios, meaning that we code a given/when/then kind of test that performs this action on a reducer and we assert the expected changes? 
Or should instead the reducer be straight forward and calling functions on the store that then are unit tested separately?

pacu (approved):

secant/Features/Onboarding/OnboardingStore.swift
dh-ecc: [discussion]

> Should the reducer be unit tested for these scenarios, meaning that we code a given/when/then kind of test that performs this action on a reducer and we assert the expected changes?
– @pacu 

Good question, I'd say this boils down to readability, which is also to say, that there is no difference in _testability_ between the two. Testing `store.send(.back)` is equivalent to testing `func goBack(OnboardingState) -> (OnboardingAction, Effect)` (or technically `func goBack(&OnboardingState) -> Effect`) where `goBack` is the code in lines 40-48.

I think it's certainly useful to extract out functions to be tested individually if they represent some functionality that is shared, though we don't _lose_ any testability by _not_ doing that.

The readability I talk about above is in relation to the source code... but the readability of the tests is also another factor and I think that might be where your given/when/then question comes from so if we have something constructed like below:

```swift
func testSkipOnboarding() {
  // GIVEN (index = 1) and expected (index = 0)
  var state = OnboardingState(
                  index: 1,
                  ...
              )
  var expectedState = OnboardingState(
                  index: 0,
                  ...
              )
  
  // WHEN (back button pressed)
  let _ = goBack(state)

  // THEN (index = 0 ... and other assertions)
  XCTAssertEqual(state.index, 0)
}

```

This can be exactly equivalent with the TCA test framework as below (this code is mostly copied and pasted from a test @adamstener has currently in the source code, with added given/when/then comments)

```swift
func testSkipOnboarding() {
  // GIVEN (index = 1)
  let initialIndex = 1
  let initialOffset: CGFloat = .zero - 20.0
  let store = TestStore(
    initialState: OnboardingState(
    index: initialIndex,
    offset: initialOffset
    ),
    reducer: onboardingReducer,
    environment: ()
  )
  
  // WHEN (back button pressed)
  store.send(.back) {
    // THEN (index = 0, or as below... 1 less than previous)
    $0.index -= 1
    ...
  }
}
```

The `// THEN` part is definitely a bit confusing at first glance! this is a TCA test framework thing, but it becomes very helpful very quickly.

The idea is that you set the `$0` argument in that closure to the _expected_ result of sending the `.back` action. This allows you to check not only the expected changes, but that nothing _else_ changed! without having to explicitly check each field... for example if you change this "goBack" to add `state.skippedAtIndex = 2`  and run the tests again, the test framework will fail and show you (in a super readable way) that this was changed, but you did not specify that it should be 😊.

secant/Features/Onboarding/OnboardingStore.swift
```diff
@@ -11,38 +11,54 @@ import ComposableArchitecture
...
     let id: UUID
+    let title: String
     let description: String
-    let imageName: String
 }
 
 struct OnboardingState: Equatable {
     var steps: IdentifiedArrayOf<OnboardingStep> = Self.onboardingSteps
     var index = 0
     var offset: CGFloat = .zero
```
dh-ecc, line 21: [minor]

[from offline chat]
Would it make sense to derive this offset from the `index` and `steps.count`? (and possibly screen dimensions?)

dh-ecc (commented):
LGTM 😊 appreciate this already merged 😆, thought I'd add some suggestions / comments for reference for any future stuff 👍

Issue #12, opened 2021-08-26
Sending Done / Failed Scaffold Navigation Placeholder


pacu:
Contains a Screen that is shown as a placeholder for the real contents for navigational purposes. 
This screen lets developers test that the navigation proposed can be properly implemented and that the user can transition between the screens that precede and follow this one. 
 
Contains mocked elements that demonstrate the nature of the screen and UI elements that allow the user to perform navigation actions

PR #121, opened 2021-11-21
Tca nav add send

dh-ecc:
Putting this up to see if we have a preference on how to go about handling the construction of connected views. (Though any other comments and suggestions are also welcome.)

The main diff is 780dd0f which goes from the original implementation where the "next" view is encoded in the view itself, to one where the connection between the views is handled by a "Parent" view... or something that acts much like a "coordinator".

There is no "correct" answer here, and it's possible that both are equally good in _this_ example, but one might shine over the other in different set ups.

I'll throw out my thoughts here is that I probably slightly prefer the "coordinator" style one, and something like this would be _necessary_ if we intended to use these views individually / in different combinations. However a big drawback is that for nested navigationLinks, you can see that this can start to make the indentation go further and further down.

I had two ideas on "fixing this". One is just creating intermediate computed views that would only go one navigationLink deep. Two is, I _think_ this can be solved by using a `ViewModifier` and doing the nesting in here when we see that the view we are modifying is already being modified by a "navigation modifier". This would allow us to "stack" the modifiers at the same level of indentation. I haven't got an implementation for this yet, it's just a thought 😅.

Please have a look at the [Commits](https://github.com/zcash/secant-ios-wallet/pull/121/commits) for some more context on all the other commits.

(This should address #12  and #13)

---

# Author

- [x] Documentation: Did you update Docs as appropiate? (E.g [README.md](../blob/main/README.md), etc.)

# Reviewer

### DIFF
--- a/secant/Features/Send/Views/SendView.swift
+++ b/secant/Features/Send/Views/SendView.swift
@@ -1,0 +1,58 @@
+import SwiftUI
+import ComposableArchitecture
+
+struct SendView: View {
+    enum Route: Equatable {
+        case showApprove
+        case showSent
+        case done
+    }
+
+    let store: Store<SendState, SendAction>
+
+    var body: some View {
+        WithViewStore(store) { viewStore in
+            Create(
+                transaction: viewStore.bindingForTransaction,
+                isComplete: viewStore.bindingForApprove
+            )
+            .navigationLinkEmpty(
+                isActive: viewStore.bindingForApprove,
+                destination: {
+                    Approve(
+                        transaction: viewStore.transaction,
+                        isComplete: viewStore.bindingForSent
+                    )
+                    .navigationLinkEmpty(
+                        isActive: viewStore.bindingForSent,
+                        destination: {
+                            Sent(
+                                transaction: viewStore.transaction,
+                                isComplete: viewStore.bindingForDone
+                            )
+                        }
+                    )
+                }
+            )
+        }
+    }
+}
+
+struct SendView_Previews: PreviewProvider {
+    static var previews: some View {
+        NavigationView {
+            SendView(
+                store: .init(
+                    initialState: .init(
+                        transaction: .demo,
+                        route: nil
+                    ),
+                    reducer: .default,
+                    environment: ()
+                )
+            )
+            .navigationBarTitleDisplayMode(.inline)
+            .navigationViewStyle(StackNavigationViewStyle())
+        }
+    }
+}
--- a/secant.xcodeproj/project.pbxproj
+++ b/secant.xcodeproj/project.pbxproj
@@ -103,6 +103,7 @@
 		F9C165C22740403600592F76 /* CreateView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F9C165BB2740403600592F76 /* CreateView.swift */; };
 		F9C165C42740403600592F76 /* SentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F9C165BD2740403600592F76 /* SentView.swift */; };
 		F9EEB8162742C2210032EEB8 /* WithStateBinding.swift in Sources */ = {isa = PBXBuildFile; fileRef = F9EEB8152742C2210032EEB8 /* WithStateBinding.swift */; };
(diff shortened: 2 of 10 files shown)


secant/Features/Send/Views/SendView.swift
```diff
@@ -0,0 +1,58 @@
...
+
+    let store: Store<SendState, SendAction>
+
+    var body: some View {
+        WithViewStore(store) { viewStore in
+            Create(
+                transaction: viewStore.bindingForTransaction,
+                isComplete: viewStore.bindingForApprove
+            )
+            .navigationLinkEmpty(
```
adamstener, line 19: As we sort of discussed offline, this seems like one of those judgment calls.   I don't think all nesting is bad here, and it certainly could be the cleanest implementation, but having ways to avoid this getting out of hand will probably become important.

adamstener (approved):
I'd be interested to see what that view modifier could look like but also nesting these views so that each one goes one level deep seems super appropriate.

secant/Features/Send/Views/SendView.swift
pacu: The main hurdle I had with Navigation is that NavigationLinks that are tied to components that might end up being disabled by state changes, start to corrupt the navigation stack.
For example. The Home screen condenses most of the state. In the reference wallet we had screens that were not accessible while syncing. those were the send screen and the wallet history screen. Innocently (and following Apple's initial documentation) one would do something like

```` swift
NavigationLink(destination: WalletHistory()) { 
    Text("Wallet History")
 }
 .disabled(viewModel.isSyncing)
 ````
 Pretty neat huh? well... no. This is a navigation bug ensured, because whenever the Synchronizer started to sync, the navigation stack would reset because the navigation link tied to the button was being disabled. 
 
Our app has the particular thing of being "passively" (it actually polls) receiving updates from the blockchain and having to react to them while the user is minding its own business. It could be the case that the user has the same seed in many different wallets and uses them depending on context (which is a not-so-good-practice, but it happens). So it could be the case that the user tried to use ZecWallet and sent zec. Then went back to Secant wallet minutes later and tried to send zec to another user he interacted originally on this app and then the blockchain updates with changes originated from the same keys but from another app, causing a balance change and disrupting a flow that needed certain conditions to be correctly shown like, the send flow, that depends on the wallet having a positive verified balance. 

If the navigation depends directly on UI components triggering the NavigationLink, the updates on the main State will trigger a navigation stack reset because the render engine understands the View Hierarchy branch has to be pruned since is not longer valid according to the State it is reading from the main view. 

So, done with the constitutional preamble 😜. Even though SwiftUI allows you to retroactively evaluate the navigation stack (mostly inadvertently), a safe and predictable Navigation state is the combination of user intent + application state at the time of the intent.
My main question is how can we express that in a way that is clear and easy to read (and debug) for developers and that doesn't leverage assumptions (like modifier evaluations) that could break by apple moving stuff around between releases, and that also is clearly assertive on code.

dh-ecc: > Then went back to Secant wallet minutes later and tried to send zec to another user he interacted originally on this app and then the blockchain updates with changes originated from the same keys but from another app, causing a balance change and disrupting a flow that needed certain conditions to be correctly shown like, the send flow, that depends on the wallet having a positive verified balance.
> 
> If the navigation depends directly on UI components triggering the NavigationLink, the updates on the main State will trigger a navigation stack reset because the render engine understands the View Hierarchy branch has to be pruned since is not longer valid according to the State it is reading from the main view.

This is a really intersting edge case! Would be good to quickly walk through it on another channel, but I think I understand what you mean. How we'd handle it depends on the desired UX for when this does happen, but in general I think things like this solvable in that the "on/off" binding does not have to corrospond to something in the state being completely on or off.

For example suppose we don't want to randomly pop people back based on something like `isSyncing` we can pull this into the state of the view itself, and the `NavigationLink` could be based on `isSyncing || isCurrentlySending`.

I'm not sure if that clarifies anything 😅 but let's catch up offline if it makes sense.

pacu (approved):

secant/Features/Send/Views/SendView.swift
pacu: Yes, probably the best thing is to actually resolve navigation programmatically when the user navigates and then resolve what happens when the child (e.g: send flow) can't operate with the current wallet state.

PR #131, opened 2021-12-08
Integrate onboarding to fake home

pacu:
This PR is just to have a show and tell build on testflight.

### DIFF
--- a/secant/Features/Home/Views/HomeView.swift
+++ b/secant/Features/Home/Views/HomeView.swift
@@ -28,6 +28,13 @@ struct HomeView: View {
                 .primaryButtonStyle
                 .frame(height: 50)
 
+                Button(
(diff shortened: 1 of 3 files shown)


secant/Features/Home/Views/HomeView.swift
```diff
@@ -74,10 +74,10 @@ struct HomeView: View {
             .navigationLinkEmpty(
                 isActive: viewStore.showOnboardingBinding,
                 destination: {
-                    OnboardingView(
+                    OnboardingScreen(
                         store: Store(
                             initialState: OnboardingState(),
-                            reducer: .default,
+                            reducer: OnboardingReducer.default,
```
adamstener, line 80: You should still be able to just use .default here

PR #134, opened 2021-12-14
[WIP] Recovery Phrase Validation. 

### LINKED ISSUE #44: Seed Backup Flow
After the onboarding flow, the user selects "create new wallet".

This action entails that the a new random 24-word  BIP-39 compliant Mnemonic Seed Phrase is created for the user.

The user must be able to Backup the seed phrase and subsequently validate the backed up words against some sort of challenge.

pacu:
(eventually when finished) Closes #44 
This is a huge chunk of code. Corresponds to the Recovery Phrase Validation.

## Disclaimer:
There are probably many improvements to be made. Starting with the fact that Drag and Drop is a HUGE feature that can be used for dragging a simple view like a blue chip into another view or exporting a huge hierarchy of contents to a different application, crossing sandboxes and doing heavy-lifting serialization and deserialization with a MIME-type like classification. (see https://swiftui-lab.com/drag-drop-with-swiftui/) 

This has the huge effect that code seems a total overkill at its bare minimum.

I wouldn't bother much to review coding guidelines yet, since I haven't cleaned up the code in that regard. 

# Drag and Drop API Consequences 

The Delegate Pattern and SwiftUI DSL don't get along. There's not much documentation about how this plays out. The fact that `DropDelegate` are implemented as **struct** puts an additional weirdness layer to the cake. The APIs assume that there is always an action to take there. It's hard to define a state were you need to establish an `.onDrop` for a case where nothing can be done. I believe that the creators of the API meant that if there's nothing to be done, then the item shouldn't be made droppable. The issue with this is the code complexity that conditional statements bring to modifiers. It's clear that `if` statements and `ViewModifier`s don't get along well and that the readability of the resulting code, is poor.

## Ideas and motivation:
The Secret Recovery Phrase validation step is a puzzle game. As such, is represented as a (humble) State Machine. 

### State
A `RecoveryPhraseValidationState` represents the given state of the puzzle. It defines some constants that I did not know where to put, because they define both visual and logic calculations. 

The state holds a given `RecoveryPhraseValidationStep` that can change depending of the user's action.

A step is represented by an  Enum that ends up having the information needed to render the View.
Why isn't this a struct?  Well, it could be, but then there must be helper methods to associate a given step struct and the user action is representing or an additional property. 

````Swift
    case initial(phrase: RecoveryPhrase, missingIndices: [Int], missingWordsChips: [PhraseChip.Kind])
    case incomplete(phrase: RecoveryPhrase, missingIndices: [Int], completion: [RecoveryPhraseStepCompletion], missingWordsChips: [PhraseChip.Kind])
    case complete(phrase: RecoveryPhrase, missingIndices: [Int], completion: [RecoveryPhraseStepCompletion], missingWordsChips: [PhraseChip.Kind])
    case valid(phrase: RecoveryPhrase, missingIndices: [Int], completion: [RecoveryPhraseStepCompletion], missingWordsChips: [PhraseChip.Kind])
    case invalid(phrase: RecoveryPhrase, missingIndices: [Int], completion: [RecoveryPhraseStepCompletion], missingWordsChips: [PhraseChip.Kind])
````

Might as well be represented by a struct 
````Swift
struct RecoveryPhraseValidationStep {
    var phrase: RecoveryPhrase
    var missingIndices: [Int]
    var missingWordChips: [PhraseChip.Kind]
    var completion: [RecoveryPhraseStepCompletion] // this would be empty on the initial state
    var stepState: State
}

enum StepState {
    case initial
    case incomplete
    case complete
    case valid
    case invalid
}
````
or a third approach an `enum` that has an associated struct instead of tuples

````Swift
enum RecoveryPhraseValidationStep {
    case initial(RecoveryPhraseValidationStepState)
    case incomplete(RecoveryPhraseValidationStepState)
    case complete(RecoveryPhraseValidationStepState)
    case valid(RecoveryPhraseValidationStepState)
    case invalid(RecoveryPhraseValidationStepState)
 }
 
 struct RecoveryPhraseValidationStepState {
    var phrase: RecoveryPhrase
    var missingIndices: [Int]
    var missingWordChips: [PhraseChip.Kind]
    var completion: [RecoveryPhraseStepCompletion]? // this would be empty on the initial state
}
````

### Driving the state machine by reacting to user actions
The idea is that the API is pretty self explanatory (although ultra verbose).

Proposed actions
````Swift
enum RecoveryPhraseValidationAction: Equatable {
    case reset
    case drag(wordChip: PhraseChip.Kind, intoGroup: Int)
    case validate
    case succeed
    case fail
}
````
When the user is presented a state of the puzzle that requires a drag and drop action, the state has to respond to that action.

Given a User Step -> apply this chip into the given word group
````Swift
static func given(step: RecoveryPhraseValidationStep, apply chip: PhraseChip.Kind, into group: Int) -> RecoveryPhraseValidationStep {...}
````

When the user has completed the puzzle, it has to be validated 
````Swift
mutating func validate()
````
**Explicit feedback request**: Here is where I have doubts on what to do in the case of application errors that might want to set the state into validating a phrase that is not complete. 

### The Reducer 

````Swift

The way I implemented it does not convince me. I ended up needing `mutating` functions on the step `enum`. Which is some sort of code smell to me. Shall the reducer call static functions that produce a new enum with the next state and assigns it too the state instead?
extension RecoveryPhraseValidationReducer {
    static let `default` = RecoveryPhraseValidationReducer { state, action, _ in
        switch action {
        case .reset:
            state.reset()
            return .none

        case let .drag(wordChip, intoGroup):
            state.apply(chip: wordChip, into: intoGroup)
            return .none

        case .validate:
            state.step.validate()
            return .none

        case .succeed:
            return .none
        case .fail:
            state.reset()
            return .none
        }
    }
}
````

### The View
The `RecoveryPhraseBackupValidationView` is the view that represents the validation puzzle in a given step.
The most evident problem is that there is some UI logic to produce the sub-views that we need to display for a given step.

I've tried to place that in an extension, since I don't think it actually belongs in the step layer. My rationale behind is the following: If I wanted to do the terminal implementation of this puzzle, would this code belong here?
````Swift
extension RecoveryPhraseValidationStep {
    func wordsChips(for group: Int, groupSize: Int, from chunk: RecoveryPhrase.Chunk, with missingIndex: Int, completing completions: [RecoveryPhraseStepCompletion]) -> [PhraseChip.Kind] { ... }
    // and so on
}
````

One of the drawback of the data structures chosen is this problem where I need the phrase accessible at the view level but, it's encapsulated inside the validation step. 👎  🚩🚩🚩🚩🚩

````Swift
private extension RecoveryPhraseValidationStep {
    var phrase: RecoveryPhrase {
        switch self {
        case .initial(let phrase, _, _):
            return phrase
        case .incomplete(let phrase, _, _, _):
            return phrase
        case .complete(let phrase, _, _, _):
            return phrase
        case .invalid(let phrase, _, _, _):
            return phrase
        case .valid(let phrase, _, _, _):
            return phrase
        }
    }
}
````

## The tests
The tests are all about verifying the state machine's integrity
image

# Author

# Reviewer

### CODE REVIEW

**Review by adamstener — COMMENTED** [review]
I think it makes sense to first discuss state and state transitions, before getting into all the view related things with drag and drop.  As you pointed out in the description of this PR, I believe a Struct should be used here to help clean up how the state is managed, transitioned and consumed by views.  Before getting to the drag and drop stuff, I would suggest to first move this to using a Struct, remove static and mutating properties and functions, and simplify the data management.  Once that is done it will be easier to concentrate on just the view related concerns for presenting the data.

File: secant/Features/BackupFlow/RecoveryPhraseValidationStep.swift [hunk 1]
```diff
@@ -0,0 +1,145 @@
...
+            // replace the word on the copy of the original phrase with the completion the user did
+            words[concreteIndex] = wordCompletion.word
+        }
+
+        return words
+    }
+    /**
+    validates that the resulting word on the complete state matches the original word and moves the state into either valid or invalid
+    */
+    mutating func validate() {
```
adamstener, line 108: In a struct based world, this would be removed completely and the phase of the state would be computed based on the current values, so the state would automatically transition to .valid when the underlying data was in the right shape.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,111 @@
...
+    */
+    mutating func apply(chip: PhraseChip.Kind, into group: Int) {
+        self.step = RecoveryPhraseValidationStep.given(step: self.step, apply: chip, into: group)
+    }
+}
+
+enum RecoveryPhraseValidationAction: Equatable {
+    case reset
+    case drag(wordChip: PhraseChip.Kind, intoGroup: Int)
+    case validate
```
adamstener, line 82: It could be possible that if you move to a struct based value with computed properties, this would end up having only two cases, as the validate succeed and fail might all happen automatically with the computed values.

pacu: I'm definitely getting there, but I'm having doubts on how to achieve a proper transition to either success of fail screens upon completion. How that end up being will definitely shape this

adamstener: So when you drag that last word into place, if it's all correct, everything is valid and the screen automatically performs a transition?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,229 @@
...
+}
+
+/// Represents the completion of a group of recovery words by de addition of one word into the given group
+struct RecoveryPhraseStepCompletion: Equatable {
+    var groupIndex: Int
+    var word: String
+}
+
+struct RecoveryPhraseValidationState: Equatable {
+    enum Step: Equatable {
```
adamstener, line 26: I think this is less of a step than it is a status or a phase.  Also, what is the semantic difference between complete/invalid/valid?  If the status is complete, aren't you intrinsically either valid or invalid?  Or rather, if you are valid or invalid, aren't you automatically complete?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,229 @@
...
+typealias RecoveryPhraseValidationStore = Store<RecoveryPhraseValidationState, RecoveryPhraseValidationAction>
+typealias RecoveryPhraseValidationViewStore = ViewStore<RecoveryPhraseValidationState, RecoveryPhraseValidationAction>
+
+struct RecoveryPhraseEnvironment {
+    var mainQueue: AnySchedulerOf<DispatchQueue>
+    var newPhrase: () -> Effect<RecoveryPhrase, AppError>
+}
+
+/// Represents the completion of a group of recovery words by de addition of one word into the given group
+struct RecoveryPhraseStepCompletion: Equatable {
```
adamstener, line 20: Using the name `completion` here causes some readability issues with the notion of what most swift developers would think of as a "completion".  Usually that word is for describing a trailing closure or a callback that is to be executed upon completion of some sort of asynchronous code.  Is there another term we could use so that we don't muddy those naming waters?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,229 @@
...
+            completion: []
+        )
+    }
+}
+
+extension RecoveryPhraseValidationState {
+    /// Given an array of RecoveryPhraseStepCompletion, missing indices, original phrase and the number of groups it was split into,
+    /// assembly the resulting phrase. This comes up with the "proposed solution" for the recovery phrase validation challenge.
+    /// - returns:an array of String containing the recovery phrase words ordered by the original phrase order.
+    static func resultingPhrase(
```
adamstener, line 92: I think this would read a lot simpler if you just calculate this right in the computed property instead of making a static function that requires you to pass in most of the sate from the store state.  Does having this static function help you achieve something that I'm missing?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    var isValid: Bool {
+        Self.resultingPhrase(from: completion, missingIndices: missingIndices, originalPhrase: phrase, numberOfGroups: missingIndices.count) == phrase.words
+    }
+}
+
+
+extension RecoveryPhraseValidationState {
+    static func initial(
```
adamstener, line 61: I can' find the usages for this function, is it being used anywhere?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,229 @@
...
+            precondition(concreteIndex < words.count)
+            // replace the word on the copy of the original phrase with the completion the user did
+            words[concreteIndex] = wordCompletion.word
+        }
+
+        return words
+    }
+
+    /// drives the state machine represented on this Enum by the action of applying a chip into a group of words containing an empty slot that has to be completed
+    static func given(_ state: RecoveryPhraseValidationState, apply chip: PhraseChip.Kind, into group: Int) -> Self {
```
adamstener, line 122: Instead of a static function on State that has to take in the State, this could just be a regular function on State I believe.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,229 @@
...
+            phrase: phrase,
+            missingIndices: missingIndices,
+            missingWordChips: missingWordsChips,
+            completion: []
+        )
+    }
+
+    /// creates an initial `RecoveryPhraseValidationState` with no completions and random missing indices.
+    /// - Note: Use this function to create a random validation puzzle for a given phrase.
+    static func random(phrase: RecoveryPhrase) -> Self {
```
adamstener, line 76: Instead of a static function, this could just be an initializer if I'm understanding this correctly.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,229 @@
...
+                    missingWordChips: Array(repeating: .empty, count: RecoveryPhraseValidationState.phraseChunks),
+                    completion: newCompletion
+                )
+            }
+        default:
+            return state
+        }
+    }
+
+    static func pickWordsFromMissingIndices(indices: [Int], phrase: RecoveryPhrase) -> [PhraseChip.Kind] {
```
adamstener, line 171: I don't think any of these functions need to be static, everything appears to be able to be inlined (doesn't need a separate function) or they can all be a function that's a part of some Value, making it easier to read and reason about, and less properties to pass around.

secant/Features/BackupFlow/Views/RecoveryPhraseBackupValidationView.swift
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+}
+
+extension RecoveryPhraseValidationState {
+    static let placeholder = RecoveryPhraseValidationState.random(phrase: RecoveryPhrase.placeholder)
+}
+
+extension RecoveryPhraseValidationStore {
+    private static let scheduler = DispatchQueue.main
```
adamstener, line 113: I think this already lives in the environment right?

secant/Features/BackupFlow/Views/RecoveryPhraseBackupValidationView.swift
```diff
@@ -0,0 +1,126 @@
...
+                        maxWidth: .infinity,
+                        minHeight: 30
+                    )
+            }
+        }
+        .padding(0)
+    }
+}
+
+extension RecoveryPhraseValidationState{
```
adamstener, line 65: These functions seems very similar, and duplicate quite a bit of logic.  Is it possible to condense this down to just a single function instead of three?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        guard completion.count >= missingIndices.count else {
+            return .incomplete
+        }
+
+        return .complete
+    }
+
+    var isValid: Bool {
```
adamstener, line 54: if valid is already a phase or status of this state, should this not be part of the other computed property that determines the current state from the enum?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: Thanks for bringing this up. I ran out of names! Suggestions are very welcome :)

> Also, what is the semantic difference between complete/invalid/valid? If the status is complete, aren't you intrinsically either valid or invalid? Or rather, if you are valid or invalid, aren't you automatically complete?
This comes from a train of thought of driving the state for the screens. It's possible that either the complete or the valid/invalid states are not needed.

I initially thought of `complete` as the stage where you have completed and now the transition to failure screen or success screen has to be carried on. I don't know how you can actually do that elegantly thorough the Reducer so tha when completions equal the amount of groups you can drive the flow forward and trigger some action.  This comes more from my lack of understanding of TCA than something else.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: Yes, you are totally right. I may defer that to you both my dear native English speakers 😄

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: Yeay, It does not make much sense to have it dangling there anymore.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: Good catch. I'll delete it if so.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: This one is used on constructors. So it must be something external to the instance being created

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: renamed to apply(chip:into:)

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
adamstener: Can it just be used inline? Its only used once.  If you want a separate function, should this live on the RecoverPhrase so you don't have to pass it in and it can be a not-static function?

pacu:
This PR is ready for review

secant/Features/BackupFlow/Views/RecoveryPhraseBackupValidationView.swift
```diff
@@ -0,0 +1,176 @@
...
+            .applyScreenBackground()
+        }
+        .navigationBarTitleDisplayMode(.inline)
+        .navigationTitle(Text("Verify Your Backup"))
+    }
+}
+
+private extension RecoveryPhraseValidationStep {
+    @ViewBuilder var completionWords: some View {
+        switch self {
```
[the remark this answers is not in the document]
pacu: Yes this is being pulled out from the state now 👍

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: Finally the states are
````Swift
enum Step: Equatable {
        case initial
        case incomplete
        case complete
    }
````

File: secant/Features/BackupFlow/RecoveryPhraseValidation.swift (hunk 4 above)
Comment (pacu) [reply secant/Features/BackupFlow/RecoveryPhraseValidation.swift:16]: According to the thesaurus, "fulfillment" is a synonym of "completion" so I renamed this to `RecoveryPhraseStepFulfillment` accordingly.

File: secant/Features/BackupFlow/RecoveryPhraseValidation.swift (hunk 8 above)
Comment (pacu) [reply secant/Features/BackupFlow/RecoveryPhraseValidation.swift:73]: I think it could be confusing to the reader. There's a substancial different between the member-wise initializer and the resulting random initializer that semantically will look like a false-friend, something that in appearance does something but it's quite the opposite in reality.

````Swift
extension RecoveryPhraseValidationState {
    init(phrase: RecoveryPhrase) {
         let missingIndices = Self.randomIndices()
        let missingWordChipKind = Self.pickWordsFromMissingIndices(indices: missingIndices, phrase: phrase).shuffled()
        return RecoveryPhraseValidationState(
            phrase: phrase,
            missingIndices: missingIndices,
            missingWordChips: missingWordChipKind,
            completion: []
        )
    }
}
````

will be innocently autocompleted as RecoveryPhraseValidationState(phrase: ... )` but there's more to it that just a convenience method. That's why I chose to be explicit about it.

File: secant/UIComponents/Chips/EmptyChip.swift [hunk 14]
```diff
@@ -24,15 +26,25 @@ struct EmptyChip: View {
                 width: 4,
                 blur: 2
             )
+            .background(chipBackground)
             .frame(
                 minWidth: 0,
                 maxWidth: .infinity,
                 minHeight: 40,
-                idealHeight: 40,
+//                idealHeight: 40,
```
adamstener, line 34: Incidental commented code?

secant/UIComponents/Chips/EnumeratedChip.swift
```diff
@@ -8,102 +8,55 @@
 import SwiftUI
 
 struct EnumeratedChip: View {
+    let basePadding: CGFloat = 14
```
adamstener, line 11: Can we add a space after this to help the property wrapper stand out and read better?  either that or change the property wrapper to be on the same line as the index, either one would improve readability.

secant/UIComponents/Chips/PhraseChip.swift
```diff
@@ -45,13 +32,13 @@ struct PhraseChip_Previews: PreviewProvider {
...
-                .frame(height: 40)
+//                .frame(height: 40)
 
             PhraseChip(kind: .empty)
-                .frame(height: 40)
+//                .frame(height: 40)
 
             PhraseChip(kind: .ordered(position: 23, word: "mutual"))
-                .frame(height: 40)
+//                .frame(height: 40)
```
adamstener, line 41: Incidental commented code?

secant/Util/ScrollableWhenScaled.swift
```diff
@@ -0,0 +1,28 @@
+//
+//  ScrollableWhenScaled.swift
+//  secant-testnet
+//
+//  Created by Francisco Gindre on 12/27/21.
+//
+
+import SwiftUI
+
+struct ScrollableWhenScaledUpModifier: ViewModifier {
```
adamstener, line 10: If you move this into the extension on View below and make it private, that makes the only way it can be used is by accessing the public func you expose.

adamstener (changes requested):
I've done another round, this is definitely starting to look much cleaner!  I haven't gone through anything with. drag and drop yet or any of the unit tests, but I wanted to get this much out so we can keep making progress.  Nice work!

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,244 @@
...
+    var resultingPhrase: [String] {
+        precondition(missingIndices.count == fulfillments.count)
+        precondition(fulfillments.count == Self.phraseChunks)
+
+        var words = phrase.words
+        let groupLength = words.count / Self.phraseChunks
+        // iterate based on the completions the user did on the UI
+        for wordCompletion in fulfillments {
+            // figure out which phrase group (chunk) this completion belongs to
+            let i = wordCompletion.groupIndex
```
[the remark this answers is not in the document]
pacu: No, I have no swift lint warnings. Maybe it's not being called? That's weird!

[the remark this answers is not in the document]
adamstener: Looks like it was added to the exclusion list, we should remove that in a different PR.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,244 @@
...
+
+import Foundation
+import ComposableArchitecture
+import SwiftUI
+
+typealias RecoveryPhraseValidationStore = Store<RecoveryPhraseValidationState, RecoveryPhraseValidationAction>
+typealias RecoveryPhraseValidationViewStore = ViewStore<RecoveryPhraseValidationState, RecoveryPhraseValidationAction>
+
+/// Represents the fulfillment (completion) of a group of recovery words by the addition of one word into the given group
+struct RecoveryPhraseStepFulfillment: Equatable {
```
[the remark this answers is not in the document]
pacu: This `RecoveryPhraseStepFulfillment` is the data that represents the action of completing a spot of the puzzle with a given word.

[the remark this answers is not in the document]
adamstener: It's a specific word a specific index and you're using an array of them to perform validation correct?  I don't think this type represents an action at all, it's just an index and a string.  Because it is used as a collection, I think `ValidationWord` is much more appropriate in this case, unless I'm missing something.

secant/Util/ScrollableWhenScaled.swift
pacu: I ended up doing this. Doing what you described above threw numerous errors. I think the result is the same.

````Swift

fileprivate struct ScrollableWhenScaledUpModifier: ViewModifier {
    @ScaledMetric var scale: CGFloat = 1

    func body(content: Content) -> some View {
        if scale > 1 {
            ScrollView {
                content
            }
        } else {
            content
        }
    }
}

extension View {
    func scrollableWhenScaledUp() -> some View {
        self.modifier(ScrollableWhenScaledUpModifier())
    }
}
````

Comment (adamstener) [reply secant/Util/ScrollableWhenScaled.swift:10]:
```suggestion
import SwiftUI

extension View {
    private struct ScrollableWhenScaledUpModifier: ViewModifier {
        @ScaledMetric var scale: CGFloat = 1

        func body(content: Content) -> some View {
            if scale > 1 {
                return ScrollView {
                    content
                }
            } else {
                return content
            }
        }
    }

    func scrollableWhenScaledUp() -> some View {
        self.modifier(ScrollableWhenScaledUpModifier())
    }
}
```

You just need return statements in an extension

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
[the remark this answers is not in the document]
pacu: It represents a _past_ action of the user. I don't like the name either, but `ValidationWord` doesn't easily represent what that struct is about though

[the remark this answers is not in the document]
adamstener: It can represent the outcome of a past action, but it doesn't represent the action itself imho.  Another way to think about that is that the unit tests don't have actions, you can unit test the validation and data management here without performing any actions, and you would do that with these `ValidationWord`s.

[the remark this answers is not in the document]
adamstener: Since this is a Word used for Validation, and that's all the data that it holds, what part of it do you think we are not representing with the name "ValidationWord"?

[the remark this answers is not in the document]
adamstener: Another follow up question for thought, what if the UI changed and there wasn't drag and drop anymore, but instead the user picked between completed versions of a phrase?  In this scenario, there would still be an incomplete phrase but various words placed at various indexes would be used for validation, and this struct would still represent that data.  Should the UI requirements shape the naming of our data model?

secant/Util/ScrollableWhenScaled.swift
pacu: this is forbidden by the lint rules

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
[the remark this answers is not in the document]
pacu: what? i? the most widely known abbreviation for indices?   The use of `i` as index is no code smell.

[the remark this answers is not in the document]
adamstener: It's not code smell these are just the standard Swift guidelines that most every project would use.  Single character variables are widely discouraged in all cases.

secant/Util/ScrollableWhenScaled.swift
adamstener: I would just go with your fileprivate solution then 👍

secant/Features/BackupFlow/View+WhenDraggable.swift
```diff
@@ -0,0 +1,57 @@
...
+                NSItemProvider(object: word as NSString)
+            }
+        default:
+            self
+        }
+    }
+}
+
+extension View {
+    func onDrop(
```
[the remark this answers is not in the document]
pacu: it's used on the same file line 53

secant/Features/BackupFlow/Views/RecoveryPhraseBackupValidationView.swift
```diff
@@ -0,0 +1,266 @@
...
+                                chunk: chunk,
+                                misingIndex: index
+                            )
+                            .background(Asset.Colors.BackgroundColors.phraseGridDarkGray.color)
+                            .whenIsDroppable(!state.groupCompleted(index: index), dropDelegate: state.dropDelegate(for: viewStore, group: index))
+                        }
+                        Spacer()
+                    }
+                    .padding()
+                    .padding(.top, 0)
```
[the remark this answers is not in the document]
pacu: No, it doesn't. if you swap the order, it does.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,244 @@
...
+            indices.append(Int.random(in: 0 ..< wordGroupSize))
+        }
+        return indices
+    }
+}
+
+extension RecoveryPhrase.Chunk {
+    /// Returns an array of words where the word at the missing index will be an empty string
+    func words(with missingIndex: Int) -> [String] {
+        precondition(missingIndex >= 0)
```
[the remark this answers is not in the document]
pacu: This failures imply a logic flaw. I'll make them assertions.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,244 @@
...
+            let realIndex = (index * Self.wordGroupSize) + position
+            words.append(.unassigned(word: phrase.words[realIndex]))
+        })
+        return words
+    }
+
+    static func randomIndices() -> [Int] {
+        var indices = [Int]()
+        for _ in (0 ..< phraseChunks) {
+            indices.append(Int.random(in: 0 ..< wordGroupSize))
```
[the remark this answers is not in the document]
pacu: For the unit tests hasn't been necessary. But if we want to do UI test we will have to inject a random provider as you say.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: not quite so because the helper function is used to create the test scenarios and other helper functions that need this behavior to be outside the struct.

secant/Util/ScrollableWhenScaled.swift
pacu: the fileprivate is forbidden by the linter :(

adamstener: You can comment it out

pacu: The problem with the suggested change is that the view builders are lost by the nesting or something like that. The compiler is not able to figure out the returning type and infers that they don't match 
image

pacu: I actually when for the case of declaring fileprivate and adding a swiftlint exception for that next line

secant/Features/App/Views/AppView.swift
```diff
@@ -18,17 +18,51 @@ struct AppView: View {
...
                 )
+
             case .startup:
                 ZStack(alignment: .topTrailing) {
                     StartupView(sendAction: viewStore.send)
                 }
+
+            case .phraseValidation:
+                NavigationView {
+                    RecoveryPhraseBackupValidationView(
```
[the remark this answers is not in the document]
adamstener: Was this missed?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
adamstener: What did we think about `Status` here?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,231 @@
...
+
+            // Trigger a delayed effect to proceed to the next step
+            if case .complete = state.step {
+                if state.isValid {
+                    return Effect(value: .succeed).delay(for: 1, scheduler: environment.mainQueue).eraseToEffect()
+                } else {
+                    return Effect(value: .fail).delay(for: 3, scheduler: environment.mainQueue).eraseToEffect()
+                }
+            }
+            return .none
```
adamstener, line 210: I was poking around in code with this reducer helper function, and I was able to clean up the function a bit and get rid of the intermediate state, which reduces the overall code by a fair amount.  If we just put this state manipulation code right in the reducer where we have access to inout state, all the state manipulations are now done directly in the reducer, and we don't need functions that produce new state where we have to overwrite the entire state value.

Your reducer grows by 19 lines of code but you get to get rid of a 50+ line of code function that contains state manipulation logic that is probably better off right here in the reducer, that is the reducers job after all.

There are a few reasons this seems very clean to me.  
1) state manipulations stay in the reducer itself
2) the Values we are working with are being widdled down to just representations of a Value in time, without mutating funcs or funcs producing new states.
3) In the future, if we want to read through any state manipulations in our store, they all happen right here in the reducer.

```suggestion
            guard case let .unassigned(word) = wordChip else { return .none }

            switch state.step {
            case .initial:
                guard let missingChipIndex = state.missingWordChips.firstIndex(of: wordChip) else { break }

                state.missingWordChips[missingChipIndex] = .empty
                state.validationWords = [ValidationWord(groupIndex: group, word: word)]

            case .incomplete:
                guard let missingChipIndex = state.missingWordChips.firstIndex(of: wordChip) else { break }

                if state.validationWords.count < (RecoveryPhraseValidationState.phraseChunks - 1) {
                    state.missingWordChips[missingChipIndex] = .empty
                    state.validationWords.append(ValidationWord(groupIndex: group, word: word))
                } else {
                    state.missingWordChips = Array(
                        repeating: .empty,
                        count: RecoveryPhraseValidationState.phraseChunks
                    )
                    state.validationWords.append(ValidationWord(groupIndex: group, word: word))
                }

            case .complete:
                if state.isValid {
                    return Effect(value: .succeed)
                        .delay(for: 1, scheduler: environment.mainQueue)
                        .eraseToEffect()
                } else {
                    return Effect(value: .fail)
                        .delay(for: 3, scheduler: environment.mainQueue)
                        .eraseToEffect()
                }
            }
            return .none
```

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,231 @@
...
+import Foundation
+import ComposableArchitecture
+import SwiftUI
+
+typealias RecoveryPhraseValidationStore = Store<RecoveryPhraseValidationState, RecoveryPhraseValidationAction>
+typealias RecoveryPhraseValidationViewStore = ViewStore<RecoveryPhraseValidationState, RecoveryPhraseValidationAction>
+
+/// Represents the completion of a group of recovery words by the addition of one word into the given group
+struct ValidationWord: Equatable {
+    var groupIndex: Int
```
adamstener, line 17: One thing that has been tough to digest in this PR, is that we seem to bounce back and forth between the terms "Group" and "Chunk".  I'm guessing this started by introducing a "chunking" function that breaks up an array into an array of arrays, which is a standard term for a function like that.  However, at some point in the line of using said chunks that are derived from that function, the "Chunk" name is used with things like chunkSize or even types like Phrase.Chunk.  

I would suggest for readability we try to keep this consistent by keeping the name "Chunk" for just the function that does said "Chunking", and then consistently use the term "group" afterwards.  I think this would help readers of this code understand how all these different pieces fit together better, because you'd have a consistent way of referencing this portion of your state.  i.e. you would end up with `Phrase.Group` and wordGroupSize or groupSize.

Code like this:
`let groupLength = words.count / Self.phraseChunks`
would read better and make more logical sense if it read:
`let groupLength = words.count / Self.phraseGroups` for example.

secant/Features/BackupFlow/Views/RecoveryPhraseBackupValidationView.swift
```diff
@@ -0,0 +1,272 @@
...
+struct RecoveryPhraseBackupValidationView: View {
+    let store: RecoveryPhraseValidationStore
+
+    var body: some View {
+        WithViewStore(self.store) { viewStore in
+            VStack {
+                header(for: viewStore)
+                    .padding(.horizontal)
+                    .padding(.bottom, 10)
+                ZStack {
```
adamstener, line 20: Can you add spaces between all these views, it makes it easier to see what's going on for the overall view structure.

secant/Features/BackupFlow/Views/RecoveryPhraseDisplayView.swift
```diff
@@ -52,28 +57,30 @@ struct RecoveryPhraseDisplayView: View {
...
-                .padding()
             }
+            .padding(.bottom, 20)
             .padding(.horizontal)
+            .padding(.top, 0)
+            .applyScreenBackground()
         }
-
+        
         // TODO: NavigationBar Style
```
adamstener, line 74: If the navigation bar is hidden now, do we need a TODO to style it?

secant/Features/BackupFlow/Views/ValidationFailedView.swift
```diff
@@ -0,0 +1,76 @@
...
+                                .bodyText()
+                                .fixedSize(horizontal: false, vertical: true)
+                        }
+
+                        Button(
+                            action: { viewStore.send(.reset) },
+                            label: { Text("I'm ready to try again") }
+                        )
+                            .activeButtonStyle
+                            .frame(height: 60)
```
adamstener, line 52: When a view initializer is multi line, making the trailing paren on a new line, the convention we've been following is to keep the modifiers indented to that same level so they aren't "dangling".

```suggestion
                        Button(
                            action: { viewStore.send(.reset) },
                            label: { Text("I'm ready to try again") }
                        )
                        .activeButtonStyle
                        .frame(height: 60)
```

secant/Features/BackupFlow/RecoveryPhraseDisplayStore.swift
```diff
@@ -83,6 +83,14 @@ struct RecoveryPhrase: Equatable {
     func toString() -> String {
         words.joined(separator: " ")
     }
+
+    func words(fromMissingIndices indices: [Int]) -> [PhraseChip.Kind] {
+        assert((indices.count - 1) * chunkSize <= self.words.count)
+
+        return indices.enumerated().map({ index, position in
+            .unassigned(word: self.words[(index * chunkSize) + position])
+        })
```
adamstener, line 92: Does trailing closure syntax work here?

```suggestion
        return indices.enumerated().map { index, position in
            .unassigned(word: self.words[(index * chunkSize) + position])
        }
```

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,231 @@
...
+
+    static let wordGroupSize = 6
+    static let phraseChunks = 4
+
+    var phrase: RecoveryPhrase
+    var missingIndices: [Int]
+    var missingWordChips: [PhraseChip.Kind]
+    var validationWords: [ValidationWord]
+    var route: Route?
+    var step: Step {
```
adamstener, line 41: Can we add a space between the stored properties and computed properties?  It helps to easily see what this type consists of.

```suggestion
    var route: Route?
    
    var step: Step {
```

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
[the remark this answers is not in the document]
adamstener: I noticed in reviewing the rest of this PR that we do use the term index and indices elsewhere.  I still think it's a good idea to do that here as well.  It fits better with Swifts naming conventions and style guides, and it keeps the whole project consistent.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,231 @@
...
+            assert(concreteIndex < words.count)
+            // replace the word on the copy of the original phrase with the completion the user did
+            words[concreteIndex] = wordCompletion.word
+        }
+
+        return words
+    }
+
+    /// drives the state machine represented on this Enum by the action of applying a chip into a group of words containing an empty slot that has to be completed
+    func apply(chip: PhraseChip.Kind, into group: Int) -> Self {
```
adamstener, line 113: See longer comment about removing this function and having the logic directly in the reducer, but more succinct.

adamstener (commented):
I think after this round we are in a really good spot with state management and manipulation 👍 .  I'm really liking how this is turning out to be Values that are mutated only in the reducer, or a few helper functions to initialize completely new state.

🎉

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: thanks this is a great suggestion :)

secant/Features/BackupFlow/Views/RecoveryPhraseDisplayView.swift
pacu: image
I understand this is a nav bar. Isn't is @geffenz

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,232 @@
...
+        assert(validationWords.count == Self.phraseChunks)
+
+        var words = phrase.words
+        let groupLength = words.count / Self.phraseChunks
+        // iterate based on the completions the user did on the UI
+        for wordCompletion in validationWords {
+            // figure out which phrase group (chunk) this completion belongs to
+            let i = wordCompletion.groupIndex
+            // validate that's the right number
+            assert(i < Self.phraseChunks)
```
adamstener, line 100: I'm thinking maybe the comments in the thread above this one aren't coming through because the conversation is resolved.  Just double checking ☝️ .

secant/Features/App/Views/AppView.swift
```diff
@@ -18,17 +18,52 @@ struct AppView: View {
...
+                        store: store.scope(
```
adamstener, line 38: Same comment wit hate above conversation, it was marked as resolved but the formatting is still off below.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,232 @@
...
+        var wordsApplyingMissing = self.words
+        wordsApplyingMissing[missingIndex] = ""
+        return wordsApplyingMissing
+    }
+}
+
+enum RecoveryPhraseValidationAction: Equatable {
+    case updateRoute(RecoveryPhraseValidationState.Route?)
+    case reset
+    case drag(wordChip: PhraseChip.Kind, intoGroup: Int)
```
adamstener, line 184: Suggestion:
Instead of naming things after UI implementation details, we keep the naming generic so that if the UI changes we don't have a store that is out of date with how the user interaction works.

I would consider naming this `move` so that it doesn't have anything to do with the term "drag and drop".

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,232 @@
...
+
+        case .updateRoute(let route):
+            state.route = route
+            return .none
+
+        case .proceedToHome:
+            return .none
+            
+        case .displayBackedUpPhrase:
+            return .none
```
adamstener, line 229: It's ok to have a `return .none` after the switch statement so you don't have to muddy up every case with the same return statement, if it works out better that way.

secant/Features/BackupFlow/DropDelegate.swift
```diff
@@ -0,0 +1,81 @@
...
+            return WordChipDropDelegate { chipKind in
+                switch chipKind {
+                case .unassigned:
+                    viewStore.send(.drag(wordChip: chipKind, intoGroup: group))
+                default:
+                    break
+                }
+            }
+        case .incomplete:
+            guard validationWords.first(where: { $0.groupIndex == group }) == nil else { return NullDelegate() }
```
adamstener, line 66:
```suggestion
            guard validationWords.first(where: { $0.groupIndex == group }) == nil else { 
                return NullDelegate()
             }
```

secant/Features/BackupFlow/DropDelegate.swift
```diff
@@ -0,0 +1,81 @@
...
+        case .initial:
+            return WordChipDropDelegate { chipKind in
+                switch chipKind {
+                case .unassigned:
+                    viewStore.send(.drag(wordChip: chipKind, intoGroup: group))
+                default:
+                    break
+                }
+            }
+        case .incomplete:
```
adamstener, line 65: Adding a space around your cases when they have this much logic can help readability of the function.

secant/Features/BackupFlow/DropDelegate.swift
```diff
@@ -0,0 +1,81 @@
...
+        }
+        return false
+    }
+}
+
+extension RecoveryPhraseValidationState {
+    func dropDelegate(
+        for viewStore: RecoveryPhraseValidationViewStore,
+        group: Int
+    ) -> DropDelegate {
```
adamstener, line 54: I'm trying to understand the reasoning behind the logic in this function.  If the user in the in the `initial` state, how can they _not_ be in the state where the chipKind is unassigned?


When the user is in the `incomplete` state, and a word is grabbed from a group but the group doesn't have a matching index, is that not an assertion failure type situation, where that state should technically never happen?  Do we allow users to put a word back or once they have chosen are they stuck with that position?

Once we are in the `complete` phase, can a user even do drag and drop? Do we allow them to put words back?

What I'm wondering is, given these questions about how this interaction works, do we even need a delegate that always returns false?  It seems like the situations where we are using one are largely error states or simply not possible to get into.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,232 @@
...
+//
+
+import Foundation
+import ComposableArchitecture
+import SwiftUI
+
+typealias RecoveryPhraseValidationStore = Store<RecoveryPhraseValidationState, RecoveryPhraseValidationAction>
+typealias RecoveryPhraseValidationViewStore = ViewStore<RecoveryPhraseValidationState, RecoveryPhraseValidationAction>
+
+/// Represents the completion of a group of recovery words by the addition of one word into the given group
```
adamstener, line 15:
```suggestion
/// Represents the data of a word that has been placed into an empty position, that will be used
/// to validate the completed phrase when all ValidationWords have been placed.
```

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,232 @@
...
+                    validationWords: newCompletionState
+                )
+            } else {
+                var newCompletion = validationWords
+                newCompletion.append(ValidationWord(groupIndex: group, word: word))
+
+                return RecoveryPhraseValidationState(
+                    phrase: phrase,
+                    missingIndices: missingIndices,
+                    missingWordChips: Array(repeating: .empty, count: RecoveryPhraseValidationState.phraseChunks),
```
adamstener, line 154:
```suggestion
                    missingWordChips: Array(
                        repeating: .empty, 
                        count: RecoveryPhraseValidationState.phraseChunks
                    ),
```

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,232 @@
...
+        case let .drag(wordChip, group):
+            state = state.apply(chip: wordChip, into: group)
+
+            // Trigger a delayed effect to proceed to the next step
+            if case .complete = state.step {
+                if state.isValid {
+                    return Effect(value: .succeed).delay(for: 1, scheduler: environment.mainQueue).eraseToEffect()
+                } else {
+                    return Effect(value: .fail).delay(for: 3, scheduler: environment.mainQueue).eraseToEffect()
+                }
```
adamstener, line 209:
```suggestion
                if state.isValid {
                    return Effect(value: .succeed)
                        .delay(for: 1, scheduler: environment.mainQueue)
                        .eraseToEffect()
                } else {
                    return Effect(value: .fail)
                        .delay(for: 3, scheduler: environment.mainQueue)
                        .eraseToEffect()
                }
```

secant/Features/BackupFlow/Views/RecoveryPhraseBackupValidationView.swift
```diff
@@ -0,0 +1,280 @@
...
+                            .background(Asset.Colors.BackgroundColors.phraseGridDarkGray.color)
+                            .whenIsDroppable(!state.groupCompleted(index: index), dropDelegate: state.dropDelegate(for: viewStore, group: index))
+                        }
+
+                        Spacer()
+                    }
+                    .padding()
+                    .padding(.top, 0)
+                    .navigationLinkEmpty(
+                        isActive: viewStore.bindingForRoute(.success),
```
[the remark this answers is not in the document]
pacu: Uhmmm I'm not sure if I follow. 
This is the custom binding is referring to. 
````Swift
 func bindingForRoute(_ route: RecoveryPhraseValidationState.Route) -> Binding<Bool> {
        self.binding(
            get: { $0.route == route },
            send: { isActive in
                return .updateRoute(isActive ? route : nil)
            }
        )
    }
````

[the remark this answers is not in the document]
Comment (adamstener) [reply secant/Features/BackupFlow/Views/RecoveryPhraseBackupValidationView.swift:45]: Yeah I saw the function, I didn't spend any time trying to see if anything else could be done but it was just a little strange to have two navigation links stacked on one view, when the isactive state and the destination use the same enum.  I thought maybe it would be possible to collapse this down to a single link by using a case path or another mechanism so that you could have a single generic function instead of stacking two.

File: secant/Features/BackupFlow/DropDelegate.swift (hunk 39 above)
Comment (pacu) [reply secant/Features/BackupFlow/DropDelegate.swift:54]: I've double-checked and the current logic no longer requires to pass a NullDelegate.

Comment (adamstener) [reply secant/Features/BackupFlow/DropDelegate.swift]: That's great news! That's huge!

File: secant/Features/App/App.swift [hunk 44]
```diff
@@ -80,7 +111,11 @@ extension AppState {
     static var placeholder: Self {
         .init(
             homeState: .placeholder,
-            onboardingState: .init()
+            onboardingState: .init(),
+            phraseValidationState: RecoveryPhraseValidationState.placeholder,
+            phraseDisplayState: RecoveryPhraseDisplayState(
+                phrase: RecoveryPhrase.placeholder
+            )
```
adamstener, line 118:
```suggestion
            phraseValidationState: .placeholder,
            phraseDisplayState: .init(phrase: .placeholder)
```

secant/Features/App/Views/AppView.swift
adamstener: Just bumping this one up

secant/Features/BackupFlow/DropDelegate.swift
```diff
@@ -0,0 +1,43 @@
...
+import Foundation
+import SwiftUI
+import OrderedCollections
+import ComposableArchitecture
+
+/// Drop delegate that accepts items conforming to `PhraseChip.validationWordTypeIdentifier`
+struct WordChipDropDelegate: DropDelegate {
+    var dropAction: ((PhraseChip.Kind) -> Void)?
+
+    func validateDrop(info: DropInfo) -> Bool {
```
adamstener, line 17: Is this used anywhere?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,236 @@
...
+/// Represents the data of a word that has been placed into an empty position, that will be used
+/// to validate the completed phrase when all ValidationWords have been placed.
+struct ValidationWord: Equatable {
+    var groupIndex: Int
+    var word: String
+}
+
+struct RecoveryPhraseValidationState: Equatable {
+    enum Step: Equatable {
+        case initial
```
adamstener, line 24: Because this can traverse from one state to another and they don't happen sequentially, I would use Status here for the name of your enum. Step implies that you will sequentially move through all of these, which could happen, but isn't required. In Onboarding, you sequentially move through steps either forward or backward, so this name applies better there, where here you can move to any state at any time, so this represents more of the current status of validating the phrase then it does an individual step.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,236 @@
...
+        guard !validationWords.isEmpty else {
+            return  .initial
+        }
+
+        guard validationWords.count >= missingIndices.count else {
+            return .incomplete
+        }
+
+        return .complete
+    }
```
adamstener, line 53:
```suggestion
    var step: Step {
        guard !validationWords.isEmpty else { return  .initial }
        guard validationWords.count >= missingIndices.count else { return .incomplete }

        return .complete
    }
```

secant/Features/BackupFlow/DropDelegate.swift
pacu: the `validateDrop`? or the whole delegate?
This is used by the drag and drop API from Swift UI (well actually UIKit through SwiftUI) to perform the drop

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: here's is when I realize I have a smaller monitor than yours ^^

adamstener: There are longer lines in this file already though

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: Strictly on the semantic meaning of the word is ok. But status is often use to denote states of a thing which is not directly actionable by the user. (like the connection status) You inquiry the status of something that's handled by a third party which owns the state of the process and informs the status. in this case the process is quite self actionable so the correct word for it as it is often used in software engineering would be State, but it's already in use by the encapsulating Struct :S

secant/Features/BackupFlow/DropDelegate.swift
pacu:
````Swift
 .whenIsDroppable(
                                !state.groupCompleted(index: index),
                                dropDelegate: WordChipDropDelegate { chipKind in
                                    viewStore.send(.move(wordChip: chipKind, intoGroup: index))
                                }
                            )
````

Comment (pacu) [reply secant/Features/BackupFlow/DropDelegate.swift:17]: then

````Swift
extension View {
    /// Makes a View accept drop types Self.validationWordTypeIdentifier when it is of kind .empty
    func whenIsDroppable(_ isDroppable: Bool, dropDelegate: DropDelegate) -> some View {
        self.modifier(MakeDroppableModifier(isDroppable: isDroppable, drop: dropDelegate))
    }
}

struct MakeDroppableModifier: ViewModifier {
    var isDroppable: Bool
    var drop: DropDelegate
    
    func body(content: Content) -> some View {
        if isDroppable {
            content.onDrop(of: [PhraseChip.validationWordTypeIdentifier], delegate: drop)
        } else {
            content
        }
    }
}
````

File: secant/Features/BackupFlow/RecoveryPhraseValidation.swift (hunk 46 above)
Comment (adamstener) [reply secant/Features/BackupFlow/RecoveryPhraseValidation.swift:24]: I don't think the word status has any implications on whom is responsible for editing it.

File: secant/Features/BackupFlow/DropDelegate.swift (hunk 45 above)
Comment (adamstener) [reply secant/Features/BackupFlow/DropDelegate.swift:17]: I misunderstood, I see now!

File: secant/Features/BackupFlow/RecoveryPhraseValidation.swift (hunk 46 above)
Comment (adamstener) [reply secant/Features/BackupFlow/RecoveryPhraseValidation.swift:24]: That being said, we do only have 3 cases in this enum now.  We had more before, that allowed for more non-sequential flow here.  With these 3 now, I think you can only move through them in order, forward or back, so over time I think this enum has turned into something that better suits the current name than what we originally had.  👍 We can leave it! 🙌

File: secant/Features/BackupFlow/RecoveryPhraseValidation.swift [hunk 48]
```diff
@@ -0,0 +1,231 @@
...
+    /// assembly the resulting phrase. This comes up with the "proposed solution" for the recovery phrase validation challenge.
+    /// - returns:an array of String containing the recovery phrase words ordered by the original phrase order.
+    var resultingPhrase: [String] {
+        assert(missingIndices.count == validationWords.count)
+        assert(validationWords.count == Self.phraseChunks)
+
+        var words = phrase.words
+        let groupLength = words.count / Self.phraseChunks
+        // iterate based on the completions the user did on the UI
+        for wordCompletion in validationWords {
```
adamstener, line 92:
```suggestion
        for validationWord in validationWords {
```

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: I must have had a preview pinned 😬

adamstener: Ah yeah, previews will do that to you for sure.

secant/UIComponents/Backgrounds/ScreenBackground.swift
```diff
@@ -10,10 +10,7 @@ import SwiftUI
     A Vertical LinearGradient that takes an array of Colors and renders them vertically in a centered fashion mostly used as a background for Screen views..
 */
 struct ScreenBackground: View {
-    var colors = [
-        Asset.Colors.ScreenBackground.gradientStart.color,
-        Asset.Colors.ScreenBackground.gradientEnd.color
-    ]
+    var colors: [Color]
```
adamstener, line 13: GitHub doesn't make this easy to see but can we add as pace below this variable?

secant/UIComponents/Backgrounds/ScreenBackground.swift
```diff
@@ -24,9 +21,10 @@ struct ScreenBackground: View {
 }
 
 struct ScreenBackgroundModifier: ViewModifier {
+    var colors: [Color]
```
adamstener, line 24: can we add a space below this one too?

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: Thanks for pointing this out. I'm implementing this change.

pacu: pushed

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: done

secant/Features/App/Views/AppView.swift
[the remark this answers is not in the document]
pacu: yes it was

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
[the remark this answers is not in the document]
pacu: I changed it to groupIndex since it doesn't hurt. 

I don't believe that single character variable names are wrong. They have their proper place and usage. Using i and j or x and y for matrices is an example of proper use.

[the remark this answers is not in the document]
adamstener: x, y and z are common for describing coordinates or measurements along axis' that's very true, but other usages are discouraged.  This is probably underscored by the fact that the default swiftlint configuration disallows them altogether.

secant/Features/BackupFlow/Views/RecoveryPhraseDisplayView.swift
pacu: Thanks, I'll do so. I never imagined that our compiler was so.... well.. I'll let the reader choose the adjective. 😅

secant/Features/BackupFlow/Views/ValidationSuccededView.swift
```diff
@@ -0,0 +1,101 @@
...
+//  SuccessView.swift
+//  secant-testnet
+//
+//  Created by Adam Stener on 12/8/21.
+//
+
+import SwiftUI
+import ComposableArchitecture
+
+struct ValidationSuccededView: View {
```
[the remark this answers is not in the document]
adamstener: Bumping this one up, this is a misspelling.

adamstener (commented):
Code wise I think this PR is in a really good spot, nice work! 🙌 I just bumped up the one comment about a misspelling, and I did run and test the code but it appears to not work currently.  I'm not able to drop any words in any location and the formatting of the columns seems to be off, where two columns are large and two are smaller, breaking the word formatting and causing it to wrap.

secant/Features/BackupFlow/Views/RecoveryPhraseBackupValidationView.swift
```diff
@@ -0,0 +1,289 @@
...
+                if let completedWord = validationWord?.word {
+                    chips.append(.unassigned(word: completedWord))
+                } else {
+                    chips.append(.empty)
+                }
+            } else {
+                chips.append(.ordered(position: (groupSize * groupIndex) + index + 1, word: word))
+            }
+        }
+        return chips
```
adamstener, line 137: You can get rid of the intermediate variable here if you just map over the words instead of using a for in loop.


```suggestion
        let validationWord = validationWords.first(where: { $0.groupIndex == groupIndex })
        return wordGroup.words.enumerated().map { index, word in
            if index == missingIndices[groupIndex] {
                if let completedWord = validationWord?.word {
                    return .unassigned(word: completedWord)
                } else {
                    return .empty
                }
            } else {
                return .ordered(position: (groupSize * groupIndex) + index + 1, word: word)
            }
        }
```

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -0,0 +1,189 @@
...
+/// Represents the data of a word that has been placed into an empty position, that will be used
+/// to validate the completed phrase when all ValidationWords have been placed.
+struct ValidationWord: Equatable {
+    var groupIndex: Int
+    var word: String
+}
+
+struct RecoveryPhraseValidationState: Equatable {
+    enum Step: Equatable {
+        case initial
```
adamstener, line 24: With the reducer change, initial and incomplete have no distinction.  You can remove one of these cases, which only leaves you with two, that could ultimately be modeled with a Boolean variable.

secant/Features/BackupFlow/Views/ValidationSuccededView.swift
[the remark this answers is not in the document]
adamstener: Looks like this is still open

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: oh yeah!

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
```diff
@@ -40,12 +34,8 @@ struct RecoveryPhraseValidationState: Equatable {
...
     var route: Route?
     
-    var step: Step {
-        guard !validationWords.isEmpty else { return  .initial }
-
-        guard validationWords.count >= missingIndices.count else { return .incomplete }
-
-        return .complete
+    var isComplete: Bool {
+        !validationWords.isEmpty && validationWords.count >= missingIndices.count
```
adamstener, line 38: Should this be >= or just == ?  When would the validation words have a larger number than the missing indeces count?

secant/Features/App/Views/AppView.swift
```diff
@@ -18,17 +18,54 @@ struct AppView: View {
...
+                    .toolbar(
+                        content: {
+                            ToolbarItem(
+                                placement: .navigationBarLeading,
+                                content: {
+                                    Button(action: {
+                                        viewStore.send(.updateRoute(.startup))
+                                    }) {
+                                        Text("Back")
+                                    }
```
adamstener, line 52:
```suggestion
                                    Button(
                                        action: { viewStore.send(.updateRoute(.startup)) },
                                        label: { Text("Back") }
                                    )
```

secant/Features/BackupFlow/Views/RecoveryPhraseBackupValidationView.swift
adamstener: This change is still valid to make, it's just getting rid of unnecessary code.

secant/Features/BackupFlow/RecoveryPhraseValidation.swift
pacu: changed

pacu:
Known issue:
UX: Secret Recovery Phrase Validation screen can't be shown on small devices #161
https://github.com/zcash/secant-ios-wallet/issues/161

Issue #159, opened 2022-02-17
Enhancement of the index clamping


LukasKorba:
## Is your feature request related to a problem? Please describe.
I noticed this cool property wrapper in the code
```swift
@Clamped(1...24)
    var index: Int = 1
```

defined as follows:
```swift
@propertyWrapper
struct Clamped<Value: Comparable> {
    var value: Value
    let range: ClosedRange<Value>
    var wrappedValue: Value {
        get { value }
        set {
            value = min(
                max(range.lowerBound, newValue),
                range.upperBound
            )
        }
    }

    init(wrappedValue: Value, _ range: ClosedRange<Value>) {
        self.value = wrappedValue
        self.range = range
    }
}
```

## Describe the solution you'd like
I think it serves the purpose to bound the seed index in a range like this but it's also very generic. In case we use the index at many places, the bound constants will be spread over several places.

## Alternatives you've considered
Since this clamping is specific one for the seed words and its limited amount, we could be less generic -> more specific to our use case as following:
```swift
    @SeedIndex
    var index = 1
```
and implementation of the property would be simplified to:
```swift
@propertyWrapper
struct SeedIndex {
    var wrappedValue: Int {
        didSet { wrappedValue = min(24, max(1, wrappedValue)) }
    }
}
```
That would also mean to have one source of bound constants.

PR #160, opened 2022-02-17
Enhancement of the clamping the index for the seed

LukasKorba:
Issue #159 

# Author

# Reviewer


secant/Util/SeedIndex.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  SeedIndex.swift
+//  secant-testnet
+//
+//  Created by Lukáš Korba on 17.02.2022.
+//
+
+@propertyWrapper
+struct SeedIndex {
```
pacu, line 9: Currently `@Clamped` is only being used on Enumerated chip. So, for this PR to be complete. the Clamped property wrapper should be deleted since it would be no longer in use.

secant/Util/SeedIndex.swift
```diff
@@ -0,0 +1,13 @@
...
+    var wrappedValue: Int {
+        didSet { wrappedValue = min(24, max(1, wrappedValue)) }
```
pacu, line 11: How is this different to what the Clamped property wrapper does? (except for the range that is now hardcoded).

pacu (changes requested):

secant/Util/SeedIndex.swift
LukasKorba: let's stick with the Clamped but check the detailed comment for the 2nd comment

secant/Util/SeedIndex.swift
LukasKorba: The SeedIndex in only syntactical and let's say contextual change. Instead of having a universal range clamp property that can be set to anything, hardcoding it as a SeedIndex would clarify the use case more. 

Let's stick with the Clamped for now but I modified it for a good reason.

here's the new code
```swift
@propertyWrapper
struct Clamped<Value: Comparable> {
    var value: Value
    let range: ClosedRange<Value>
    var wrappedValue: Value {
        get {
            min(
                max(range.lowerBound, value),
                range.upperBound
            )
        }
        set { value = newValue }
    }

    init(wrappedValue: Value, _ range: ClosedRange<Value>) {
        self.value = wrappedValue
        self.range = range
    }
}
```

Notice that I swapped bounding function from set to get. The reason is that at initialisation time the set is not called so the clamped index would hold the initial value even out of bounds as long as newValue is set.

The previous version of Clamped was used in 
```swift
struct EnumeratedChip: View {
    @Clamped(1...24)
    var index: Int = 1
```
so anytime EnumeratedChip is initialised 
```swift
// example
var myView = EnumeratedChip(index: 101, text: "dummy")
print(myView.index) // prints 101
```
The reason is `init` is called, `get` is called with the simple { value } set to 101.

LukasKorba:
Actually the proposed version is not efficient because of the logic on a side of get. We need to enforce `set` to be called in the init, then we can optimise the computed property.

Final proposal:
```swift
@propertyWrapper
struct Clamped<Value: Comparable> {
    var value: Value
    let range: ClosedRange<Value>
    var wrappedValue: Value {
        get { value }
        set { value = clamp(value: newValue, range: range) }
    }

    init(wrappedValue: Value, _ range: ClosedRange<Value>) {
        self.value = wrappedValue
        self.range = range
        
        value = clamp(value: wrappedValue, range: range)
    }
    
    private func clamp(value: Value, range: ClosedRange<Value>) -> Value {
        min(range.upperBound, max(range.lowerBound, wrappedValue))
    }
}
```

secant/Util/Clamped.swift
```diff
@@ -16,16 +16,17 @@ struct Clamped<Value: Comparable> {
...
-            )
-        }
+        set { value = clamp(value: newValue, range: range) }
     }
 
     init(wrappedValue: Value, _ range: ClosedRange<Value>) {
         self.value = wrappedValue
         self.range = range
+        
+        value = clamp(value: wrappedValue, range: range)
```
adamstener, line 26: Is this duplicating the setter already happening on line 23?

secant/Util/Clamped.swift
```diff
@@ -16,16 +16,17 @@ struct Clamped<Value: Comparable> {
...
+    }
+    
+    private func clamp(value: Value, range: ClosedRange<Value>) -> Value {
```
adamstener, line 29:
```suggestion
    private func clamp(_ value: Value, using range: ClosedRange<Value>) -> Value {
```

secant/Util/Clamped.swift
LukasKorba: Yeah, this is unfortunate but the clamp function can't be used before being initialised. So both value and range need to be set first in order to call clamp, hence "double" set. Aka set the value to the input one and clamp it later once initialised.

pacu (approved):

Issue #157, opened 2022-02-04
[Functional] Implement KeyStoring protocol 


pacu:
Implement the KeyStoring Protocol

````Swift
protocol KeyStoring {
    func importBirthday(_ height: BlockHeight) throws
    func exportBirthday() throws -> BlockHeight
    func importPhrase(bip39 phrase: String) throws
    func exportPhrase() throws -> String
    func areKeysPresent() throws -> Bool
    /**
    Use carefully: Deletes the seed phrase from the keychain
    */
    func nukePhrase()

    /**
    Use carefully: deletes the wallet birthday from the keychain
    */
    func nukeBirthday()

    /**
    There's no fate but what we make for ourselves - Sarah Connor
    */
    func nukeWallet()

    var keysPresent: Bool { get }
}

enum KeyStoringError: Error {
    case alreadyImported
    case uninitializedWallet
    case storageError(Error)
}
````

So that the wallet recovery phrase and birthday can be stored to keychain.

KeyStoring is an interface abstracted from the SeedManager class of the ECC Reference wallet
https://github.com/zcash/zcash-ios-wallet/blob/master/wallet/wallet/Utils/SeedManagement.swift

It uses the dependency KeychainSwift as a wrapper for they keychain C API. 

This is a key component to the wallet's security. The PR closing this issue must be reviewed and approved by @defuse  in addition to the usual iOS reviewers

pacu:
Update:

Given the suggestions that @ccjernigan made in the PR.
some approach could be (this is totally for reference, please apply your best judgement)
````Swift
struct StoredPhrase {
   var version: Int
   var phrase: String
   var birthday: BlockHeight? //Int
}
protocol KeyStoring {
    func importPhrase(_ phrase: StoredPhrase) throws
    func exportPhrase() throws -> StoredPhrase
    func areKeysPresent() throws -> Bool
    func updateBirthday(_ height: BlockHeight) throws
    
    /**
    There's no fate but what we make for ourselves - Sarah Connor
    */
    func nukeWallet()
}

enum KeyStoringError: Error {
    case alreadyImported
    case uninitializedWallet
    case storageError(Error)
    case unsupportedVersion(Int)
}
````

PR #204, opened 2022-03-11
Implementation of KeyStoring protocol

LukasKorba:
#157 

The RecoveryPhraseStorage is an implementation of KeyStoring protocol. The unit tests are testing the keychain functionality to ensure security and proper error handling.

@pacu please check first and once we agrree on the implementation part, let's invite more poeple (from security) to review

# Author

# Reviewer


secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,238 @@
...
+            /// Data to be stored
+            kSecValueData as String: data as AnyObject,
+            /// Uniquely identify this keychain accessor
+            kSecAttrService as String: forKey as AnyObject,
+            kSecAttrAccount as String: account as AnyObject,
+            kSecClass as String: kSecClassGenericPassword,
+            /// The data in the keychain item can be accessed only while the device is unlocked by the user.
+            /// This is recommended for items that need to be accessible only while the application is in the foreground.
+            /// Items with this attribute do not migrate to a new device.
+            /// Thus, after restoring from a backup of a different device, these items will not be present.
```
pacu, line 106: @geffenz @defuse  This is relevant in terms of backup and restore experience. If this is the accessibility we want for this keychain item (I'm pretty sure it is), we should communicate the side-effects to the users.

pacu (changes requested):

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: FYI:
There are several options of accessibility and for now I used the most protected and secure one possible. Other potential option is to set the accessibility to `whenUnlocked`. The result would be:

`This is recommended for items that need to be accessible only while the application is in the foreground. Items with this attribute migrate to a new device when using encrypted backups.`

I think these two options are the only possible for our case so the discussion should be based on "Do we want to allow migration of the stored seed and birthday from device to device and rely on Apple's encryption?". I personally experienced both, one wallet was seamlessly restored when I restored from iCloud backup while other requested to import the seed and start over.

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,238 @@
...
+    
+    enum KeychainError: Error, Equatable {
+        case duplicate
+        case unknown(OSStatus)
+        case dataEncodingFailed
+        case dataDecodingFailed
+        case noDataFound
+    }
+
+    private static let zcashSeedPhrase = "zcashSeedPhrase"
```
adamstener, line 27: static constants should be grouped together.

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,238 @@
...
+/// Zcash implementation of the keychain that is not universal but designed to deliver functionality needed by the wallet itself.
+/// All the APIs should be thread safe according to official doc:
+/// https://developer.apple.com/documentation/security/certificate_key_and_trust_services/working_with_concurrency?language=objc
+/// For seamless usage the interface is designed to deal with simple types:
+/// `Seed Phrase: String` - possible to `store` and `retrive` from the keychain. `Delete` operation is also available, use it with caution.
+/// `Birthday: BlockHeight` - possible to `store` and `retrive` from the keychain. `Delete` operation is also available, use it with caution.
+final class RecoveryPhraseStorage {
+    static let `default`:RecoveryPhraseStorage = RecoveryPhraseStorage()
+    
+    enum KeychainError: Error, Equatable {
```
adamstener, line 19: Nested types should come before all static properties.
From the docs:
Type
-nested types
-static properties
-constants
-variables
-computed properties
-init methods
-instance methods

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,238 @@
...
+            throw KeychainError.dataDecodingFailed
+        } catch {
+            throw error
+        }
+    }
+    
+    /// Use carefully:  Deletes birthday from the keychain!!!
+    @discardableResult
+    func deleteBirthday() -> Bool {
+        return deleteData(forKey: RecoveryPhraseStorage.zcashBirthday)
```
adamstener, line 85: You don't need to use the return keyword here.

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,238 @@
...
+            throw KeychainError.noDataFound
+        }
+        
+        return String(decoding: data, as: UTF8.self)
+    }
+    
+    /// Use carefully:  Deletes seed phrase from the keychain!!!
+    @discardableResult
+    func deleteSeedPhrase() -> Bool {
+        return deleteData(forKey: RecoveryPhraseStorage.zcashSeedPhrase)
```
adamstener, line 51: You don't need to use the return keyword here.

secantTests/Util/KeychainManagerTests.swift
```diff
@@ -0,0 +1,256 @@
...
+    override func setUp() {
+        super.setUp()
+        storage = RecoveryPhraseStorage()
+        deleteData(forKey: "zcashSeedPhrase")
+        deleteData(forKey: "zcashBirthday")
+    }
+    
+    override func tearDown() {
+        storage = nil
+        super.tearDown()
```
adamstener, line 27: It would be more typical to call super prior to executing any other functionality right?

secantTests/Util/KeychainManagerTests.swift
```diff
@@ -0,0 +1,256 @@
...
+            // check presence of the key/data
+            let data = data(forKey: "zcashBirthday")
+            
+            XCTAssertEqual(data, nil, "Keychain: keychain is expected to not find anything for key `zcashBirthday` but received some data.")
+        }
+    }
+    
+    // MARK: - Misc
+    
+    /// The followings methods are here purposely to not rely on `RecoveryPhraseStorage` in order to test functionality of JUST ONE method at a time
```
adamstener, line 175: I've noticed these methods are duplicated from the private functionality that lives inside the app.  The primary logic here is setting up the query though, what would you think about potentially sharing the query data definitions, so that there is less duplication and easier maintenance over time?

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,238 @@
...
+import Foundation
+
+/// Zcash implementation of the keychain that is not universal but designed to deliver functionality needed by the wallet itself.
+/// All the APIs should be thread safe according to official doc:
+/// https://developer.apple.com/documentation/security/certificate_key_and_trust_services/working_with_concurrency?language=objc
+/// For seamless usage the interface is designed to deal with simple types:
+/// `Seed Phrase: String` - possible to `store` and `retrive` from the keychain. `Delete` operation is also available, use it with caution.
+/// `Birthday: BlockHeight` - possible to `store` and `retrive` from the keychain. `Delete` operation is also available, use it with caution.
+final class RecoveryPhraseStorage {
+    static let `default`:RecoveryPhraseStorage = RecoveryPhraseStorage()
```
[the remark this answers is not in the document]
LukasKorba: removed

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,238 @@
...
+        case dataDecodingFailed
+        case noDataFound
+    }
+
+    private static let zcashSeedPhrase = "zcashSeedPhrase"
+    private static let zcashBirthday = "zcashBirthday"
+    
+    // MARK: - Seed Phrase Public Interface
+    
+    func setSeedPhrase(_ phrase: String) throws {
```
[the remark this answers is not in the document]
LukasKorba: After sync discussion  the agreement is to go with the `RecoveryPhrase` term. Refactored at all places

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,238 @@
...
+/// `Seed Phrase: String` - possible to `store` and `retrive` from the keychain. `Delete` operation is also available, use it with caution.
+/// `Birthday: BlockHeight` - possible to `store` and `retrive` from the keychain. `Delete` operation is also available, use it with caution.
+final class RecoveryPhraseStorage {
+    static let `default`:RecoveryPhraseStorage = RecoveryPhraseStorage()
+    
+    enum KeychainError: Error, Equatable {
+        case duplicate
+        case unknown(OSStatus)
+        case dataEncodingFailed
+        case dataDecodingFailed
```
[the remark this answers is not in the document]
LukasKorba: done

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: done

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,238 @@
...
+
+    func seedPhrase() throws -> String {
+        guard let data = data(forKey: RecoveryPhraseStorage.zcashSeedPhrase) else {
+            throw KeychainError.noDataFound
+        }
+        
+        return String(decoding: data, as: UTF8.self)
+    }
+    
+    /// Use carefully:  Deletes seed phrase from the keychain!!!
```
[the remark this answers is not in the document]
LukasKorba: After sync discussion the agreement is to go with the RecoveryPhrase term. Refactored at all places

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: done

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: done

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: done

secantTests/Util/KeychainManagerTests.swift
LukasKorba: done

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,229 @@
...
+        case duplicate
+        case unknown(OSStatus)
+        case encoding
+        case decoding
+        case noDataFound
+    }
+
+    static let `default`:RecoveryPhraseStorage = RecoveryPhraseStorage()
+
+    enum Constants {
```
adamstener, line 25: If you make another nested type for your constants, it should go above the static constants.

adamstener: Ideally this one would be above the keychain errors

secantTests/Util/RecoveryPhraseStorageTests.swift
```diff
@@ -0,0 +1,228 @@
...
+        case .storageError: return "storageError"
+        }
+    }
+}
+
+class RecoveryPhraseStorageTests: XCTestCase {
+    // MARK: - Seed Phrase tests
+    
+    var storage: RecoveryPhraseStorage?
+    let recoveryPhraseSet = "one two three"
```
adamstener, line 25: Constants should come before variables.

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,229 @@
...
+        static let zcashRecoveryPhrase = "zcashRecoveryPhrase"
+        static let zcashBirthday = "zcashBirthday"
+    }
+    
+    // MARK: - Recovery Phrase Interface
+    
+    private func setRecoveryPhrase(_ phrase: String) throws {
```
adamstener, line 32: Would a private extension help clean this file up some, instead of private functions?

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: done

secantTests/Util/RecoveryPhraseStorageTests.swift
LukasKorba: done

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: done

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,229 @@
...
+    
+    /// Use carefully:  Deletes birthday from the keychain!!!
+    @discardableResult
+    private func deleteBirthday() -> Bool {
+        deleteData(forKey: Constants.zcashBirthday)
+    }
+    
+    // MARK: - Query Helpers
+
+    static func baseQuery(forAccount account: String = "", andKey forKey: String) -> [String: Any] {
```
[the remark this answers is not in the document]
LukasKorba: done

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,229 @@
...
+        case noDataFound
+        case unknown(OSStatus)
+    }
+
+    static let `default`:RecoveryPhraseStorage = RecoveryPhraseStorage()
+
+    enum Constants {
+        static let zcashBirthday = "zcashBirthday"
+        static let zcashRecoveryPhrase = "zcashRecoveryPhrase"
+    }
```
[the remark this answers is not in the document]
LukasKorba: done already on my side but thx for the commit

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,232 @@
...
+        query[kSecReturnRef as String] = kCFBooleanFalse
+        query[kSecReturnPersistentRef as String] = kCFBooleanFalse
+        query[kSecReturnAttributes as String] = kCFBooleanFalse
+        
+        return query
+    }
+}
+
+private extension RecoveryPhraseStorage {
+    // MARK: - Recovery Phrase Interface
```
adamstener, line 62:
```suggestion
}

    // MARK: - Recovery Phrase Interface

private extension RecoveryPhraseStorage {
```

This doesn't look like an interface though, is that the right term?

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,232 @@
...
+        do {
+            let data = try NSKeyedArchiver.archivedData(withRootObject: NSNumber(value: birthday), requiringSecureCoding: true)
+            
+            try setData(data, forKey: Constants.zcashBirthday)
+        } catch {
+            throw error
+        }
+    }
+    
+    // MARK: - General Interface
```
adamstener, line 120: Is this an interface, or is this comment not really needed?

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,232 @@
...
+        }
+        
+        guard status == errSecSuccess else {
+            throw KeychainError.unknown(status)
+        }
+    }
+}
+
+/// Implementation of KeyStoring protocol that uses iOS Keychain
+extension RecoveryPhraseStorage: KeyStoring {
```
adamstener, line 169:
```suggestion
}

// MARK: -  KeyStoring

extension RecoveryPhraseStorage: KeyStoring {
```

secantTests/Util/RecoveryPhraseStorageTests.swift
```diff
@@ -0,0 +1,228 @@
...
+        switch self {
+        case .alreadyImported: return "alreadyImported"
+        case .uninitializedWallet: return "uninitializedWallet"
+        case .storageError: return "storageError"
+        }
+    }
+}
+
+class RecoveryPhraseStorageTests: XCTestCase {
+    // MARK: - Recovery Phrase tests
```
adamstener, line 22: This seems pretty redundant and not needed

secant/Util/RecoveryPhraseStorage.swift
adamstener: Maybe ?
// MARK: - Recovery Phrase Public Interface

secantTests/Util/RecoveryPhraseStorageTests.swift
LukasKorba: done

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: removed

secantTests/Util/RecoveryPhraseStorageTests.swift
```diff
@@ -0,0 +1,228 @@
...
+            // check presence of the key/data
+            let data = data(forKey: "zcashBirthday")
+            
+            XCTAssertEqual(data, nil, "Keychain: keychain is expected to not find anything for key `zcashBirthday` but received some data.")
+        }
+    }
+    
+    // MARK: - Misc
+    
+    /// The followings methods are here purposely to not rely on `RecoveryPhraseStorage` in order to test functionality of JUST ONE method at a time
```
adamstener, line 193: Since you are PRAGMA marking these and commenting about the usage of the entire group, you could also do a private extension here, it might be easier to read.

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,232 @@
...
+
+    func setRecoveryPhrase(_ phrase: String) throws {
+        guard let data = phrase.data(using: .utf8) else {
+            throw KeychainError.encoding
+        }
+            
+        try setData(data, forKey: Constants.zcashRecoveryPhrase)
+    }
+
+    // MARK: - Birthday Interface
```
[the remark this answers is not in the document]
LukasKorba: removed

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,234 @@
...
+        query[kSecReturnData as String] = kCFBooleanTrue
+        query[kSecReturnRef as String] = kCFBooleanFalse
+        query[kSecReturnPersistentRef as String] = kCFBooleanFalse
+        query[kSecReturnAttributes as String] = kCFBooleanFalse
+        
+        return query
+    }
+}
+
+    // MARK: - Recovery Phrase Interface
```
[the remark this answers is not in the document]
LukasKorba: removed

secantTests/Util/RecoveryPhraseStorageTests.swift
LukasKorba: done

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,229 @@
...
+        query[kSecReturnData as String] = kCFBooleanTrue
+        query[kSecReturnRef as String] = kCFBooleanFalse
+        query[kSecReturnPersistentRef as String] = kCFBooleanFalse
+        query[kSecReturnAttributes as String] = kCFBooleanFalse
+        
+        return query
+    }
+}
+
+// MARK: - Recovery Phrase Interface
```
adamstener, line 61: These are private helper functions right? Not the interface?

LukasKorba: yes

LukasKorba: changed

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,228 @@
...
+        let query = RecoveryPhraseStorage.baseQuery(forAccount: account, andKey: forKey)
+
+        let status = SecItemDelete(query as CFDictionary)
+
+        return status == noErr
+    }
+    
+    /// Store data for key
+    func setData(
+        account: String = "",
```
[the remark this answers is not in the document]
LukasKorba: done

pacu (approved):

pacu:
utACK.

Waiting for @defuse to review

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,228 @@
+//
+//  RecoveryPhraseStorage.swift
+//  secant-testnet
```
ccjernigan, line 3: I think we can delete the line `secant-testnet`

pacu (approved):

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// MARK: - KeyStoring
+
+extension RecoveryPhraseStorage: KeyStoring {
+    func importBirthday(_ height: BlockHeight) throws {
+        do {
+            try setBirthday(height)
+        } catch RecoveryPhraseStorage.KeychainError.duplicate {
+            throw KeyStoringError.alreadyImported
```
ccjernigan, line 170: Will the default birthday be stored during wallet import if there is none provided by the user?  I want to clarify what our representation invariants are, and how we deal with updating the birthday later.

A possible scenario:
 - User does a wallet import
 - Birthday is optional, so user doesn't put one in.
 - App could either write nothing for birthday, or it could write the network default.  This is currently unspecified behavior
 - Later when the blockchain is scanned, it can update the birthday with the known correct value.  The main use would be for re-display to the user later.  This would necessitate writing the birthday which implies the prior step would need to write only the seed and not the birthday.

And what happens if the user puts in a bad birthday?  When we scan the blockchain, will we need to update the birthday with a corrected value?

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,228 @@
...
+            return false
+        } catch {
+            // TODO: - report & log error.localizedDescription
+            return false
+        }
+        
+        return true
+    }
+    
+    func nukePhrase() {
```
ccjernigan, line 216: Do we ever expect callers to call delete individually?  I wonder whether these individual methods should be `private`?

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,228 @@
...
+        deleteRecoveryPhrase()
+    }
+    
+    func nukeBirthday() {
+        deleteBirthday()
+    }
+    
+    func nukeWallet() {
```
ccjernigan, line 224: I'm concerned about atomicity here.  I think we might have a problem if a crash happens between `nukePhrase` and `nukeBirthday`.  Assuming that is true, some possible mitigations:
 - Re-order the delete so that we nuke the birthday first, then the phrase.  A nuked birthday is recoverable but a nuked phrase is not.
 - We store phrase and birthday as a single keychain item.  This guarantees the two items are always written atomically.  This is what the Android wallet is doing currently, although that doesn't necessarily mean it is the ideal design.

secantTests/Util/RecoveryPhraseStorageTests.swift
```diff
@@ -0,0 +1,227 @@
...
+    
+    override func tearDown() {
+        super.tearDown()
+        storage = nil
+    }
+    
+    func testRecoveryPhraseStoredSucessfuly() throws {
+        do {
+            try storage?.importPhrase(bip39: recoveryPhraseSet)
+            guard let data = data(forKey: "zcashRecoveryPhrase") else {
```
ccjernigan, line 41: Can this test not access `Constants.zcashRecoveryPhrase` so that we don't have to use a string here?

ccjernigan (commented):
Its great to see this functionality coming along!  I had a few comments inline, plus a few questions:

 - What's the performance of reading/writing the keychain?  When the UI accesses this, will it want to dispatch to a background thread to avoid blocking the UI?  If the answer is yes, should we consider an API that writes both phrase and birthday in one method?  It reduces the likelihood of a non-atomic write because we can control the ordering (write seed first, birthday second)
 - Do we need observability for changes to the keychain?  On Android, we're using the persisted seed as the source of truth for the app so the UI has an observer on it.  There are other design approaches, so curious to hear what we're thinking for iOS.
 - Should we store language with the seed phrase?  E.g. a default of en-us.  I think that's a parameter to the Mnemonic generator.  We only support en-us, but explicitly storing it saves us from possible migrations in the future.  Android doesn't store this current, but we have a followup issued filed that tracks it

secant/Util/RecoveryPhraseStorage.swift
pacu: This is a really good observation. There are many things to this point. When restoring from. recovery phrase, birthday is optional, since we don't know what's the origin of it. BIP-39 will give us reassurance of its integrity and ability to reconstruct the seed bytes, it won't prescript anything regarding wallet birthday. 

The user might not have the birthday or be unsure about it. So If it doesn't have it the wallet should assume sapling activation for syncing height. 

The ECC reference wallets differed in this point. The iOS wallet stored the resulting birthday from the restore process, that being the one that the uses wrote in, or sapling activation. The Android wallet does something additional which is updating the birthday to the least amount of blocks as possible. 

What does this mean?

Providing a birthday doesn't mean that the wallet has been initiated at that block height. The first assumption is that the wallet won't have any transactions involving those keys prior to the height signaled as birthday. The second assumption is that the wallet factually starts at that given height.

The first assumption is rare and expensive to verify, the second one could be conveniently verified when the wallet syncs past that given height.

The Android app does this. Even thought it accepts the birthday that the user provides as input, it overwrites it with a more recent block height if it has been found that the first received transaction has a greater height than the one stored in the secure storage. 

Although this is convenient, it can be confusing if it's not communicated properly to the user. I, as a. user, can be shown a backup screen that has a different birthday than the one I remember having backed up.

secant/Util/RecoveryPhraseStorage.swift
pacu: I find this suggestion really useful. Can you point us to the implementation of that so we can match it?

pacu (changes requested):
I'd like to take in @ccjernigan suggestion of implementing the underlying data as one item to ensure atomicity.

Although increments complexity of the code and was not initially required, I believe that it will prevent some considerable amount of boilerplate code around the UI of the wallet and error handling.

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: I think we need some general approach to this. There are 44 instances of `secant-testnet` in the project. The project itself is called secant-testnet, any contributor forking the repo will see secant-testnet xcode project, opens secant-testnet project, anytime when contributor creates a swift file, this line will be there automatically in the header. Removing it here is not going to do any meaningful work. If we won't do deal with it, we need separate ticket that is proposing general solution and deals with the future as well. I see several steps:
1. rename the project
2. refactor all secant-testnet instances
3. commit
4. any contributor/internal dev from that moment on will have right project name in the header and there will be no secant-testnet anymore

LukasKorba:
> * What's the performance of reading/writing the keychain?  When the UI accesses this, will it want to dispatch to a background thread to avoid blocking the UI?  If the answer is yes, should we consider an API that writes both phrase and birthday in one method?  It reduces the likelihood of a non-atomic write because we can control the ordering (write seed first, birthday second)

Good question, there are very little info, articles and doc about it. Based on my experience and also some of the keychain wrappers, no one deals with the threading. The API is thread-safe though. I found only some people who are heavily relying on keychain, storing massive data and frequently loading it. I would say keychain is as fast as UserDefaults except maybe some encryption layer involved in the computation. No one I know actually tried to put the store/restore code on background thread just for worrying it could block UI main thread for a millisecond. 

> * Do we need observability for changes to the keychain?  On Android, we're using the persisted seed as the source of truth for the app so the UI has an observer on it.  There are other design approaches, so curious to hear what we're thinking for iOS.

As far as I know there is no way how to observe the ios keychain. The keychain is very old and offers only simple API allowing to store/restore only, written in C basically. I see only one possible way, pull out the value at some intervals.

> * Should we store language with the seed phrase?  E.g. a default of en-us.  I think that's a parameter to the Mnemonic generator.  We only support en-us, but explicitly storing it saves us from possible migrations in the future.  Android doesn't store this current, but we have a followup issued filed that tracks it

@pacu to decide. Maybe an idea, it should be possible to write some regex to check the mnemonic language, right? I think as long as we have string, we should be able to say it's EN vs. Chinese..

LukasKorba:
> I'd like to take in @ccjernigan suggestion of implementing the underlying data as one item to ensure atomicity.
> 
> Although increments complexity of the code and was not initially required, I believe that it will prevent some considerable amount of boilerplate code around the UI of the wallet and error handling.

Can we list down the requirements and action points please? So far I think we'll merge recovery phrase with the birthday, that would mean to rewrite some code and tests. Some of the functions could be deleted completely if we do that.

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: is there any outcome of this discussion please?

secant/Util/RecoveryPhraseStorage.swift
ccjernigan: This is the Android implementation

Key-value preference store:
https://github.com/zcash/secant-android-wallet/blob/main/preference-impl-android-lib/src/main/java/co/electriccoin/zcash/preference/AndroidPreferenceProvider.kt

Serialization of wallet: (note that I will might want to change my format to be storing a `BirthdayHeight` instead of a `Birthday` since the `Birthday` has extra fields that I'm not sure we need to persist)
https://github.com/zcash/secant-android-wallet/blob/main/sdk-ext-lib/src/main/java/cash/z/ecc/sdk/model/PersistableWallet.kt

UI entry point to write the secret:
https://github.com/zcash/secant-android-wallet/blob/bb43d2f7a612b01ef0256bff3a45c932c6d44e60/ui-lib/src/main/java/co/electriccoin/zcash/ui/screen/home/viewmodel/WalletViewModel.kt#L152

ccjernigan:
If I can add one more suggestion: if we store the data as a single item, I'd like to suggest versioning the data too.

On Android, we went with storing it as a JSON string with a {version = 1} key-value pair.  That allows us to migrate what we store more easily in the future.

I don't know whether we want to use NSCoding to serialize or whether we want to store as JSON.  But at least having consideration for versioning will make our lives easier when we invariably need to modify the data format.

secant/Util/RecoveryPhraseStorage.swift
pacu: I created this issue. 
https://github.com/zcash/secant-ios-wallet/issues/207

secant/Util/RecoveryPhraseStorage.swift
pacu: Let's got with storing `nil` instead of default `saplingActivation` for when no birthday is provided

LukasKorba:
Updated to the agreed and updated concept (https://github.com/zcash/secant-ios-wallet/issues/157#issuecomment-1068316119)

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: solved, nukeWallet() is the only function now

secantTests/Util/RecoveryPhraseStorageTests.swift
LukasKorba: done

secant/MockedDependencies/KeyStoring.swift
```diff
@@ -7,32 +7,27 @@
 
 import Foundation
 
+struct StoredWallet: Codable, Equatable {
+    var birthday: BlockHeight?
+    let seedPhrase: String
+    let version: Int
+}
+
 protocol KeyStoring {
```
pacu, line 16: Now that we have defined the desired behavior, we should add documentation to this methods

secant/Stubs/MockServices.swift
```diff
@@ -6,6 +6,7 @@
 //
 // TODO: Move this to different Target when real functionality is developed.
 import Foundation
+import UIKit
```
pacu, line 9: Why is UIKit being added here?

pacu (changes requested):

pacu:
Great work @LukasKorba, I've marked a few little things

secant/Stubs/MockServices.swift
LukasKorba: removed

secant/MockedDependencies/KeyStoring.swift
LukasKorba: done

pacu:
> As far as I know there is no way how to observe the ios keychain. The keychain is very old and offers only simple API allowing to store/restore only, written in C basically. I see only one possible way, pull out the value at some intervals.
> 
@ccjernigan what changes is the UI observing on Android? Because of its nature there are not many (probably none) changes to the seed phrase related information on the app.


> Should we store language with the seed phrase? E.g. a default of en-us. I think that's a parameter to the Mnemonic generator. We only support en-us, but explicitly storing it saves us from possible migrations in the future. Android doesn't store this current, but we have a followup issued filed that tracks it
> @pacu to decide.

I think it doesn't hurt to do so. I don't know how it would be useful if not shown to the user when the phrase is displayed to the user for backup.

ccjernigan:
> @ccjernigan what changes is the UI observing on Android? Because of its nature there are not many (probably none) changes to the seed phrase related information on the app.

The UI is entirely based on an observer of the secret phrase/birthday, so there's a single source of truth whether a secret has been set or not. Think of it like a Combine operator that the rest of the UI observes.  It's then a simple switch statement to decide whether to show loading, onboarding, or the main view.  This is where that's happening: https://github.com/zcash/secant-android-wallet/blob/fb5fc153a533b5af802b31c3a695cec54a2ce07d/ui-lib/src/main/java/co/electriccoin/zcash/ui/MainActivity.kt#L106

Besides a single source of truth, another part of the motivation for this is that loading the secret requires disk IO, and we're avoiding that IO on the main thread by doing it this way.

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,248 @@
...
+        static let zcashKeychainVersion = 1
+    }
+
+    enum KeychainError: Error, Equatable {
+        case decoding
+        case duplicate
+        case encoding
+        case noDataFound
+        case unknown(OSStatus)
+        case unsupportedVersion(Int)
```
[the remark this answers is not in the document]
ccjernigan: I suggested adding locale to the serialized data format to avoid a data format migration in the future.

A new error code, on the other hand, could be added more easily in the future as it doesn't require a data format migration.  So we could add this now, but I don't think the timing is as important.

secant/MockedDependencies/KeyStoring.swift
```diff
@@ -7,32 +7,45 @@
 
 import Foundation
 
+/// Representation of the wallet stored in the persistent storage (typically keychain, handled by `RecoveryPhraseStorage`).
+struct StoredWallet: Codable, Equatable {
+    var birthday: BlockHeight?
+    let seedPhrase: String
```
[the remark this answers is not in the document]
ccjernigan: Yes here's where the locale could be added.  Even if it is always the same value, having it here would make things easier because we don't have to worry about the `Codeable` serializers changing.

pacu:
> > @ccjernigan what changes is the UI observing on Android? Because of its nature there are not many (probably none) changes to the seed phrase related information on the app.
> 
> The UI is entirely based on an observer of the secret phrase/birthday, so there's a single source of truth whether a secret has been set or not. Think of it like a Combine operator that the rest of the UI observes. It's then a simple switch statement to decide whether to show loading, onboarding, or the main view. This is where that's happening: https://github.com/zcash/secant-android-wallet/blob/fb5fc153a533b5af802b31c3a695cec54a2ce07d/ui-lib/src/main/java/co/electriccoin/zcash/ui/MainActivity.kt#L106
> 
> Besides a single source of truth, another part of the motivation for this is that loading the secret requires disk IO, and we're avoiding that IO on the main thread by doing it this way.

Oh, ok. Yes the app does have to check the keychain to figure out what to show.  On iOS, the keychain might not be available right away so it has to be deferred to a later stage where the keychain is available for sure.

secant/MockedDependencies/KeyStoring.swift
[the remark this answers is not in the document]
LukasKorba: Could we rely on?
```swift
    public static func determineLanguage(from mnemonicWords: [String]) throws -> MnemonicLanguageType {
        guard mnemonicWords.count > 0 else {
            throw MnemonicError.wrongWordCount
        }

        if String.englishMnemonics.contains(mnemonicWords[0]) {
            return .english
        } else if String.chineseMnemonics.contains(mnemonicWords[0]) {
            return .chinese
        } else {
            throw MnemonicError.unsupportedLanguage
        }
    }
```

It's already returning unsupportedLanguage. The method is clearly counting with uniqueness in the dictionaries but I think this would work for many possible languages as well. Extending it to spanish, french, german would work as well. 

@pacu @ccjernigan ?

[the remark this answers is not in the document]
LukasKorba: Otherwise if we really want to store `MnemonicLanguageType` we need to update the MnemonicSwift repo so the enum is `Codable`

secant/Stubs/MockServices.swift
```diff
@@ -58,117 +58,82 @@ class MockMnemonicPhraseHandling: MnemonicSeedPhraseHandling {
 }
 
 class KeysPresentStub: KeyStoring {
+    var called = false
+    var returnBlock: () throws -> Bool
```
adamstener, line 62: Should we name these after what is being called and what is being returned?  Since there are multiple functions here it might be easier to read the tests if the naming was more associated with what it is you are testing.
i.e. `areKeysPresentCalled` etc.

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    static let `default`: RecoveryPhraseStorage = RecoveryPhraseStorage()
```
adamstener, line 30: Do we need this to be a singleton?  Is there a reason we can't rely on dependency injection?

secant/Stubs/MockServices.swift
LukasKorba: Since MockServices.swift is not used anywhere, removing it completely from the project

secant/Stubs/MockServices.swift
```diff
@@ -58,117 +58,82 @@ class MockMnemonicPhraseHandling: MnemonicSeedPhraseHandling {
...
-    
-    func nukeBirthday() {
-        self.birthday = nil
+        storedWallet.birthday = height
     }
     
     func nukeWallet() {
-        nukePhrase()
-        nukeBirthday()
+        self.storedWallet = nil
```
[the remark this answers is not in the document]
LukasKorba: Since MockServices.swift is not used anywhere, removing it completely from the project

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: thinking about it, you are right, there's no special value in the singleton, removed

secant/MockedDependencies/KeyStoring.swift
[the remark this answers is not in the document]
pacu: As we spoke we probably want to store the locale of the phrase but only accept english to future-proof the bundle and avoid a potential migration.

secant/Util/RecoveryPhraseStorage.swift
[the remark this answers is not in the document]
LukasKorba: done

pacu (approved):

pacu:
utACK! 

please rebase and squash 💪

PR #209, opened 2022-03-21
User preferences storage

LukasKorba:
Issue #196 

Initial user preferences storage

# Author

# Reviewer

- [x] Automated tests: Did you review the automated tests?

### DIFF
--- a/secant/Util/UserPreferencesStorage.swift
+++ b/secant/Util/UserPreferencesStorage.swift
@@ -1,0 +1,115 @@
+//
+//  UserPreferencesStorage.swift
+//  secant-testnet
+//
+//  Created by Lukáš Korba on 03/18/2022.
+//
+
+import Foundation
+
+/// Representation of the user preferences stored in the local persistent storage (non-encrypted, no security needed)
+protocol UserPreferences {
+    /// From when the app is on and uninterrupted
+    var activeAppSessionFrom: TimeInterval { get set }
+    /// What is the set up currency
+    var currency: String { get set }
+    /// Whether the fiat conversion is on/off
+    var isFiatConverted: Bool { get set }
+    /// Whether user finished recovery phrase backup test
+    var isRecoveryPhraseTestCompleted: Bool { get set }
+    /// Whether the user has been autoshielded in the running session
+    var isSessionAutoshielded: Bool { get set }
+}
+
+/// Live implementation of the `UserPreferences` using User Defaults
+/// according to https://developer.apple.com/documentation/foundation/userdefaults
+/// the UserDefaults class is thread-safe.
+struct UserPreferencesStorage: UserPreferences {
+    enum Constants: String, CaseIterable {
+        case zcashActiveAppSessionFrom
+        case zcashCurrency
+        case zcashFiatConverted
+        case zcashRecoveryPhraseTestCompleted
+        case zcashSessionAutoshielded
+    }
+
+    static let `default` = UserPreferencesStorage(
+        appSessionFrom: Date().timeIntervalSince1970,
+        convertedCurrency: "USD",
+        fiatConvertion: true,
+        recoveryPhraseTestCompleted: false,
+        sessionAutoshielded: true,
+        userDefaults: UserDefaults.standard
+    )
+    
+    /// Default values for all preferences in case there is no value stored (counterparts to `Constants`)
+    private let appSessionFrom: TimeInterval
+    private let convertedCurrency: String
+    private let fiatConvertion: Bool
+    private let recoveryPhraseTestCompleted: Bool
+    private let sessionAutoshielded: Bool
+    
+    private let userDefaults: UserDefaults
+    
+    init(
+        appSessionFrom: TimeInterval,
+        convertedCurrency: String,
+        fiatConvertion: Bool,
+        recoveryPhraseTestCompleted: Bool,
+        sessionAutoshielded: Bool,
+        userDefaults: UserDefaults
+    ) {
+        self.appSessionFrom = appSessionFrom
+        self.convertedCurrency = convertedCurrency
+        self.fiatConvertion = fiatConvertion
+        self.recoveryPhraseTestCompleted = recoveryPhraseTestCompleted
+        self.sessionAutoshielded = sessionAutoshielded
+        self.userDefaults = userDefaults
+    }
+    
+    /// From when the app is on and uninterrupted
+    var activeAppSessionFrom: TimeInterval {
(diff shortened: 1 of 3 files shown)


pacu (approved):
utACK.

@LukasKorba how do you envision handling the default values for the App?

LukasKorba:
> @LukasKorba how do you envision handling the default values for the App?

Good question. At first I finished implementation with some default values but later I thought it's better to switch to optionals and let users of the storage to know IF the value is even present in the storage instead of fallback it to some value. 

We have several options how to deal with it. It's very likely that we will hold only 1 instance of the storage so we can extend the struct with defaults:
```swift
    enum Defaults {
        static let zcashRecoveryPhraseTestCompletedDefault = false
        // ...
    }

    // and use it 

    var isRecoveryPhraseTestCompleted: Bool? {
        get {
            userDefaults.object(
                forKey: Constants.zcashRecoveryPhraseTestCompleted
            ) as? Bool ?? Defaults.zcashRecoveryPhraseTestCompletedDefault
        }
        set { setValue(newValue, forKey: Constants.zcashRecoveryPhraseTestCompleted) }
    }
```

another option is to implement init and let users inject default values = hold constants for all default values from the input.

third option is to leave this storage untouched but all readers of the values would need to handle nil values which I think is in the end less comfortable. What do you think about Defaults enum? All defaults at one place. Nicer version of 
```swift
var isRecoveryPhraseTestCompleted: Bool? {
        get {
            userDefaults.object forKey: Constants.zcashRecoveryPhraseTestCompleted) as? Bool ?? false
        }
        set { setValue(newValue, forKey: Constants.zcashRecoveryPhraseTestCompleted) }
    }
```
as default values are not spread over the file... @pacu

pacu:
> another option is to implement init and let users inject default values = hold constants for all default values from the input.

@LukasKorba, I feel this option is the most testable of the ones you provided, what are your thoughts?

LukasKorba:
> thoughts?

agreed, already implemented and pushed

pacu (changes requested):

pacu (approved):

LukasKorba:
rebased and resolved conflicts @pacu

secant/Util/UserPreferencesStorage.swift
```diff
@@ -0,0 +1,115 @@
...
+//  UserPreferencesStorage.swift
+//  secant-testnet
+//
+//  Created by Lukáš Korba on 03/18/2022.
+//
+
+import Foundation
+
+/// Representation of the user preferences stored in the local persistent storage (non-encrypted, no security needed)
+protocol UserPreferences {
```
adamstener, line 11: Instead of a protocol based approach, have you considered extending UesrDefaults to work with your own `Key` type that defines the name and the Value Type that is returned (and maybe a default?), so that we can strongly type and build the keys we need but use the regular UserDefaults as the dependency?

One downside to the current approach that I see is that UserDefaults has many capabilities, and a larger API surface than just setting and retrieving values.  If we need any of that functionality at all, it will need to be exposed through this protocol because the underlying UserDefaults is privately held, or you'll have to work with the underlying source of truth external to this protocol.

Another factor here is that when you define a new key with this approach, it needs to be added to the protocol, added to the initializer for a default, and then implemented with a custom getter and setter for every key.

If we instead extend UserDefaults to work with strongly typed keys that we define, we could stick with a more vanilla dependency, reduce some of the duplications of data and allow us to work directly with UserDefaults.

From a testing perspective, you can setup static initializers on UserDefaults to put yourself in any state that you like, without needing a protocol.

https://www.swiftbysundell.com/articles/the-power-of-userdefaults-in-swift/
Some ideas here to reference if you want to consider this approach.

secant/Util/UserPreferencesStorage.swift
```diff
@@ -0,0 +1,115 @@
...
+        fiatConvertion: true,
+        recoveryPhraseTestCompleted: false,
+        sessionAutoshielded: true,
+        userDefaults: UserDefaults.standard
+    )
+    
+    /// Default values for all preferences in case there is no value stored (counterparts to `Constants`)
+    private let appSessionFrom: TimeInterval
+    private let convertedCurrency: String
+    private let fiatConvertion: Bool
```
adamstener, line 48: If we stick with this protocol approach, I don't think you need to hold onto these values.  You could instead just register the defaults values with UserDefaults every time this is initialized, which would also simplify your getters and setters.

ccjernigan:
How do we plan to manage observability for preferences?  E.g. a preference changes and other parts of the app can observe that change?

LukasKorba:
> How do we plan to manage observability for preferences? E.g. a preference changes and other parts of the app can observe that change?

based on @adamstener comment, I think I can refactor it to not use protocol and also achieve observability, on it..

adamstener:
> How do we plan to manage observability for preferences? E.g. a preference changes and other parts of the app can observe that change?

This is where TCA comes into play.  The state should be maintained in a store, and the store should be shared wherever this data is needed.  The source of truth for the UI should be the store, and the store uses the UserDefaults for persistence.  That way, whenever a store is updated, the internal value is updated, and all the users of the store are up to date, and the data is persisted to disk as a side effect.

secant/Util/UserPreferencesStorage.swift
pacu: >Instead of a protocol based approach, have you considered extending UesrDefaults to work with your own Key type that defines the name and the Value Type that is returned (and maybe a default?), so that we can strongly type and build the keys we need but use the regular UserDefaults as the dependency?

This sounds like a good idea to explore but I can't wrap my head around it. It sounds like a whole re-write though.

are you referring to so something like

````Swift
struct SettingKey<T: Any>: Hashable {
    associatedType T
    var key: String
    var value: T
}

extension UserDefaults {
    func getValue<K, T>(for key: K) -> T where K: SettingsKey, T: Settings.T {}
    func setValue<K, T>(_ value: T, for key: T) where K: SettingsKey, T: Settings.T {}
}
````

note: I wrote this on the comment editor, I'm not even sure if this is possible.

pacu (approved):
Looks good to me. Waiting on final comments from @adamstener to define how to follow up on his suggestion.

pacu:
> 3\. I think it's reasonable to have this as a new ticket and can be done with the next piece of work that uses this functionality



> 

This is a good path forward. 👍

pacu:
@LukasKorba can you solve the xcproj conflict?

pacu:
@LukasKorba 
build is failing with this error 

`❌  error: Build input file cannot be found: '/Users/vagrant/git/RecoveryPhraseStorageTests.swift' (in target 'secantTests' from project 'secant')`

Issue #222, opened 2022-03-30
Tests for the initialisation check and process


LukasKorba:
When issue #201 is successfully done and merged, the tests need to be provided. The issue itself is awaiting [PR 217](https://github.com/zcash/secant-ios-wallet/pull/217) to be merged before it can be cleaned up and finalised. When both PR and Issue are merged, provide tests for new actions and outcomes of those actions.

Namely:
```
.checkWalletInitialisation
.walletInitialisationResult(let walletInitialisationState):
.initialiseApp
```

Issue #220, opened 2022-03-30
Dependency for the SDK database files/file management


LukasKorba:
The SDK uses database files (cache, data, pending, ...). Create a dependency so it can be passed as a TCA environment.

Such dependency should handle:
- providing the URLs to the appropriate files (cache, data, ..) inside documents folder
- checking presence of the database(s)
- nuke files

PR #223, opened 2022-03-30
App initialization and checks

LukasKorba:
Issue #201 

Initialisation of the app and checks in place. The next is to update the Keychain dependency (PR 217) so this PR can be simplified and finished (issue #220).

Once 217 and 220 are merged, the tests need to be implemented (Issue #222)

# Author

# Reviewer


secant/Features/App/App.swift
```diff
@@ -43,6 +65,86 @@ extension AppReducer {
...
+            switch walletInitialisationState {
+            case .uninitialized:
+                return Effect(value: .updateRoute(.onboarding))
+                    .delay(for: 3, scheduler: environment.mainQueue)
+                    .eraseToEffect()
+            case .initialized, .keysPresentFilesMissing:
+                return Effect(value: .initialiseApp)
+            case .keysMissingFilesPresent:
+                state.appInitialisationState = walletInitialisationState
+                // TODO: error we need to handle, issue #221 (https://github.com/zcash/secant-ios-wallet/issues/221)
```
pacu, line 119: [non-blocking] 

This is currently an error but this state could be a "feature" when we support "Viewing Key Only" wallets.

secant/Util/InitialisationState.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+//  Created by Lukáš Korba on 30.03.2022.
+//
+
+import Foundation
+
+enum InitialisationState: Equatable {
+    case failure(ErrorWrapper)
+    case initialized
+    case keysMissingFilesPresent
```
pacu, line 13: [non-blocking] just FYI
On ECC Reference Wallet we called this state "Unprepared" but it's ok to leave it as it is for now since the `prepare` step on the sdk may not be around for the next iteration :)

secant/Util/RecoveryPhraseStorage.swift
```diff
@@ -220,7 +220,7 @@ extension RecoveryPhraseStorage: KeyStoring {
         }
     }
     
-    func areKeysPresent() throws -> Bool {
+    func areKeysPresent() -> Bool {
```
pacu, line 223: ⚠️ Warning: we might be masking a severe error under a harmless boolean. the throwing nature of this function was intended to allow us to have a "ternary state". Keys might be present or not, and that's fine. But then, it can be the case that they could also be not present or unavailable for other reasons (errors or inconsistencies of some sort) and we want to be able to tell that "not present" case apart.

secant/Util/ErrorWrapper.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+
+import Foundation
+
+/// Equatable wrapper for the errors, used in TCA's states
+struct ErrorWrapper: Equatable {
+    let error: Error
+    
+    public static func == (lhs: Self, rhs: Self) -> Bool {
+        String(reflecting: lhs.error) == String(reflecting: rhs.error)
```
pacu, line 15: This is really clever 😊

pacu (changes requested):

secant/Features/App/App.swift
```diff
@@ -43,6 +65,86 @@ extension AppReducer {
...
+        case .appDelegate(.didFinishLaunching):
+            return Effect(value: .checkWalletInitialisation)
+            
+            /// Checking presense of stored wallet in the keychain and presense of database files in documents directory.
+        case .checkWalletInitialisation:
+            var walletInitialisationState: InitialisationState = .uninitialized
+
+            let fileManager = FileManager()
+            // TODO: use updated dependency from PR #217 (https://github.com/zcash/secant-ios-wallet/pull/217)
+            let keysPresent = environment.recoveryPhraseStorage.areKeysPresent()
```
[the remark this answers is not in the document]
LukasKorba: without too much time spending googling about this topic, I moved the check to onAppear of WelcomeScreen, that should be enough but I'll spend some time around this topic later

secant/Features/App/App.swift
LukasKorba: Noted, seems like there's no action point at the moment but might be in the future.

secant/Util/InitialisationState.swift
LukasKorba: Noted, similarly, seems like there is no action point at the moment. Let's wait on how the SDK evolves.

secant/Util/RecoveryPhraseStorage.swift
LukasKorba: Agreed, PR 217 (https://github.com/zcash/secant-ios-wallet/pull/217) brings it back.

secant/Features/App/App.swift
```diff
@@ -43,6 +65,83 @@ extension AppReducer {
     )
     .debug()
 
+    private static let appReducer = AppReducer { state, action, environment in
+        switch action {
+            /// Checking presense of stored wallet in the keychain and presense of database files in documents directory.
+        case .checkWalletInitialisation:
+            var walletInitialisationState: InitialisationState = .uninitialized
+
+            let fileManager = FileManager()
```
adamstener, line 74: I think this would belong in the environment if I understand correctly.

secant/Features/App/App.swift
```diff
@@ -43,6 +65,83 @@ extension AppReducer {
...
+                walletInitialisationState = keysPresent ? .keysPresentFilesMissing : .uninitialized
+            } catch {
+                walletInitialisationState = .failure(ErrorWrapper(error: error))
+            }
+            
+            return Effect(value: .walletInitialisationResult(walletInitialisationState))
+
+            /// All possible responses to the `.checkWalletInitialisation` action result.
+            /// `uninitialized` results in the start of the Onboarding flow for the user.
+        case .walletInitialisationResult(let walletInitialisationState):
```
adamstener, line 106: Do you think this could just be combined with the other `check` action and have the one action just return these effects or set this state?  If so the action could probably then be called something like `initialize`?

secant/Features/App/App.swift
```diff
@@ -43,6 +65,83 @@ extension AppReducer {
...
+                return .none
+            case .failure(error: let error):
+                state.appInitialisationState = walletInitialisationState
+                // TODO: error we need to handle, issue #221 (https://github.com/zcash/secant-ios-wallet/issues/221)
+                return .none
+            }
+
+            /// Stored wallet is present, database files may or may not be present, trying to initialise app state variables and environments.
+            /// When initialisation succeeds user is taken to the home screen.
+        case .initialiseApp:
```
adamstener, line 126: In the US I think we standardize on using a `z` in the word initialize.  Can we standardize on that like we did in the other PR?

secant/Features/App/App.swift
```diff
@@ -43,6 +65,83 @@ extension AppReducer {
...
+                // TODO: error we need to handle, issue #221 (https://github.com/zcash/secant-ios-wallet/issues/221)
+                return .none
+            }
+            
+            state.appInitialisationState = .initialized
+            return Effect(value: .updateRoute(.startup))
+                .delay(for: 3, scheduler: environment.mainQueue)
+                .eraseToEffect()
+
+        default:
```
adamstener, line 140: Do we need a default here or can we rely on the compiler to help us know when actions have been added?

secant/Util/ErrorWrapper.swift
```diff
@@ -0,0 +1,17 @@
...
+//  ErrorWrapper.swift
+//  secant-testnet
+//
+//  Created by Lukáš Korba on 30.03.2022.
+//
+
+import Foundation
+
+/// Equatable wrapper for the errors, used in TCA's states
+struct ErrorWrapper: Equatable {
```
adamstener, line 11: Ideally errors are handled in the store and the state is updated based on the specific error with context so that the user of the store doesn't have to deal with `any error` but rather specifically known errors that are useful to the UI if needed.  I've noticed this is only used in one place that isn't marked with a TODO, should both places be marked with a todo for handling errors?  Is this wrapper temporary if we do implement error handling?  If so, we should probably have a TODO here on this type then as well I would think.

secant/Features/App/App.swift
pacu: For me is the same but the SDK already uses `z` so looking at it from that perspective, that's a good idea.

secant/Features/App/App.swift
pacu: good catch, I don't think we need a default action

pacu (changes requested):

secant/Features/App/App.swift
LukasKorba: yes, exactly issue #220 but I reordered some lines, the TODO was at different place

secant/Features/App/App.swift
LukasKorba: I had all in one action and then I decided to split it. The first part of the action was deciding X (many possible state) and other part of the action was deciding Y based on X. I think by splitting it we can test both separately. Mock some states of the keychain and test checkWallet action. And separately test the 2nd action. Thoughts @adamstener ?

secant/Features/App/App.swift
LukasKorba: fixed

secant/Features/App/App.swift
LukasKorba: the default here is a tradeoff. There's already AppReducer handling the routes and I made another AppReducer handling the actions. The routes one already had default. I think we are missing compilers switch check but I think this time it's worth it. Otherwise we would have plenty of unused cases. Thoughts @pacu @adamstener 

```
    private static let appReducer = AppReducer { state, action, environment in

    private static let routeReducer = AppReducer { state, action, _ in
```

secant/Util/ErrorWrapper.swift
pacu: There are going to be changes to Error handling so to avoid doing "future development" I wouldn't go around this too much

secant/Features/App/App.swift
LukasKorba: done

pacu (approved):
LGTM!

@adamstener @LukasKorba merge after resolving open conversations. 👍

secant/Features/App/App.swift
```diff
@@ -15,17 +15,42 @@ struct AppState: Equatable {
...
+    case initializeApp
     case onboarding(OnboardingAction)
     case phraseDisplay(RecoveryPhraseDisplayAction)
     case phraseValidation(RecoveryPhraseValidationAction)
+    case updateRoute(AppState.Route)
+    case walletInitialisationResult(InitialisationState)
+}
+
+struct AppEnvironment {
+    let mainQueue: AnySchedulerOf<DispatchQueue>
```
adamstener, line 36: I think you'll want to name this one scheduler.  Not only is this a type erased scheduler, it's also passed to functions that ask for a scheduler.  The fact that you end up using the main queue is just an implementation detail.

secant/Util/ErrorWrapper.swift
adamstener: I don't think you'll need the reflection or wrapper at all, you'll probably just need to define your own Error types that are equatable.  Throwing any error from underlying dependencies probably won't be useful to the UI or the consumer of the store.

secant/Util/InitialisationState.swift
adamstener, line 13: I think you can just call out what is missing here.  Because when you have both missing it's "uninitialized" then when either one of these is missing, you can just report on the missing piece.
i.e. `keysMissing` and `databaseMissing`

secant/SecantApp.swift
```diff
@@ -6,24 +6,31 @@
 //
 
 import SwiftUI
+import ComposableArchitecture
+
+final class AppDelegate: NSObject, UIApplicationDelegate {
+    var appStore: AppStore = .placeholder
+    lazy var appViewStore = ViewStore(
+        appStore.scope(state: { _ in () }),
```
adamstener, line 14: I _think_ you can call stateless here instead of using this scope
https://github.com/pointfreeco/swift-composable-architecture/blob/2828dc44f6e3f81d84bcaba72c1ab1c0121d66f6/Sources/ComposableArchitecture/Store.swift#L395

secant/Features/App/App.swift
adamstener: It doesn't seem like it's benefiting anything to separate these out and its adding a bit of indirection.  It might be easier to understand what's going on here as an external reader if it was just inline.

secant/Features/App/App.swift
LukasKorba: fixed

secant/Util/InitialisationState.swift
LukasKorba: fixed

secant/SecantApp.swift
LukasKorba: this is really great, thanks for letting me know about stateless as well as actionless, fixed

secant/Features/App/App.swift
LukasKorba: In the end you are right.. merged together, extra action removed

secant/Util/ErrorWrapper.swift
LukasKorba: simplified for now with `.failed` state

secant/Features/App/App.swift
```diff
@@ -82,11 +90,77 @@ extension AppReducer {
...
+                case (true, true):
+                    walletInitialisationState = .initialized
+                }
+            } catch CocoaError.fileNoSuchFile, CocoaError.fileReadNoSuchFile {
+                walletInitialisationState = keysPresent ? .filesMissing : .uninitialized
+            } catch {
+                walletInitialisationState = .failed
+            }
+
+            switch walletInitialisationState {
```
adamstener, line 125: Do we need this switch or could we just return these effects right inline?  I think you can remove the intermediate variable up there and just return what you need in the one do block.

adamstener (approved):

PR #292, opened 2022-04-25
WrappedSynchronizer

### LINKED ISSUE #291: Wrapped Synchronzier
The current CombineSynchronzier and way how publishing new values is done can be reimplemented to much simpler and testable version. 

### The issue(s) and solution(s)
**Publishing Issue**
Every time synchronzier updates the method updatePublishers is called. That one gets values for all publishers and broadcast those even though user can be looking at screen that doesn't need any of it. 

**Publishing Solutiuon**
Let's reimplement it so we inform any interested part by broadcasting synchronizerSynced as the only information. Anyone interested needs to subscribe to get know it and unsubscribes once it doesn't it anymore. 
The values needed will be requested as side effects for the requested values only. 

**Naming Issue**
The CombineSynchronizer is not very much Combine one, we don't have any other Synchronizers like RxSwift or similar, the Combine also doesn't reflect exactly the purpose. Technically it's more about wrapping Synchronizer. 

**Naming Solution**
We should then use WrappedSynchronizer or SynchronizerInteractor.

NOTE: when .synchronizerSynced is sent let's define possible stated of the synchronizes and pass it along. It should be all states from SDKSycnhronizer.swift file under public extension Notification.Name {.

LukasKorba:
Closes #291 

# Author

# Reviewer

### DIFF
--- a/secant.xcodeproj/project.pbxproj
+++ b/secant.xcodeproj/project.pbxproj
@@ -106,7 +106,7 @@
 		9EAFEB822805793200199FC9 /* AppReducerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EAFEB812805793200199FC9 /* AppReducerTests.swift */; };
 		9EAFEB84280597B700199FC9 /* WrappedSecItem.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EAFEB83280597B700199FC9 /* WrappedSecItem.swift */; };
 		9EAFEB862805A23100199FC9 /* WrappedSecItemTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EAFEB852805A23100199FC9 /* WrappedSecItemTests.swift */; };
-		9EAFEB882806E5AE00199FC9 /* CombineSynchronizer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EAFEB872806E5AE00199FC9 /* CombineSynchronizer.swift */; };
+		9EAFEB882806E5AE00199FC9 /* WrappedSDKSynchronizer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EAFEB872806E5AE00199FC9 /* WrappedSDKSynchronizer.swift */; };
 		9EAFEB8A2806F48100199FC9 /* ZCashSDKEnvironment.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EAFEB892806F48100199FC9 /* ZCashSDKEnvironment.swift */; };
 		9EAFEB8F2808183D00199FC9 /* SandboxView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EAFEB8D2808183D00199FC9 /* SandboxView.swift */; };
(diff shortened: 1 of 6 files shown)


secant/Wrappers/WrappedSynchronizer.swift
```diff
@@ -8,26 +8,51 @@
...
+    case stopped
+    case disconnected
+    case syncing
+    case downloading
+    case validating
+    case scanning
+    case enhancing
+    case fetching
+    case minedTransaction
+    case foundTransactions
```
pacu, line 30: What would be the difference between mined transaction and foundTransactions?

We are currently not using the mempool to fetch unmined transactions so every transaction not originated on this wallet will be detected when mined.

pacu (changes requested):

secant/Wrappers/WrappedSynchronizer.swift
LukasKorba: as agreed on slack, we need to support both

secant/Wrappers/WrappedSynchronizer.swift
```diff
@@ -8,26 +8,51 @@
...
+    case connectionStateChanged
+}
 
 struct Balance: WalletBalance, Equatable {
     var verified: Int64
     var total: Int64
 }
 
-protocol CombineSynchronizer {
+protocol WrappedSynchronizer {
```
pacu, line 40: I had made this comment but I don't see it anymore. I'm not sure this is the correct name for this protocol. Why is the synchronizer being wrapped? Shouldn't that be reflected in the name of the protocol? 

Wouldn't `ComposableSynchronizer` or something along those lines be more suitable?

pacu (changes requested):
Adding a comment that somehow got lost.

secant/Wrappers/WrappedSynchronizer.swift
LukasKorba: The naming is challenging, I know. Technically it's "wrapping" the SDKSynchronizer
```swift
class LiveWrappedSynchronizer: WrappedSynchronizer {
    private var cancellables: [AnyCancellable] = []
    private(set) var synchronizer: SDKSynchronizer?
```

`ComposableSynchronizer` is not fit as well, there is no real composable in it. Neither the TCA, neither the composability. This synchronizer is basically a dependency. Object providing interface to get some data in a form of Effects. It is not much different from WrappedMnemonic, WrappedFileManager, etc. It's using some live object iunder the hood to interact with, hence wrapped.

I'll give it today to think something else but WrappedSynchronizer is I believe "the best" for now.

LukasKorba: After it settled, I think the naming misses the context. What if we add SDK to it? 
```swift
protocol WrappedSDKSynchronizer
```
and then we simply define (exclude wrapped)
```swift
class LiveSDKSynchronizer: WrappedSDKSynchronizer { }

class MockSDKSynchronizer: WrappedSDKSynchronizer { }
```

pacu (approved):

Issue #329, opened 2022-05-26
Update wallet to use Zatoshi type


LukasKorba:
Once the task Zatoshi and Decimals #272 is considered done and approved (including unit tests) we will need to update the project to use the Zatoshi type. 

We will also need to cleanup Double extensions `asZec()` and `asZecString()`.

Steps:
- [x] Remove the supported extension Int64+Zcash and Double+Zcash
- [x] Make the project buildable again
- [x] Update balances and amounts to use Zatoshi (= find occurences of Int64 and replace it with Zatoshi)
- [x] remove TODOs
- [x] Update tests

Checks:
- [x] No Int64 related zatoshi should be present in the project anymore

Issues:
- [x] Transaction amount being in Zatoshi is clamping $ value to 21M max, that is a bug

Issue #312, opened 2022-05-09
WrappedNumberFormatter


LukasKorba:
The current implementation of the NumberFormatter:
```swift
extension NumberFormatter {
    static let zcashFormatter: NumberFormatter = {
        var formatter = NumberFormatter()
        formatter.maximumFractionDigits = 8
        formatter.maximumIntegerDigits = 8
        formatter.numberStyle = .decimal
        formatter.usesGroupingSeparator = true
        return formatter
    }()
}
```
is handling all the locale functionality we need to the wallet. The problem is that we can't control it as by default it uses the system language and region.

The task:
Implement the number formatter as a dependency, the mocked/test version can then be used to set the .locale on it:
```swift
formatter.locale = Locale(identifier: "en_US")
```

That was we can be sure we test right locale and string representatives of input text fields.

PR #330, opened 2022-05-26
[272] Decimals and Zatoshi type

### LINKED ISSUE #272: Decimals and Zatoshi type
We currently treat the balance as Int64, divided by some constants. Let's improve by going with decimals and Zatoshi type. Android team has already been working with holding it internally and converting to human readable format at the very last step, typically rendering the value for the UI.

LukasKorba:
Closes #272 
- Zatoshi type implemented
- ZEC type implemented with set of math operators
- conversions implemented
- unit tests for all kinds of operations with the Zatoshi and ZEC

# Author

# Reviewer

### DIFF
--- a/secant/Utils/Zatoshi.swift
+++ b/secant/Utils/Zatoshi.swift
@@ -1,0 +1,74 @@
+//
+//  Zatoshi.swift
+//  secant-testnet
+//
+//  Created by Lukáš Korba on 26.05.2022.
+//
+
+import Foundation
+
+struct Zatoshi {
+    enum Constants {
+        static let oneZecInZatoshi: Int64 = 100_000_000
+        static let maxZecSupply: Int64 = 21_000_000
+        static let maxZatoshi: Int64 = Constants.oneZecInZatoshi * Constants.maxZecSupply
+    }
+    
+    static var decimalHandler = NSDecimalNumberHandler(
+        roundingMode: NSDecimalNumber.RoundingMode.bankers,
+        scale: 8,
+        raiseOnExactness: true,
+        raiseOnOverflow: true,
+        raiseOnUnderflow: true,
+        raiseOnDivideByZero: true
+    )
+    
+    @Clamped(-Constants.maxZatoshi...Constants.maxZatoshi)
+    var amount: Int64 = 0
+
+    /// Converts `Zatoshi` to `NSDecimalNumber`
+    var decimalValue: NSDecimalNumber {
+        NSDecimalNumber(decimal: Decimal(amount) / Decimal(Constants.oneZecInZatoshi))
+    }
+
+    /// Converts `Zatoshi` to human readable format, up to 8 fraction digits
+    var decimalString: String {
+        decimalValue.roundedZec.stringValue
+    }
+
+    /// Converts `Decimal` to `Zatoshi`
+    static func from(decimal: Decimal) -> Zatoshi {
+        let roundedZec = NSDecimalNumber(decimal: decimal).roundedZec
+        let zec2zatoshi = Decimal(Constants.oneZecInZatoshi) * roundedZec.decimalValue
+        return Zatoshi(amount: NSDecimalNumber(decimal: zec2zatoshi).int64Value)
+    }
+
+    /// Converts `String` to `Zatoshi`
+    static func from(decimalString: String, formatter: NumberFormatter = NumberFormatter.zcashNumberFormatter) -> Zatoshi? {
+        if let number = formatter.number(from: decimalString) {
+            return Zatoshi.from(decimal: number.decimalValue)
+        }
+        
+        return nil
+    }
+    
+    static func + (left: Zatoshi, right: Zatoshi) -> Zatoshi {
+        Zatoshi(amount: left.amount + right.amount)
+    }
+
+    static func - (left: Zatoshi, right: Zatoshi) -> Zatoshi {
+        Zatoshi(amount: left.amount - right.amount)
+    }
+}
+
+extension NSDecimalNumber {
+    /// Converts `NSDecimalNumber` to human readable format, up to 8 fraction digits
+    var decimalString: String {
+        self.rounding(accordingToBehavior: Zatoshi.decimalHandler).stringValue
+    }
+    
+    /// Round the decimal to 8 fraction digits
+    var roundedZec: NSDecimalNumber {
+        self.rounding(accordingToBehavior: Zatoshi.decimalHandler)
+    }
+}
--- a/secant.xcodeproj/project.pbxproj
+++ b/secant.xcodeproj/project.pbxproj
@@ -101,6 +101,9 @@
 		9E2F1C8F280EDE09004E65FE /* Drawer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E2F1C8E280EDE09004E65FE /* Drawer.swift */; };
 		9E37A2B827C8F59F00AE57B3 /* Localizable.strings in Resources */ = {isa = PBXBuildFile; fileRef = 9E37A2B727C8F59F00AE57B3 /* Localizable.strings */; };
 		9E391124283E4CAC0073DD9A /* ImportWalletTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E391123283E4CAC0073DD9A /* ImportWalletTests.swift */; };
+		9E391129283F74590073DD9A /* Zatoshi.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E391128283F74590073DD9A /* Zatoshi.swift */; };
+		9E39112A283F90F10073DD9A /* Double+Zcash.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EDDEA8B28250F9C00B4100C /* Double+Zcash.swift */; };
+		9E39112E283F91600073DD9A /* ZatoshiTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E39112D283F91600073DD9A /* ZatoshiTests.swift */; };
 		9E4DC6E027C409A100E657F4 /* NeumorphicDesignModifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E4DC6DF27C409A100E657F4 /* NeumorphicDesignModifier.swift */; };
 		9E4DC6E227C4C6B700E657F4 /* SecantButtonStyles.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E4DC6E127C4C6B700E657F4 /* SecantButtonStyles.swift */; };
 		9E5BF63F2819542C00BA3F17 /* TransactionHistoryTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E5BF63E2819542C00BA3F17 /* TransactionHistoryTests.swift */; };
@@ -138,7 +141,6 @@
 		9EAFEB9228081E9400199FC9 /* HomeView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F93874EF273C4DE200F0E875 /* HomeView.swift */; };
 		9EBEF87A27CE369800B4F343 /* RecoveryPhraseValidationFlowView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EBEF87927CE369800B4F343 /* RecoveryPhraseValidationFlowView.swift */; };
 		9ECAE56827FC713C0089A0EF /* DatabaseFiles.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9ECAE56727FC713C0089A0EF /* DatabaseFiles.swift */; };
-		9EDDEA8C28250F9C00B4100C /* Double+Zcash.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EDDEA8B28250F9C00B4100C /* Double+Zcash.swift */; };
 		9EDDEAA22829610D00B4100C /* CurrencySelectionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EDDEA9F2829610D00B4100C /* CurrencySelectionTests.swift */; };
 		9EDDEAA32829610D00B4100C /* TransactionAmountInputTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EDDEAA02829610D00B4100C /* TransactionAmountInputTests.swift */; };
 		9EDDEAA42829610D00B4100C /* TransactionAddressInputTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EDDEAA12829610D00B4100C /* TransactionAddressInputTests.swift */; };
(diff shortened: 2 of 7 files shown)


LukasKorba:
I've done this task in a way that once we agree on this, we update the wallet code to use the Zatoshi/ZEC. Ticket #329

pacu:
@ccjernigan would you mind taking a look at this approach for representing Zatoshi and ZEC?

pacu (commented):
I've left a comment around the `isValid` field which seems it's kind of confusing.

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,103 @@
...
+    var humanReadable: ZEC {
+        self.rounding(accordingToBehavior: ZEC.nsDecimalHandler)
+    }
+    
+    /// Converts `ZEC` to `Zatoshi`
+    var asZatoshi: Zatoshi {
+        Zatoshi(amount: (self * NSDecimalNumber(value: 100_000_000)).int64Value)
+    }
+
+    static func + (left: ZEC, right: ZEC) -> ZEC {
```
pacu, line 70: I don't see where we could use these math operators.  Am I missing something?

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,103 @@
+//
+//  Zatoshi.swift
+//  secant-testnet
+//
+//  Created by Lukáš Korba on 26.05.2022.
+//
+
+import Foundation
+
+typealias ZEC = NSDecimalNumber
```
ccjernigan, line 10: Do we want to treat Zec as a "first class citizen" in terms of being a numeric value?

I've been thinking about it as basically just a UI string for display only, with the intention of preventing Zec from ever being used for anything other than display.

As a NSDecimalNumber, a client could theoretically add/subtract Zec values and go outside of the valid range for Zatoshi (when converted back).  As a string, a client would need to go to Zatoshi and do the operations there, then come back to the string representation.

Both ways could work, although always using Zatoshi seems to reduce our surface area for bugs.

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,103 @@
...
+extension Int64 {
+    var asZatoshi: Zatoshi { Zatoshi(amount: self) }
+    var asZec: ZEC { ZEC(value: Double(self)) }
+}
+
+// MARK: - ZEC
+
+extension ZEC {
+    static var nsDecimalHandler = NSDecimalNumberHandler(
+        roundingMode: ZEC.RoundingMode.plain,
```
pacu, line 53: I believe the expected rounding mode should be Bankers' 
see: https://github.com/zcash/secant-android-wallet/blob/e44e9016769450c4e312c25992ad73d3d123936e/sdk-ext-ui-lib/src/main/java/cash/z/ecc/sdk/ext/ui/model/ZecString.kt#L90

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,103 @@
...
+
+// MARK: - Zatoshi
+
+struct Zatoshi {
+    @Clamped(-(100_000_000 * 21_000_000)...(100_000_000 * 21_000_000))
+    var amount: Int64 = 0
+
+    /// Converts `Zatoshi` to `ZEC`
+    var asZec: ZEC {
+        ZEC(value: Double(amount) / Double(100_000_000))
```
ccjernigan, line 20: When doing this conversion, we may want to be doing this as NSDecimals instead of Doubles to avoid potential floating point precision issues.  I'm less familiar with the Swift APIs, although I think that's roughly equivalent to using Java's BigDecimal versus double.

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    static func - (left: Zatoshi, right: Zatoshi) -> Zatoshi {
+        Zatoshi(amount: left.amount - right.amount)
+    }
+}
+
+extension Int {
+    var asZatoshi: Zatoshi { Zatoshi(amount: Int64(self)) }
+    var asZec: ZEC { ZEC(value: Double(self)) }
```
ccjernigan, line 41: In line with the comment above, I might suggest that we eliminate the Zec conversion from a numeric value and instead require going through Zatoshi first.

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,103 @@
...
+//
+
+import Foundation
+
+typealias ZEC = NSDecimalNumber
+
+// MARK: - Zatoshi
+
+struct Zatoshi {
+    @Clamped(-(100_000_000 * 21_000_000)...(100_000_000 * 21_000_000))
```
ccjernigan, line 15: Could we put some of these values into named constants for clarity?  Then they could be re-used in the instances they appear below too.

secant/Utils/Zatoshi.swift
pacu, line 10: I could be missing something. But in terms of UI, I don't think we will be using ZEC as a numeric representation after all, but as a Decimal String representation of Zatoshi.

I see three approaches that are more limited but probably less error prone as we..

A. make ZEC a `struct` containing a Zatoshi and providing functions to and from string representations.
B. make a `String` extension and a `Zatoshi` extension for string representations (from and to)
C. create a ZecString struct that assures a given set of valid characters (see https://github.com/zcash/secant-android-wallet/blob/e44e9016769450c4e312c25992ad73d3d123936e/sdk-ext-ui-lib/src/main/java/cash/z/ecc/sdk/ext/ui/model/ZecString.kt#L59)

ccjernigan (commented):
I'm a bit less familiar with the Swift APIs, although I added some comments inline for discussion about how we want to think about Zec vs Zatoshi throughout our codebases.

pacu (changes requested):

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    /// Converts `Zatoshi` to `ZEC`
+    var asZec: ZEC {
+        ZEC(value: Double(amount) / Double(100_000_000))
+    }
+    
+    /// The value of `Zatoshi` is valid when zero or positive number. User can't send or receive negative amount of `Zatoshi`.
+    /// Negative `Zatoshi` makes sense only as some partial result, generally speaking during some computations.
+    /// Another example would be summing up last 10 transactions resulting in negative `Zatoshi` value.
+    var isValid: Bool {
```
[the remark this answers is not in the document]
LukasKorba: removed

secant/Utils/Zatoshi.swift
LukasKorba: refactored

secant/Utils/Zatoshi.swift
LukasKorba: removed

secant/Utils/Zatoshi.swift
LukasKorba: moved to Constants

secant/Utils/Zatoshi.swift
LukasKorba: reimplemented, the ZEC is no more, we always work with Zatoshi and only when we want to display the value we "convert" it using .asHumanReadableZec (returning String)

secant/Utils/Zatoshi.swift
LukasKorba: removed

secant/Utils/Zatoshi.swift
LukasKorba: bankers it is

secant/Utils/Zatoshi.swift
LukasKorba: ZEC removed completely

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,68 @@
...
+import Foundation
+
+struct Zatoshi {
+    enum Constants {
+        static let oneZecInZatoshi: Int64 = 100_000_000
+        static let maxZecs: Int64 = 21_000_000
+        static let maxZatoshi: Int64 = Constants.oneZecInZatoshi * Constants.maxZecs
+    }
+    
+    static var nsDecimalHandler = NSDecimalNumberHandler(
```
ccjernigan, line 17: How does this work within iOS when it comes to locale changes?

Is this instance frozen with the locale at the time of construction, or does it pick up the current locale during the conversion?

In Java, we construct the formatter with a specific locale and have to create a new instance when the locale changes.

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,68 @@
...
+//
+//  Created by Lukáš Korba on 26.05.2022.
+//
+
+import Foundation
+
+struct Zatoshi {
+    enum Constants {
+        static let oneZecInZatoshi: Int64 = 100_000_000
+        static let maxZecs: Int64 = 21_000_000
```
pacu, line 13:
```suggestion
        static let maxZecSupply: Int64 = 21_000_000
```

secant/Utils/Zatoshi.swift
ccjernigan: Also are there currency specific formatting APIs?

In some countries, I think they use different separators for numbers versus currency.  In Java, we're doing something like this: https://github.com/zcash/secant-android-wallet/blob/94d259d5b0c0175c40dad8b89940bcb4b4e48691/sdk-ext-ui-lib/src/main/java/cash/z/ecc/sdk/ext/ui/model/ZecString.kt#L44

I'm not 100% certain about this, but the fact that there's an explicit `monetaryDecimalSeparator` API in Java indicates it is a consideration.

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /// Converts `Decimal` to `Zatoshi`
+    static func from(zec: Decimal) -> Zatoshi {
+        let zec2zatoshi = Decimal(Constants.oneZecInZatoshi) * zec
+        return Zatoshi(amount: NSDecimalNumber(decimal: zec2zatoshi).int64Value)
+    }
+
+    /// Converts `String` to `Zatoshi`
+    static func from(zec: String, formatter: NumberFormatter) -> Zatoshi? {
```
ccjernigan, line 46: Who creates the `NumberFormatter`?  It seems like clients should only have a little bit of influence on the parsing.

On the Android side, we don't give clients full control over the formatting but we did need a way to cache the formatting objects a bit to ensure that multiple method calls could use the same objects.  (A concern is that system locale can change at any time, so we don't want to keep asking for locale when making several API calls in a row but rather ask once during a composition of the UI).

ccjernigan (commented):
The changes from the last review look great, thanks for the effort that went into this iteration.

Now we're onto some localization-specific questions.  Again I don't know the Swift APIs as well, but hopefully we can make sure we get the localization correct too.

secant/Utils/Zatoshi.swift
pacu, line 17:
```suggestion
    static var decimalHandler = NSDecimalNumberHandler(
```

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,68 @@
...
+        raiseOnOverflow: true,
+        raiseOnUnderflow: true,
+        raiseOnDivideByZero: true
+    )
+    
+    @Clamped(-Constants.maxZatoshi...Constants.maxZatoshi)
+    var amount: Int64 = 0
+
+    /// Converts `Zatoshi` to `NSDecimalNumber`
+    var asZec: NSDecimalNumber {
```
pacu, line 30: Since ZEC is not explicitly in the API I suggest we change it to something more descriptive

```suggestion
    var decimalValue: NSDecimalNumber {
```

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,68 @@
...
+        NSDecimalNumber(decimal: Decimal(amount) / Decimal(Constants.oneZecInZatoshi))
+    }
+
+    /// Converts `Zatoshi` to human readable format, up to 8 fraction digits
+    var asHumanReadableZec: String {
```
pacu, line 35: Same idea as above
```suggestion
    var decimalString: String {
```

secant/Utils/Zatoshi.swift
```diff
@@ -0,0 +1,68 @@
...
+        asZec.rounding(accordingToBehavior: Zatoshi.nsDecimalHandler).stringValue
+    }
+
+    /// Converts `Decimal` to `Zatoshi`
+    static func from(zec: Decimal) -> Zatoshi {
```
pacu, line 40: Should this receive an NSDecimal or are they Bridged values types?

secant/Utils/Zatoshi.swift
pacu, line 47:
```suggestion
    static func from(decimalString: String, formatter: NumberFormatter) -> Zatoshi? {
```

pacu (changes requested):
Great work! suggested some changes

secant/Utils/Zatoshi.swift
LukasKorba: The reducer (TCA). This function is pure and independent, the formatter is going to be injected (#312)

secant/Utils/Zatoshi.swift
LukasKorba: updated

secant/Utils/Zatoshi.swift
LukasKorba: updated

secant/Utils/Zatoshi.swift
LukasKorba: updated

secant/Utils/Zatoshi.swift
LukasKorba: updated

secant/Utils/Zatoshi.swift
LukasKorba: NSDecimal is bridged to Decimal just like NSString is bridged to String

secant/Utils/Zatoshi.swift
ccjernigan: Imagine this code ends up moving into the SDK (or a helper library of the SDK).

Would you want a wrapper class so that only the SDK can generate the number formatter?  It can still be constructed by the UI and injected in, but this ensures that clients of the API can't create arbitrary number formatters.

e.g. this sort of pseudo-swift example
```
struct ZecFormatter {
    let numberFormatter: NumberFormatter
    constructor ZecFormatter() {
           var formatter = NumberFormatter()
            formatter.maximumFractionDigits = 8
            formatter.maximumIntegerDigits = 8
            formatter.numberStyle = .decimal
            formatter.usesGroupingSeparator = true
        
            numberFormatter = formatter
    }
}
```

secant/Utils/Zatoshi.swift
LukasKorba: `NSDecimalNumberHandler` is not a formatter. The NSDecimalNumberHandler only defines the strategy for the rounding, fully respecting the locale. 

The NumberFormatter though is used to handle locale formats, floating point separator, currency symbols, etc. In iOS case we are lucky the NumberFormatter is fully automatic and can be used instead of regex you use on Android side. 

So to answer your question, any change in the locale even during runtime is handled by the system, the strategy for rounding stays the same.

ccjernigan: Awesome, thanks for the clarification!

secant/Utils/Zatoshi.swift
ccjernigan: @LukasKorba Is this fully resolved?

We will be moving Zatoshi into the SDK for sure in the future, since it is a key public API.

We may or may not move the Zec formatting.  (Zec formatting could be turned into extension functions on Zatoshi that live in the app if we'd like).  But if we do move the Zec formatting into the SDK, I feel like the SDK should have a mechanism to control generation of the formatter.  So this is one item for consideration.  We can defer it until #312 but I don't think NumberFormatter is the right API for `static func from(zec: String, formatter: NumberFormatter)`

pacu: This is a great point. 

Maybe if we move this to the SDK we can use a default value for the formatter parameter.

I think that most of the multi coin wallets we want to aim for as potential users must have their own number formatters

LukasKorba: As long as we want to create an API that is flexible enough to inject formatter, the location doesn't matter. Either in the wallet or SDK. For either of those we will create a WrappedNumberFormatter and inject it when using the API `static func from(zec: String, formatter: NumberFormatter)`. So the SDK can be in control of let's say default SDK formatter unless anybody want's to override it.

Inflexible solution would implement number formatter being part of the Zatoshi type and never let anybody to control it. I hope it answered your point, otherwise I'm not sure if I understand the concern.

pacu: > As long as we want to create an API that is flexible enough to inject formatter, the location doesn't matter. Either in the wallet or SDK. For either of those we will create a WrappedNumberFormatter and inject it when using the API `static func from(zec: String, formatter: NumberFormatter)`. So the SDK can be in control of let's say default SDK formatter unless anybody want's to override it.
> 
> Inflexible solution would implement number formatter being part of the Zatoshi type and never let anybody to control it. I hope it answered your point, otherwise I'm not sure if I understand the concern.

I'm ok with this. I think the idea is to make a clear statement about what the default is and whomever uses the injection does it at their own risk :)

ccjernigan: Yes, Pacu has it.  I just want it to be super clear what we typically expect clients to pass in here and deviating from that is at their own risk.

pacu: @LukasKorba let's try to add the default value and make the API really clear. 

Excellent work gentlemen

pacu (approved):
Approved w/ comments

secant/Utils/Zatoshi.swift
LukasKorba: I think we are all on the same boat. My plan was to:
1. Implement our (Zcash) formatter and let it be in the SDK - the default one
2. Move Zatoshi to the SDK as well
3. Create WrappedNumberFormatter, its live instance is going to use the one from the SDK
4. TCA will inject WrappedNumberFormatter dependency (live one)

The wallet will never define another NumberFormatter, users of SDK can do whatever they want

PR #403, opened 2022-08-06
Improve Profile feature reusability by applying Interface Segregation Principle

ibru:
This PR improves robustness of the code by applying ISP at one place inside `ProfileEnvironment`.

## Summary
Recently I've come across your code and spotted some things I thought I could bring to the light.
I know you accept only PRs fixing some existing issue. If you think this PR is not relevant, feel free to abandon it.

In Profile feature, I've spotted some places violating [ISP principle](https://en.wikipedia.org/wiki/Interface_segregation_principle) (one of the SOLID principles) and thought I might give a suggestion for improvement in for of Pull request.

## Problem

The `ProfileEnvironment` is strongly coupled to whole `WrappedSDKSynchronizer` even though it actually needs to use only one of its methods.
This strong coupling makes the code more difficult to reuse and change in the future.

## Solution

Inside the `ProfileEnvironment` we define only what it really needs - a method that returns Shielded Address.

I've solved it by introducing `typealias ShieldedAddressSource`, representing only functionality of getting the address. And replaced it with `WrappedSDKSynchronizer` dependency.

## Benefits of ISP
- Not being dependent on changes to `WrappedSDKSynchronizer` - before, if `WrappedSDKSynchronizer` would change, it would require to change (or at least recompile) Profile feature too. It's not the case any more

- Flexible for future changes - Imagine a situation that we'd want to compose the address from `WrappedSDKSynchronizer` plus some other component, making operation on it, before we display it inside Profile. In current situation we'd have to add new dependency to `ProfileEnvironment` for the other component, and do the logic inside Profile. With this suggested solution, the logic could be performed outside of Profile (perhaps reused for other places) and result passed to `ProfileEnvironment`, without any changes to Profile feature itself.

- Profile Feature is now easier to reason about, because its' Enironment has now more straightforward dependencies.

- Avoiding situation that someone would accidentally call methods of `WrappedSDKSynchronizer` that are not intended to be called inside Profile feature.

- When removed this strong coupling, it makes it much easier to split code to separate packages / frameworks

- Easier unit testing - no need to mock stuff which is not used.

---

# Author

- [x] Documentation: Did you update Docs as appropiate? (E.g [README.md](../blob/main/README.md), etc.)
       _(Authors note: Sorry, I'm unable to run the app and verify the changes, since I have no Zcash account yet.  I'm relying on passing unit tests.)_

# Reviewer

### DIFF
--- a/secant/Features/Profile/ProfileStore.swift
+++ b/secant/Features/Profile/ProfileStore.swift
@@ -35,9 +35,11 @@ enum ProfileAction: Equatable {
 // MARK: - Environment
 
 struct ProfileEnvironment {
+    typealias ShieldedAddressSource = () -> String?
+
     let appVersionHandler: AppVersionHandler
     let mnemonic: WrappedMnemonic
-    let SDKSynchronizer: WrappedSDKSynchronizer
+    let shieldedAddress: ShieldedAddressSource
     let scheduler: AnySchedulerOf<DispatchQueue>
     let walletStorage: WrappedWalletStorage
     let zcashSDKEnvironment: ZCashSDKEnvironment
@@ -47,7 +49,7 @@ extension ProfileEnvironment {
     static let live = ProfileEnvironment(
         appVersionHandler: .live,
         mnemonic: .live,
-        SDKSynchronizer: LiveWrappedSDKSynchronizer(),
+        shieldedAddress: LiveWrappedSDKSynchronizer().getShieldedAddress,
         scheduler: DispatchQueue.main.eraseToAnyScheduler(),
         walletStorage: .live(),
         zcashSDKEnvironment: .mainnet
@@ -56,7 +58,7 @@ extension ProfileEnvironment {
     static let mock = ProfileEnvironment(
         appVersionHandler: .test,
         mnemonic: .mock,
-        SDKSynchronizer: MockWrappedSDKSynchronizer(),
+        shieldedAddress: MockWrappedSDKSynchronizer().getShieldedAddress,
         scheduler: DispatchQueue.main.eraseToAnyScheduler(),
         walletStorage: .live(),
         zcashSDKEnvironment: .testnet
@@ -77,7 +79,7 @@ extension ProfileReducer {
     private static let profileReducer = ProfileReducer { state, action, environment in
         switch action {
         case .onAppear:
-            state.address = environment.SDKSynchronizer.getShieldedAddress() ?? ""
+            state.address = environment.shieldedAddress() ?? ""
             state.appBuild = environment.appVersionHandler.appBuild()
             state.appVersion = environment.appVersionHandler.appVersion()
             state.sdkVersion = environment.zcashSDKEnvironment.sdkVersion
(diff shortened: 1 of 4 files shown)


LukasKorba:
I'm familiar with the concept of ISP but here it's not done properly.

1. True, the Profile doesn't need all the methods and could be refactored to closure only. The profile holds the whole synchronizer so it can pass it to the Settings though. 

2. You replaced `environment.SDKSynchronizer` with the new instance of `LiveWrappedSDKSynchronizer` but that one won't never work properly. We need to hold one and only instance live throughout the whole lifecycle of the wallet. New instance is not properly initialised and doing so over and over is not efficient. See the `AppStore`, the only place where synchronizer is initialised with the `Initializer`.

3. ISP could be used at many places, another example would be Mnemonic. Offering many methods yet usually used with one only and the list continues. 

Current commit is not functional and without bigger changes it won't be. I propose to create a ticket with ISP candidates and focus on it later.

secant/Features/Profile/ProfileStore.swift
```diff
@@ -115,7 +117,7 @@ extension ProfileReducer {
             SettingsEnvironment(
                 localAuthenticationHandler: .live,
                 mnemonic: environment.mnemonic,
-                SDKSynchronizer: environment.SDKSynchronizer,
+                SDKSynchronizer: LiveWrappedSDKSynchronizer(),
```
LukasKorba, line 120: This is breaking the functionality, `LiveWrappedSDKSynchronizer()` won't be properly initialised + it's potentially breaking all subscribers for the synchronizerStateChanged.

secant/Features/Profile/ProfileStore.swift
```diff
@@ -35,9 +35,11 @@ enum ProfileAction: Equatable {
 // MARK: - Environment
 
 struct ProfileEnvironment {
+    typealias ShieldedAddressSource = () -> String?
```
LukasKorba, line 38: We don't have specification for this so first it's probably good to have a conversation where to have such definitions.

ibru:
Thanks @LukasKorba for quick review.

From your comments, it sounds like I've picked a not that straightforward part of the code to refactor :)
Didn't spot that the `LiveWrappedSDKSynchronizer` cannot be instantiated more than once.

It looks like there has to happen some internal discussion in your company first, on how to solve scenarios like this. Because it would basically influence whole codebase, because as you stated in point 3, there might me more places affected by similar scenario.

I'm happy to assist in the discussions if you'd like.

The point 2 almost sounds like the `LiveWrappedSDKSynchronizer` could be set up as a singleton or perhaps better there could be a static `live` variable / method that would internally return the time same instance on each call.

But IMHO best approach would be to use Composition Root design pattern - define one place where Environments would be created and in that place it would be easier to control which instances are passed where. And it could also solve the issues that some Environments contain dependencies only to pass them to another environments.
However, this Composition Root pattern would require more significant changes and had to be adopted gradually over long time.

LukasKorba:
I created a specific issue for the ISP #405, closing for now till we focus on the topic and make some decisions how to approach it.

Issue #1071, opened 2024-02-19
False positive keysMissing edge cases ultimate solution


LukasKorba:
This is a follow up ticket for #1068. 

What's needed:
1. cleanup of code, retry logic for keysMissing presented in #1000 is no longer valid or meaningful
2. all possible flags for the keychain leave the door open for some edge case (usually very rare but possible). So under some very specific circumstances, false positive case is still true. 

some solutions for the 2nd
- Kris mentioned to rework the SDK a bit to not need the seed for the sync, only for the initial setup of the SDK when the DB is created and then only for some specific cases when migration of the DB is needed
- To distinguish between app runs when a user operates with it and BGTask operates it. When BGTask awakes the app and TCA part of Zashi knows about it, it can behave differently, not show the alert and mark some flag. Ideally the initialization pipeline marks some flag as "ok, the seed can't be read so I stop here and wait until the zashi is opened to recover from this state".

PR #1073, opened 2024-02-19
[#1068] False positive keysMissing detection

### LINKED ISSUE #1068: False positive keysMissing detection
### Long story short
Josh & Pacu experienced this error state to be prompted even though the next start of Zashi resolved the issue. Unfortunately, there are no steps to reproduce.

### Sorry, the long story
I put a lot of thoughts into this and here is my assumption of what might be going on and what the implications are.

There are 2 scenarios:
1. The keys are truly missing, typically after iPhone restore
2. The keys are present but Zashi ends up in a conclusion the keys are missing (the next of this ticket explains why it might be the case)

First of all, this is a serious issue that has direct implication on #1024. @true-jared, I'm really worried about this because whenever this happens and it's a false positive we prompt them with message that is not true and even worse, we try to pull them into a flow of seed validation.

So the way I see it we must resolve this issue first before we can focus on a valid case when the keys are truly missing.

#### The seed rules Zashi
With seed, Zashi works and can work. Where is no seed, there is no working Zashi. This is very simple fact. Zashi tries to get the seed from the Keychain as soon as possible, ideally right after the start of Zashi.

#### The keychain
A few facts:
- The seed is stored in the keychain of an iPhone, nowhere else. The keychain API is probably the oldest API in iOS SDK, it doesn't support async await syntax whatsoever, it's very low level old way of coding. 
- The keychain is a system "storage", so it lives outside of Zashi. We can't control it, we can't demand a priority read or write, anything like it. So if we try to read the values from it, there are no guarantees it'll work at that very exact moment.

Especially the last point hit me and basically created a space for speculation(s). What if keychain is locked at the moment we want to read it? With async syntax, the system would wait and provide the value(s) asap. The API is not async one though so it either can provide the value or not. If not, it fails, simple as that.

The assumption is the keychain uses either a lock inside or operates with an operation queue so it should always return value. Yet there are cases when the retrival of the seed fails. We need to understand why.

#### The options we have
Since the API is something we can't fix or improve, we need to do some precautions. I implemented some retry simple [logic](https://github.com/Electric-Coin-Company/zashi-ios/pull/1000) that is apparently not enough and hasn't solved the issue. 

We could try to be more aggressive, increase the number of retry attempts, wait more time. This is something we will probably need to do anyway.

Another option is to implement a prompt with a spinner that says something like `syncing with the keychain...`, that will block users from accessing Zashi and periodically tries to read the values from the keychain until it gets the seed. The assumption here is that it must work and at some point it'll pass and unblocks users from accessing Zashi. We should try this as a prototype and internal solution so we can get some experience and answers.

My biggest worry now is that there might be some bug on iOS side and only new start of Zashi gets job done. This is also and still an option. Maybe Josh's and Pacu's run of Zashi simply failed because the API or system failed and fresh start was basically a restart of a computer (you know those cases when a restart is the only option). What I mean by this is maybe even a periodical try won't work until Zashi is restarted. We don't know that yet until we try to implement proposed prototype.

LukasKorba:
Closes #1068 

- Change of the setup of the keychain massively reduces the probability and frequency of cases when the keychain is required to provide data. Background task for the sync was the main producer of these attempts to read the data, the new flag allows BGTask to operate freely with no errors related to the keychain. Ultimate solution how to make it even better is covered in #1071, TODO added to the code.

# Author

# Reviewer

### DIFF
--- a/modules/Sources/Dependencies/WalletStorage/WalletStorage.swift
+++ b/modules/Sources/Dependencies/WalletStorage/WalletStorage.swift
@@ -166,11 +166,13 @@ public struct WalletStorage {
             kSecAttrService as String: (zcashStoredWalletPrefix + forKey) as AnyObject,
             kSecAttrAccount as String: account as AnyObject,
             kSecClass as String: kSecClassGenericPassword,
-            /// The data in the keychain item can be accessed only while the device is unlocked by the user.
-            /// This is recommended for items that need to be accessible only while the application is in the foreground.
+            /// The data in the keychain item can be accessed only after the device has been unlocked by the user
+            /// (aka won't be accessible after restart of the device until unlock).
             /// Items with this attribute do not migrate to a new device.
             /// Thus, after restoring from a backup of a different device, these items will not be present.
-            kSecAttrAccessible as String: kSecAttrAccessibleWhenUnlockedThisDeviceOnly
+            // TODO: [#1071] ultimate solution to handle background sync and the keychain setup
+            // https://github.com/Electric-Coin-Company/zashi-ios/issues/1071
+            kSecAttrAccessible as String: kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly
         ]
         
         return query


pacu (approved):

nuttycom (approved):
utACK. Please ensure that the Slack discussion we had about how to correctly handle this situation is reflected in the DAG.

PR #1075, opened 2024-02-19
[#1060] Memo field behaviours for transparent address

### LINKED ISSUE #1060: Memo field behaviours for transparent address
Send memo to the transparent address is not allowed and the memo field should grey-out. See the figma

https://www.figma.com/file/vYQD6uZtkPgtmgJ3zfauq0/Zashi-UI?type=design&node-id=2030-16345&mode=design&t=IBEE7tnzlc2jvGQY-0

This worked in the past but newer iOS overlays the field with white box.

Simulator Screenshot - iPhone 15 Pro - 2024-02-14 at 14 02 04

Another question and a case we never solved on a product management level is what should happen when the memo holds some text and only after that a user uses some transparent address. Should we remove and leave the text inside the memo and all be greyed-out? @true-jared

LukasKorba:
Closes #1060 

- MessageEditor updated to present disabled state in a different way
- Text in the memo is not persisted when the transparent address is used and MessageBox disabled
- New unit tests implemented and many more updated

Simulator Screenshot - iPhone 15 Pro - 2024-02-19 at 16 57 36

# Author

# Reviewer

### DIFF
--- a/modules/Sources/Features/SendFlow/SendFlowStore.swift
+++ b/modules/Sources/Features/SendFlow/SendFlowStore.swift
@@ -90,7 +90,6 @@ public struct SendFlowReducer: Reducer {
         }
         
         public var isMemoInputEnabled: Bool {
-            transactionAddressInputState.textFieldState.text.data.isEmpty ||
             !transactionAddressInputState.isValidTransparentAddress
         }
         
@@ -256,6 +255,12 @@ public struct SendFlowReducer: Reducer {
             case .transactionAmountInput:
                 return .none
 
+            case .transactionAddressInput(.textField):
+                if !state.isMemoInputEnabled {
+                    state.memoState.text = "".redacted
+                }
+                return .none
+                
             case .transactionAddressInput(.scanQR):
                 return Effect.send(.updateDestination(.scanQR))
 
--- a/modules/Sources/Features/SendFlow/SendFlowView.swift
+++ b/modules/Sources/Features/SendFlow/SendFlowView.swift
@@ -99,7 +99,7 @@ public struct SendFlowView: View {
                             .padding(.top, 40)
                             
                             MessageEditor(store: store.memoStore())
-                                .frame(height: 140)
+                                .frame(height: 175)
                                 .disabled(!viewStore.isMemoInputEnabled)
                                 .focused($isMessageFocused)
                                 .toolbar {
--- a/modules/Sources/UIComponents/TextFields/MessageEditor/MessageEditor.swift
+++ b/modules/Sources/UIComponents/TextFields/MessageEditor/MessageEditor.swift
@@ -37,50 +37,51 @@ public struct MessageEditor: View {
                     Spacer()
                 }
                 .padding(.bottom, 2)
-                
-                TextEditor(text: $message)
-                    .focused($isFocused)
-                    .padding(2)
-                    .font(.custom(FontFamily.Inter.regular.name, size: 14))
-                    .messageShape(
-                        filled: isEnabled
-                        ? nil
-                        : Asset.Colors.shade72.color
-                    )
-                    .overlay {
-                        if message.isEmpty || !isEnabled {
-                            HStack {
-                                VStack {
-                                    Text(L10n.Send.memoPlaceholder)
-                                        .font(.custom(FontFamily.Inter.regular.name, size: 13))
-                                        .foregroundColor(Asset.Colors.shade72.color)
-                                        .onTapGesture {
-                                            isFocused = true
-                                        }
+
+                if !isEnabled {
+                    Asset.Colors.shade72.color
+                        .messageShape(filled: Asset.Colors.shade72.color)
+                } else {
+                    TextEditor(text: $message)
+                        .focused($isFocused)
+                        .padding(2)
+                        .font(.custom(FontFamily.Inter.regular.name, size: 14))
+                        .messageShape(filled: nil)
+                        .overlay {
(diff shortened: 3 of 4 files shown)


modules/Sources/Features/SendFlow/SendFlowStore.swift
```diff
@@ -79,7 +79,8 @@ public struct SendFlowReducer: Reducer {
         public var isValidForm: Bool {
             @Dependency(\.zcashSDKEnvironment) var zcashSDKEnvironment
 
-            return transactionAmountInputState.amount.data > zcashSDKEnvironment.network.constants.defaultFee().amount
```
str4d, line 82: This validity check was confirming that the user was not sending a transaction worth less than the fee. Was this removed to enable users to send zero-value shielded transactions (i.e. paying a fee to send just a memo to a recipient)? I don't actually know what receiving wallets will do in this case, because zero-value shielded outputs are dummy notes, and I would not be surprised if they get filtered out somewhere.

In any case, it's unclear to me why this change was made in a PR about the memo field behaviour for transparent addresses.

modules/Sources/Features/SendFlow/SendFlowStore.swift
```diff
@@ -90,7 +91,10 @@ public struct SendFlowReducer: Reducer {
         }
         
         public var isMemoInputEnabled: Bool {
-            transactionAddressInputState.textFieldState.text.data.isEmpty ||
+            print(transactionAddressInputState.textFieldState.text.data.isEmpty ||
+                  !transactionAddressInputState.isValidTransparentAddress)
+            
+            return transactionAddressInputState.textFieldState.text.data.isEmpty ||
```
str4d, line 97: This looks like a local debugging change that should be undone:
```suggestion
            transactionAddressInputState.textFieldState.text.data.isEmpty ||
```

modules/Sources/UIComponents/TextFields/TCATextField/TCATextFieldStore.swift
```diff
@@ -16,9 +16,14 @@ public struct TCATextFieldReducer: Reducer {
         public var text = "".redacted
         public var valid = false
 
-        public init(validationType: String.ValidationType?, text: RedactableString) {
+        public init(
+            validationType: String.ValidationType?,
+            text: RedactableString,
+            valid: Bool = true
```
str4d, line 22: I don't understand why the default initialization of this field (which I presume is the cached "field is valid" state from the UI) is being changed to `true`, or why that is happening here instead of line 17.

modules/Sources/Features/SendFlow/SendFlowStore.swift
```diff
@@ -79,7 +79,8 @@ public struct SendFlowReducer: Reducer {
         public var isValidForm: Bool {
             @Dependency(\.zcashSDKEnvironment) var zcashSDKEnvironment
 
-            return transactionAmountInputState.amount.data > zcashSDKEnvironment.network.constants.defaultFee().amount
+            return transactionAmountInputState.amount.data + zcashSDKEnvironment.network.constants.defaultFee().amount
```
nuttycom, line 82: This is a very different check, and I think maybe not the one you want. It says that the input amount plus the fee must be a valid Zcash amount; is the `maxValue` here derived from the user's balance? It also can have false positives if the ZIP 317 fee would cause the actual balance of the transaction to exceed available funds. Ultimately what is probably needed here is a way to obtain the maximum spendable value as part of the wallet state, and check against that.

modules/Sources/Features/SendFlow/SendFlowStore.swift
```diff
@@ -90,7 +91,10 @@ public struct SendFlowReducer: Reducer {
         }
         
         public var isMemoInputEnabled: Bool {
-            transactionAddressInputState.textFieldState.text.data.isEmpty ||
+            print(transactionAddressInputState.textFieldState.text.data.isEmpty ||
```
nuttycom, line 94: Is this a debug print that is inadvertently left in place?

str4d (approved):
utACK 756bffaeb01dbd36ea83381ffb84caa15816de0e with a cleanup and some questions.

nuttycom (dismissed):
I don't think that the max value check here is correct, and it may produce misleading results if someone is trying to send close to their full wallet value to a recipient.

modules/Sources/UIComponents/TextFields/TCATextField/TCATextFieldStore.swift
LukasKorba: Good catch, I needed to expose `valid` variable for tests and by default it must be false. Fixed

modules/Sources/Features/SendFlow/SendFlowStore.swift
LukasKorba: Yes, debug, removed.

modules/Sources/Features/SendFlow/SendFlowStore.swift
LukasKorba: I actually reverted the code because any change here is put of the scope of this ticket. If any change is needed we will create a follow up ticket.

modules/Sources/Features/SendFlow/SendFlowStore.swift
LukasKorba: I actually reverted the code because any change here is put of the scope of this ticket. If any change is needed we will create a follow up ticket.

daira (approved):
utACK

PR #1117, opened 2024-03-11
[#1092] Adopt proposal API

### LINKED ISSUE #1092: Adopt proposal API
Adopt this [PR](https://github.com/Electric-Coin-Company/zcash-swift-wallet-sdk/pull/1382)

LukasKorba:
Closes #1092 

- TCA sdkSynchronizer dependency extended with 3 new Proposal APIs
- `Typical Fee < 0.001` localized and updated in the UI
- Awaiting all transaction results with use of new proposal.transactionCount() method
- Implemented new PartialProposalError reducer and view
- Contact support mail with transaction IDs logic implemented
- Fallback to share logic implemented
- `PPE` integrated into SendFlow
- `PPE`  integrated into Shielding
- Changelog updated

IMG_BA548C067207-1
IMG_6997C261A6F8-12

# Author

# Reviewer


daira:
For ZIP 320 support, the case where only the first transaction succeeds is essentially identical to the case where the exchange sends back the funds: both leave the funds in the ephemeral t-addr. So why can't we just handle this correctly?

str4d:
> For ZIP 320 support, the case where only the first transaction succeeds is essentially identical to the case where the exchange sends back the funds: both leave the funds in the ephemeral t-addr.

For ZIP 320 support, from a control of funds position, yes. But they are semantically two very different cases, and we have not done the UX work to distinguish them, or handle how the resulting transaction history is conveyed to users.

> I was under the impression that handling the latter was required for Zashi 1.0.

I was not: my understanding is that we need to be able to _send_ to TEX addresses (which covers the majority use case), but not that we need to handle exchanges sending funds back (which is an edge case).

I think we should handle both exchanges sending funds back, and transaction submission failures, in-app eventually. But I also think that handling this properly with good UX requires more engineering than we have time for in the 1.0 timeline, so it would go in a post-1.0 release. Having an error indicating to users when they have hit this edge case is the minimal amount of engineering we can do here, while avoiding those edge case users believing they have lost funds.

> So why can't we handle the former correctly?

There are at least four different states we could be in here that I can think of (for ZIP 320 proposals; more if we also consider arbitrary proposals which the SDK might generate or need to handle in future):

- Transient failure to submit the original transactions; logical transaction is unfinished.
- Permanent failure to submit the original transactions; logical transaction is unfinished.
- Logical transaction is rejected; funds sent back.
- Drive-by funding: someone saw the ephemeral address and sent a (maybe intentionally identifiable) amount of funds to it.

Each state needs to be handled differently; some states have multiple possible resolutions. None of this has been discussed or planned yet, and I don't think we have time to do so for 1.0.

daira:
I completely agree that an ideal UX would distinguish those (except possibly that the refund and drive-by cases might be indistinguishable).

Where I disagree is that I don't think any of them need to result in *any* possibility of funds loss. Just allocate the ephemeral address in a way that guarantees that we scan it.

modules/Sources/Dependencies/SupportDataGenerator/SupportDataGenerator.swift
```diff
@@ -43,6 +44,31 @@ public enum SupportDataGenerator {
...
+        let idsString = txIds.map { id in
+            "- ID: \(id)"
+        }
+        .joined(separator: "\n")
+
+        let message =
+        """
+        Hi Zashi Team,
+
+        While sending a transaction, I encountered an error state and I'm reaching out to get help with recovering my funds.
```
daira, line 60: As noted in the conversation on the PR, I disagree that zcash_client_backend should implement ZIP 320 in a way that leads to risk of funds loss or a user ever needing to do something special to recover funds. It is not necessary.

modules/Sources/Features/BalanceBreakdown/BalanceBreakdownView.swift
```diff
@@ -221,9 +228,9 @@ extension BalanceBreakdownView {
                 shadowOffset: 6
             )
             .padding(.bottom, 15)
-            .disabled(!viewStore.isShieldableBalanceAvailable || viewStore.isShieldingFunds)
+            //.disabled(!viewStore.isShieldableBalanceAvailable || viewStore.isShieldingFunds)
```
daira, line 231: Is this temporary? If it is then add a TODO and file an issue, otherwise delete the commented-out code.

modules/Sources/Features/SendFlow/SendFlowView.swift
```diff
@@ -124,7 +125,7 @@ public struct SendFlowView: View {
                                 Text(L10n.Send.review.uppercased())
                             }
                             .zcashStyle()
-                            .disabled(!viewStore.isValidForm || viewStore.isSending)
+                            //.disabled(!viewStore.isValidForm || viewStore.isSending)
```
daira, line 128: Either add a TODO and file an issue, or delete the commented-out code.

modules/Sources/Generated/L10n.swift
```diff
@@ -289,6 +285,18 @@ public enum L10n {
...
   }
+  public enum ProposalPartial {
+    /// Contact Support
+    public static let contactSupport = L10n.tr("Localizable", "proposalPartial.contactSupport", fallback: "Contact Support")
+    /// Sending to this recipient required multiple transactions but only some of them succeeded.
+    public static let message1 = L10n.tr("Localizable", "proposalPartial.message1", fallback: "Sending to this recipient required multiple transactions but only some of them succeeded.")
+    /// Your funds are safe but need to be recovered with assistance from our side.
+    public static let message2 = L10n.tr("Localizable", "proposalPartial.message2", fallback: "Your funds are safe but need to be recovered with assistance from our side.")
+    /// Please use the button below to contact us. It automatically prepares all the data we need to help you recover your funds.
+    public static let message3 = L10n.tr("Localizable", "proposalPartial.message3", fallback: "Please use the button below to contact us. It automatically prepares all the data we need to help you recover your funds.")
```
daira, line 296: And this.

daira (approved):
utACK modulo text changes.

str4d:
> Where I disagree is that I don't think any of them need to result in _any_ possibility of funds loss.

I agree, and am not arguing for this!

> Just allocate the ephemeral address in a way that guarantees that we scan it.

I agree with "Just allocate the ephemeral address in a way that guarantees that we _**can**_ scan it" - this is sufficient to guarantee no loss of funds. I disagree with actually scanning it in 1.0, because per my comment above that would require UX work to explain to the user what is going on in all of the different states that scan might occur from, that we do not have time for.

Only one of the four states I outlined is transient (it decays to the second state after about an hour), but three of the states are permanent, distinguishable, and "automatically re-shield" is not necessarily the correct resolution to all three.

For 1.0, informing the user and telling them to contact us means that a) we don't make permanent changes to their wallet history that might be undesirable, and b) means that we learn about these edge cases if they occur, which helps us to craft better UX here.

modules/Sources/Features/SendFlow/SendFlowStore.swift
```diff
@@ -199,63 +212,94 @@ public struct SendFlowReducer: Reducer {
...
+                        await send(.sendFailed(error.toZcashError()))
+                    }
+                }
 
             case .sendPressed:
+                return .send(.sendPartial(["ba5113d9e78c885bdb03d44f784dc35aa0822c0f212f364b5ffc994a3e219d1f", "asdfasfqerqrafdb03d44f784dc35aa0822c0f212f364b5ffc99werwrt5wewerqwtr"]))
+                state.isSending = true
+
+                guard let proposal = state.proposal else {
+                    return .send(.sendFailed("missing proposal".toZcashError()))
```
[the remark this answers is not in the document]
LukasKorba: In theory - never but I'm still handling this scenario just in case. 

The thing is I need to pass a Proposal in the TCA (composable architecture) and the initial state is nil (Proposal?). Once the proposal is available, I pass it in TCA actions but the state.proposal is still an Optional type. I need to unwrap it before it can be used in send action and that is exactly what's going on here. Guard is unwrapping it but if that fails or proposal is nil, something must happen other than `return .none` side effect. That would mean UX when nothing happens when a user taps on send button.

modules/Sources/Features/BalanceBreakdown/BalanceBreakdownView.swift
LukasKorba: It was temporary, disable reactivated for the final PR

modules/Sources/Features/SendFlow/SendFlowView.swift
LukasKorba: It was temporary, disable reactivated for the final PR

secantTests/BalanceBreakdownTests/BalanceBreakdownTests.swift
```diff
@@ -28,7 +28,9 @@ class BalanceBreakdownTests: XCTestCase {
         store.dependencies.mainQueue = .immediate
         store.dependencies.numberFormatter = .noOp
         
-        await store.send(.onAppear)
+        await store.send(.onAppear) { state in
+            state.autoShieldingThreshold = Zatoshi(100_000)
```
HonzaR, line 32: We're using the shielding threshold across several files. Could we move it into a constant?

secantTests/ZatoshiStringRepresentation/ZatoshiStringRepresentationTests.swift
```diff
@@ -28,63 +28,63 @@ final class ZatoshiStringRepresentationTests: XCTestCase {
         
         XCTAssertEqual(zatoshiStringRepresentation.mostSignificantDigits, "0.000")
         XCTAssertEqual(zatoshiStringRepresentation.leastSignificantDigits, "")
-        XCTAssertEqual(zatoshiStringRepresentation.feeFormat, "< 0.001")
+        XCTAssertEqual(zatoshiStringRepresentation.feeFormat, "(Typical Fee < 0.001)")
```
HonzaR, line 31: Is there a way to obtain the text from string resources so we don't need to change all its occurrences in tests once the string resource is changed?

HonzaR (approved):
Approving with two non-blocking questions inline. Thank you.

secantTests/BalanceBreakdownTests/BalanceBreakdownTests.swift
LukasKorba: Updated

secantTests/ZatoshiStringRepresentation/ZatoshiStringRepresentationTests.swift
LukasKorba: Addressed, there's only 1 instance from now on, once we will support different languages, this Sting will be localized.

PR #1383, opened 2024-10-22
Address book encryption

str4d:
Blocked on changes to the SDK to expose ZIP 32 Arbitrary key derivation.

# Author

# Reviewer


modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -185,20 +199,36 @@ extension AddressBookClient: DependencyKey {
...
+        let salt = SymmetricKey(size: SymmetricKeySize.bits256)
+        return try salt.withUnsafeBytes { salt in
+            let salt = Data(salt)
+            let subKey = deriveEncryptionKey(addressBookKey: encryptionKey, salt: salt)
+
+            // Encrypt the serialized address book.
+            // CryptoKit encodes the SealedBox as `nonce || ciphertext || tag`.
+            let sealed = try ChaChaPoly.seal(data, using: subKey)
+
+            // Prepand the salt to the SealedBox so we can re-derive the sub-key.
```
daira, line 212:
```suggestion
            // Prepend the salt to the SealedBox so we can re-derive the sub-key.
```

modules/Sources/Models/AddressBookEncryptionKeys.swift
```diff
@@ -6,18 +6,27 @@
...
+    public init(key: SymmetricKey?) {
         self.key = key
     }
+    
+    public init(from decoder: any Decoder) throws {
+        self.key = nil
+    }
+    
+    public func encode(to encoder: any Encoder) throws {
+
```
daira, line 24:
```suggestion
        // encryption keys should not be encoded
```

daira (approved):
Looks good so far.

modules/Sources/Models/AddressBookEncryptionKeys.swift
str4d: We need to be able to encode the key in some fashion in order to cache it separately from the seed. IDK if this is actually where we want to encode it though, per my other todo.

modules/Sources/Generated/XCAssets+Generated.swift
```diff
@@ -305,6 +306,34 @@ public extension ColorAsset.SystemColor {
   }
 }
 
+public struct DataAsset {
```
str4d, line 309: It's unclear to me what this is for @LukasKorba; it seems auto-generated but I don't see any changes that would have triggered it.

modules/Sources/Features/Root/RootInitialization.swift
```diff
@@ -296,14 +296,16 @@ extension Root {
...
-                        // try walletStorage.importAddressBookEncryptionKeys(keys)
+                        var keys = AddressBookEncryptionKeys.empty
+                        keys.cacheFor(
+                            seed: seedBytes,
+                            account: 0,
+                            network: zcashSDKEnvironment.network
+                        )
+                        // TODO: Why are the address book keys stored in wallet storage?
+                        // Doesn't that mean they require the same authentication as the
+                        // seed phrase, instead of being accessible without biometric auth?
```
daira, line 307: There could be situations where you want the history and address book to be behind biometric auth. Maybe this is something we can ask when you add the first address book entry? 

If it's configurable then there should be a single config for the history and the address book, because it's expected that the history will contain addresses that need to be looked up in the address book.

daira: Note that, at least on Android, we have to store the address book encrypted so that it is encrypted when it is backed up. And since the key derivation is from the master seed, there are only two options: 
* require user auth to unlock the address book; or
* store the *derived* key in a place that does not require user auth (we obviously do not want to store the master seed in such a place).

In the latter case, storing the ciphertext and storing the current derived key can't be atomic, because they are in two different places that can't be written together transactionally (I think). Therefore there is necessarily a corner case where the ciphertext doesn't successfully decrypt with the stored key, e.g. when the app is killed after storing one of the ciphertext and the key, and before storing the other. In that case we have to ask for auth to rederive it from the stored salt (which *is* stored atomically with the ciphertext).

daira: Having written that out, I think it might be simplest to require user auth to get the master seed for the derivation now, and defer the implementation of storing the derived key if the user configures it not to require auth to another release. The argument for that is that storing the key isn't required in order for the address book encryption to work, and we would have to handle the case where the correct key isn't stored anyway.

modules/Sources/Generated/XCAssets+Generated.swift
LukasKorba: This is ok, swiftgen has a stencil file that is checking several possible asset types and since project has now a new type of generic data, it must create DataAsset for it, same like ImageAsset, FontAsset, etc.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -0,0 +1,286 @@
...
+
+extension AddressBookClient {
+    /// Encrypts address book contacts. The structure:
+    ///     [Unencrypted data]    `encryption version`
+    ///     [Unencrypted data]    `salt`
+    ///     [Encrypted data]        `address book version`
+    ///     [Encrypted data]        `timestamp`
+    ///     [Encrypted data]        `contacts`
+    ///
+    /// This method always produces the latest structure with he latest encryption version.
```
daira, line 23:
```suggestion
    /// This method always produces the latest structure with the latest
    /// encryption version.
```

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -0,0 +1,286 @@
...
+        @Dependency(\.walletStorage) var walletStorage
+        
+        guard let encryptionKeys = try? walletStorage.exportAddressBookEncryptionKeys(), let addressBookKey = encryptionKeys.getCached(account: 0) else {
+            throw AddressBookClient.AddressBookClientError.missingEncryptionKey
+        }
+        
+        var offset = 0
+        
+        // Deserialize `encryption version`
+        let encryptionVersionBytes = encryptedData.subdata(in: offset..<(offset + MemoryLayout<Int>.size))
```
daira, line 87: `subdata(in:)` appears in practice to be memory-unsafe (https://www.programiz.com/online-compiler/0eYmDmico8t92). So for example, if the length of `encryptedData` were less than 4 bytes, this would be unsafe. What I would do in Rust is write a safe helper method that wraps the unsafe one, and only use the helper.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -0,0 +1,286 @@
...
+                if let contact = AddressBookClient.deserializeContact(from: data, at: &offset) {
+                    contacts.append(contact)
+                }
+            }
+            
+            let abContacts = AddressBookContacts(
+                lastUpdated: lastUpdated,
+                version: abVersion,
+                contacts: IdentifiedArrayOf(uniqueElements: contacts)
+            )
```
daira, line 137: This duplicates `contactsFrom(unencryptedData: Data)`. It should just call it.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -34,304 +39,208 @@ extension AddressBookClient: DependencyKey {
...
+                        let contacts = try AddressBookClient.contactsFrom(encryptedData: contactsData)
+
+                        latestKnownContacts = contacts
+                        return contacts
+                    } else {
+                        // Fallback to the unencrypted file check and resolution
+                        let unencryptedFileURL = documentsDirectory.appendingPathComponent(Constants.unencryptedFilename)
+                        
+                        if let contactsData = try? Data(contentsOf: unencryptedFileURL) {
+                            // file exists, ensure data are parsed, re-saved and file deteled
```
daira, line 72:
```suggestion
                            // file exists; ensure data are parsed, re-saved and file deleted
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -34,304 +39,208 @@ extension AddressBookClient: DependencyKey {
...
 
-                    return decryptedContacts
+                    if let contactsData = try? Data(contentsOf: encryptedFileURL) {
+                        // file exists, try to find the unencrypted file in case the delete failed
+                        let unencryptedFileURL = documentsDirectory.appendingPathComponent(Constants.unencryptedFilename)
+                        if FileManager.default.fileExists(atPath: unencryptedFileURL.path) {
+                            try? FileManager.default.removeItem(at: unencryptedFileURL)
+                        }
+                        
+                        let contacts = try AddressBookClient.contactsFrom(encryptedData: contactsData)
```
daira, line 63: This doesn't follow the design:
> If the parse succeeds, also look for a file with the previous filename of the unencrypted address book format and delete it if present.

The attempt to delete the unencrypted file should occur only if the parse succeeds.
```suggestion
                        let contacts = try AddressBookClient.contactsFrom(encryptedData: contactsData)

                        // The encrypted file exists, and can be decrypted and parsed. Try to find the
                        // unencrypted file and delete it, in case an earlier delete failed.
                        let unencryptedFileURL = documentsDirectory.appendingPathComponent(Constants.unencryptedFilename)
                        if FileManager.default.fileExists(atPath: unencryptedFileURL.path) {
                            try? FileManager.default.removeItem(at: unencryptedFileURL)
                        }
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -34,304 +39,208 @@ extension AddressBookClient: DependencyKey {
...
+                        
+                        if let contactsData = try? Data(contentsOf: unencryptedFileURL) {
+                            // file exists, ensure data are parsed, re-saved and file deteled
+                            let contacts = try AddressBookClient.contactsFrom(unencryptedData: contactsData)
+
+                            // try to encrypt and store the data
+                            do {
+                                try AddressBookClient.storeContacts(contacts, remoteStorage: remoteStorage, remoteStore: false)
+                            } catch {
+                                // the store of the new file failed, skip the file remove
```
daira, line 79: This is correct. I updated the design doc to explicitly say that the deletion here should only be done if the store succeeds.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -34,304 +39,208 @@ extension AddressBookClient: DependencyKey {
...
+                        // file exists, try to find the unencrypted file in case the delete failed
+                        let unencryptedFileURL = documentsDirectory.appendingPathComponent(Constants.unencryptedFilename)
+                        if FileManager.default.fileExists(atPath: unencryptedFileURL.path) {
+                            try? FileManager.default.removeItem(at: unencryptedFileURL)
+                        }
+                        
+                        let contacts = try AddressBookClient.contactsFrom(encryptedData: contactsData)
+
+                        latestKnownContacts = contacts
+                        return contacts
```
daira, line 66: Why do we modify `latestKnownContacts` at every return, rather than having the caller set it?

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -34,304 +39,208 @@ extension AddressBookClient: DependencyKey {
...
+                            // file exists, ensure data are parsed, re-saved and file deteled
+                            let contacts = try AddressBookClient.contactsFrom(unencryptedData: contactsData)
+
+                            // try to encrypt and store the data
+                            do {
+                                try AddressBookClient.storeContacts(contacts, remoteStorage: remoteStorage, remoteStore: false)
+                            } catch {
+                                // the store of the new file failed, skip the file remove
+                                latestKnownContacts = contacts
+                                return contacts
```
daira, line 81: Shouldn't this be reported as an error?

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -34,304 +39,208 @@ extension AddressBookClient: DependencyKey {
...
-        var contacts: [Contact] = []
-        for _ in 0..<contactsCount {
-            if let contact = AddressBookClient.deserializeContact(from: data, at: &offset) {
-                contacts.append(contact)
-            }
-        }
-        
-        let abContacts = AddressBookContacts(
-            lastUpdated: lastUpdated,
+        // Merge strategy
```
daira, line 167: This function mixes the merge with the rest of the store logic. I would factor the merge out.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -34,304 +39,208 @@ extension AddressBookClient: DependencyKey {
...
-
-        // Serialize `address` (length + UTF-8 bytes)
-        let addressBytes = stringToBytes(contact.id)
-        data.append(contentsOf: intToBytes(addressBytes.count))
-        data.append(contentsOf: addressBytes)
+        remoteContacts.contacts.forEach {
+            var notFound = true
+            var indexToUpdate = -1
+            
+            for i in 0..<syncedContacts.contacts.count {
```
daira, line 178: This is $O(n \cdot n')$ where $n$ is the number of remote contacts and $n'$ is the number of synced contacts. If the contacts are sorted then it should only take $O(m \cdot \log m)$ where $m = \max(n, n')$. Swift probably has a sorted map type with a merge operation. It might be [this one](https://developer.apple.com/documentation/swift/dictionary/merge(_:uniquingkeyswith:)-m2ub).

modules/Sources/Features/AddressBook/AddressBookStore.swift
```diff
@@ -266,21 +266,21 @@ public struct AddressBook {
...
-//                    }
-//                } else {
+                if requestToSync {
+                    return .run { send in
+                        do {
+                            let syncedContacts = try await addressBook.syncContacts(abContacts)
+                            await send(.fetchedABContacts(syncedContacts, false))
+                        } catch {
+                            print("__LD syncContacts Error: \(error.localizedDescription)")
+                            // TODO: FIXME
```
daira, line 278: I know this is an existing TODO, but please file an issue and link it.

modules/Sources/Models/AddressBookEncryptionKeys.swift
```diff
@@ -6,18 +6,103 @@
...
+        guard let info = "file_identifier".data(using: .utf8) else {
+            fatalError("Unable to prepare `file_identifier` info")
+        }
+
+        // Perform HKDF with SHA-256
+        let hkdfKey = HKDF<SHA256>.deriveKey(inputKeyMaterial: key, info: info, outputByteCount: 32)
+        
+        // Convert the HKDF output to a hex string
+        let fileIdentifier = hkdfKey.withUnsafeBytes { rawBytes in
+            rawBytes.map { String(format: "%02x", $0) }.joined()
```
daira, line 102: I documented in the design doc that lowercase hex is used.

zashi-internal.entitlements
```diff
@@ -1,5 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
-<dict/>
+<dict>
+	<key>com.apple.developer.associated-domains</key>
+	<array>
+		<string>applinks:secant.flexa.link</string>
+	</array>
```
daira, line 8: Should this be in a different PR?

daira (changes requested):
Changes needed: at least the divergence from the design when attempting to delete the old plaintext file, and error handling for the case where the encrypted file cannot be stored remotely.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
LukasKorba: Updated in code directly.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
daira: This is blocking because parsing code especially shouldn't be duplicated.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -0,0 +1,286 @@
...
+        
+        guard let encryptionKeys = try? walletStorage.exportAddressBookEncryptionKeys(), let addressBookKey = encryptionKeys.getCached(account: 0) else {
+            throw AddressBookClient.AddressBookClientError.missingEncryptionKey
+        }
+        
+        var offset = 0
+        
+        // Deserialize `encryption version`
+        let encryptionVersionBytes = encryptedData.subdata(in: offset..<(offset + MemoryLayout<Int>.size))
+        offset += MemoryLayout<Int>.size
```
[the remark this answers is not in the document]
LukasKorba: If any, Int64 would be the one. Android made it as 8 bytes values, so iOS must follow 8 bytes as well. Int64 can deliver that.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -0,0 +1,286 @@
...
+        dataForEncryption.append(contentsOf: AddressBookClient.serializeDate(Date()))
+        
+        // Serialize `contacts.count`
+        dataForEncryption.append(contentsOf: intToBytes(abContacts.contacts.count))
+        
+        // Serialize `contacts`
+        abContacts.contacts.forEach { contact in
+            let serializedContact = serializeContact(contact)
+            dataForEncryption.append(serializedContact)
+        }
```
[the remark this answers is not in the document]
daira: Although not strictly blocking, it makes sense to factor this out now. The tests can be deferred.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -34,304 +39,208 @@ extension AddressBookClient: DependencyKey {
...
+        let fileURL = documentsDirectory.appendingPathComponent(filenameForEncryptedFile)
+        try encryptedContacts.write(to: fileURL)
         
-        // Deserialize `name`
-        guard let name = readString(from: data, at: &offset) else {
-            return nil
+        // store encrypted data to the remote storage
+        if remoteStore {
+            try? remoteStorage.storeAddressBookContacts(encryptedContacts, filenameForEncryptedFile)
         }
```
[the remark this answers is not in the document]
daira: Blocking: report the error.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Goal here is to get the local contacts but we don't want to read it from the file every time `allLocalContacts` is called. Instead, it's read only for the first time and next time `latestKnownContacts` delivers the contacts from in-memory. The technology behind this is the dependency in immutable so there must be a subject that can do it for us. `latestKnownContacts` is the mutable subject.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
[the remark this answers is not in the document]
LukasKorba: Or switch to Int32 on both platforms.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
LukasKorba: Resolved with custom

```swift
    private static func subdata(of data: Data, in range: Range<Data.Index>) throws -> Data {
        guard data.count >= range.lowerBound.distance(to: range.upperBound) else {
            throw AddressBookClient.AddressBookClientError.subdataRange
        }
        
        return data.subdata(in: range)
    }
```

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
LukasKorba:
```swift
return try contactsFrom(unencryptedData: data)
```

used inside instead

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Swapped the order of functions, attempt to delete is now performed only if parse passed.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Throwing an error but leaving the `latestKnownContacts = contacts` so a user can continue and use in-memory contacts.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Set for the follow up task #1408

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: #1408 follow up

modules/Sources/Features/AddressBook/AddressBookStore.swift
LukasKorba: Follow up #1408

zashi-internal.entitlements
LukasKorba: Removed from this PR

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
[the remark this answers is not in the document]
LukasKorba: Updated to MemoryLayout<Int64>.size + moved the value to the constant Constants.int64Size so it's resolved only once and reused.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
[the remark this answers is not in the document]
LukasKorba: Serizalization code taken out into a separate function.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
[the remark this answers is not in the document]
LukasKorba: Re-worked to be able to report `remote store failed` even though local store succeeded.

LukasKorba (approved):
LGTM

Issue #1408, opened 2024-11-14
Address Book encryption follow up tweaks


LukasKorba:
- Merge strategy (This function mixes the merge with the rest of the store logic. I would factor the merge out, since that's a pure function. Maybe also add a test for the merge.)
-

daira:
Copying some comment threads.

[Concerning the call to `storeContacts` from `allLocalContacts:`](https://github.com/Electric-Coin-Company/zashi-ios/pull/1409/files#r1843928154) :

@daira:
> I read your comment https://github.com/Electric-Coin-Company/zashi-ios/pull/1409#discussion_r1843528714 and I agree it's necessary to call `syncContacts` (with `storeAfterSync: true`), but given that change, I don't understand why we need to also call `storeContacts`. `syncContacts` should correctly handle the case where the remote file doesn't already exist.

@LukasKorba replied:
> I agree it's redundant in some scenarios but `syncContacts` is done in a way that it could throw an error and terminate the store completely. The current code ensure there is at least local store done and sync "can" fail. We might want to improve the order and robustness of the solution but I put it to follow up work in #1408

@daira:
> If we want `syncContacts` to always store locally, there's an easy way to do that: set `cannotUpdateRemote = true` in those error cases (i.e. treat them the same way as `RemoteStorageError.containerURL`), rather than propagating the error directly. If we want to record the reason then that can be done as well. Anyway this has been merged as-is now, but please consider that approach for #1408.

and [in `deleteContact:`](https://github.com/Electric-Coin-Company/zashi-ios/pull/1409/files#r1844052179) :

@daira:
> Non-blocking but please document this on #1408:
>
> If the remote store fails here but a previous store containing the deleted contact succeeded, then the remote file will still contain that contact, and so on the next sync, we'll restore it contrary to the user's intent.
>
> I think the best way to fix this is probably for the deletion to fail if we can't store remotely. That would mean that we should try to delete the contact from the remote store *first*, so that there's no partial failure case where we delete it locally but then report failure.

daira:
Also we should probably record the reason for failure (i.e. the exception object) in the `.failure` case of `RemoteStoreResult`.

PR #1409, opened 2024-11-14
Resolution of review comments

LukasKorba:
- All blocking comments from the PR have been resolved

# Author

# Reviewer

### DIFF
--- a/modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
+++ b/modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
@@ -218,8 +218,23 @@ extension AddressBookClient {
             return nil
         }
         
-        return bytes.withUnsafeBytes {
-            $0.load(as: Int.self).bigEndian
+        return bytes.withUnsafeBytes { ptr -> Int in
+            // Check if the pointer is properly aligned for Int
+            if ptr.baseAddress?.alignedUp(toMultipleOf: MemoryLayout<Int>.alignment) == ptr.baseAddress {
+                // Memory is already aligned
+                return ptr.load(as: Int.self).bigEndian
+            } else {
+                // Handle unaligned memory
+                var value: Int = 0
+                
+                withUnsafeMutableBytes(of: &value) { valuePtr in
+                    // Copy bytes manually to handle unaligned access
+                    for i in 0..<Swift.min(ptr.count, MemoryLayout<Int>.size) {
+                        valuePtr[i] = ptr[i]
+                    }
+                }
+                return value.bigEndian
+            }
         }
     }
     
--- a/modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
+++ b/modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
@@ -152,7 +152,7 @@ extension AddressBookClient {
         
         let abContacts = AddressBookContacts(
             lastUpdated: lastUpdated,
-            version: AddressBookContacts.Constants.version,
+            version: version,
             contacts: IdentifiedArrayOf(uniqueElements: contacts)
         )
         
--- a/modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
+++ b/modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
@@ -108,7 +108,7 @@ extension AddressBookClient {
             let sealed = try ChaChaPoly.SealedBox.init(combined: encryptedSubData.suffix(from: 32))
             let data = try ChaChaPoly.open(sealed, using: subKey)
             
-            return try contactsFrom(unencryptedData: data)
+            return try contactsFrom(data)
         } else {
             throw AddressBookClientError.encryptionVersionNotSupported
         }
@@ -118,11 +118,11 @@ extension AddressBookClient {
     ///     [Unencrypted data]        `address book version`
     ///     [Unencrypted data]        `timestamp`
     ///     [Unencrypted data]        `contacts`
-    static func contactsFrom(unencryptedData: Data) throws -> AddressBookContacts {
+    static func contactsFrom(_ data: Data) throws -> AddressBookContacts {
         var offset = 0
         
         // Deserialize `version`
-        let versionBytes = try AddressBookClient.subdata(of: unencryptedData, in: offset..<(offset + Constants.int64Size))
+        let versionBytes = try AddressBookClient.subdata(of: data, in: offset..<(offset + Constants.int64Size))
         offset += Constants.int64Size
         
         // Deserialize and check `address book version`
@@ -131,12 +131,12 @@ extension AddressBookClient {
         }
         
         // Deserialize `lastUpdated`
-        guard let lastUpdated = try AddressBookClient.deserializeDate(from: unencryptedData, at: &offset) else {
+        guard let lastUpdated = try AddressBookClient.deserializeDate(from: data, at: &offset) else {
             return .empty
         }
         
         // Deserialize `contactsCount`
-        let contactsCountBytes = try AddressBookClient.subdata(of: unencryptedData, in: offset..<(offset + Constants.int64Size))
+        let contactsCountBytes = try AddressBookClient.subdata(of: data, in: offset..<(offset + Constants.int64Size))
         offset += Constants.int64Size
         
         guard let contactsCount = AddressBookClient.bytesToInt(Array(contactsCountBytes)) else {
@@ -145,7 +145,7 @@ extension AddressBookClient {
         
         var contacts: [Contact] = []
         for _ in 0..<contactsCount {
-            if let contact = try AddressBookClient.deserializeContact(from: unencryptedData, at: &offset) {
+            if let contact = try AddressBookClient.deserializeContact(from: data, at: &offset) {
                 contacts.append(contact)
             }
         }
@@ -257,7 +257,7 @@ extension AddressBookClient {
     }
     
     private static func subdata(of data: Data, in range: Range<Data.Index>) throws -> Data {
-        guard data.count >= range.lowerBound.distance(to: range.upperBound) else {
+        guard data.count >= range.upperBound else {
             throw AddressBookClient.AddressBookClientError.subdataRange
         }
         
--- a/modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
+++ b/modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
@@ -108,7 +108,7 @@ extension AddressBookClient {
             let sealed = try ChaChaPoly.SealedBox.init(combined: encryptedSubData.suffix(from: 32))
             let data = try ChaChaPoly.open(sealed, using: subKey)
             
-            return try contactsFrom(data)
+            return try contactsFrom(plainData: data)
         } else {
             throw AddressBookClientError.encryptionVersionNotSupported
         }
@@ -118,11 +118,11 @@ extension AddressBookClient {
     ///     [Unencrypted data]        `address book version`
     ///     [Unencrypted data]        `timestamp`
     ///     [Unencrypted data]        `contacts`
-    static func contactsFrom(_ data: Data) throws -> AddressBookContacts {
+    static func contactsFrom(plainData: Data) throws -> AddressBookContacts {
         var offset = 0
         
         // Deserialize `version`
-        let versionBytes = try AddressBookClient.subdata(of: data, in: offset..<(offset + Constants.int64Size))
+        let versionBytes = try AddressBookClient.subdata(of: plainData, in: offset..<(offset + Constants.int64Size))
         offset += Constants.int64Size
         
         // Deserialize and check `address book version`
@@ -131,12 +131,12 @@ extension AddressBookClient {
         }
         
         // Deserialize `lastUpdated`
-        guard let lastUpdated = try AddressBookClient.deserializeDate(from: data, at: &offset) else {
+        guard let lastUpdated = try AddressBookClient.deserializeDate(from: plainData, at: &offset) else {
             return .empty
         }
         
         // Deserialize `contactsCount`
-        let contactsCountBytes = try AddressBookClient.subdata(of: data, in: offset..<(offset + Constants.int64Size))
+        let contactsCountBytes = try AddressBookClient.subdata(of: plainData, in: offset..<(offset + Constants.int64Size))
         offset += Constants.int64Size
         
         guard let contactsCount = AddressBookClient.bytesToInt(Array(contactsCountBytes)) else {
@@ -145,7 +145,7 @@ extension AddressBookClient {
         
         var contacts: [Contact] = []
         for _ in 0..<contactsCount {
-            if let contact = try AddressBookClient.deserializeContact(from: data, at: &offset) {
+            if let contact = try AddressBookClient.deserializeContact(from: plainData, at: &offset) {
                 contacts.append(contact)
             }
         }
@@ -218,23 +218,8 @@ extension AddressBookClient {
             return nil
         }
         
-        return bytes.withUnsafeBytes { ptr -> Int in
-            // Check if the pointer is properly aligned for Int
-            if ptr.baseAddress?.alignedUp(toMultipleOf: MemoryLayout<Int>.alignment) == ptr.baseAddress {
-                // Memory is already aligned
-                return ptr.load(as: Int.self).bigEndian
-            } else {
-                // Handle unaligned memory
-                var value: Int = 0
-                
-                withUnsafeMutableBytes(of: &value) { valuePtr in
-                    // Copy bytes manually to handle unaligned access
-                    for i in 0..<Swift.min(ptr.count, MemoryLayout<Int>.size) {
-                        valuePtr[i] = ptr[i]
-                    }
-                }
-                return value.bigEndian
-            }
+        return bytes.withUnsafeBytes { ptr -> Int? in
+            Int.init(exactly: ptr.loadUnaligned(as: Int64.self).bigEndian)
         }
     }
     
--- a/modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
+++ b/modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
@@ -13,24 +13,7 @@ import CryptoKit
 import WalletStorage
 
 extension AddressBookClient {
-    /// Encrypts address book contacts. The structure:
-    ///     [Unencrypted data]    `encryption version`
-    ///     [Unencrypted data]    `salt`
-    ///     [Encrypted data]        `address book version`
-    ///     [Encrypted data]        `timestamp`
-    ///     [Encrypted data]        `contacts`
-    ///
-    /// This method always produces the latest structure with he latest encryption version.
-    static func encryptContacts(_ abContacts: AddressBookContacts) throws -> Data {
-        @Dependency(\.walletStorage) var walletStorage
-        
-        guard let encryptionKeys = try? walletStorage.exportAddressBookEncryptionKeys(), let addressBookKey = encryptionKeys.getCached(account: 0) else {
-            throw AddressBookClient.AddressBookClientError.missingEncryptionKey
-        }
-        
-        var encryptionVersionData = Data()
-        encryptionVersionData.append(contentsOf: intToBytes(AddressBookEncryptionKeys.Constants.version))
-        
+    static func serializeContacts(_ abContacts: AddressBookContacts) -> Data {
         var dataForEncryption = Data()
         
         // Serialize `address book version`
@@ -47,6 +30,29 @@ extension AddressBookClient {
             let serializedContact = serializeContact(contact)
             dataForEncryption.append(serializedContact)
         }
+        
+        return dataForEncryption
+    }
+    
+    /// Encrypts address book contacts. The structure:
+    ///     [Unencrypted data]    `encryption version`
+    ///     [Unencrypted data]    `salt`
+    ///     [Encrypted data]        `address book version`
+    ///     [Encrypted data]        `timestamp`
+    ///     [Encrypted data]        `contacts`
+    ///
+    /// This method always produces the latest structure with the latest encryption version.
+    static func encryptContacts(_ abContacts: AddressBookContacts) throws -> Data {
+        @Dependency(\.walletStorage) var walletStorage
+        
+        guard let encryptionKeys = try? walletStorage.exportAddressBookEncryptionKeys(), let addressBookKey = encryptionKeys.getCached(account: 0) else {
+            throw AddressBookClient.AddressBookClientError.missingEncryptionKey
+        }
+        
+        var encryptionVersionData = Data()
+        encryptionVersionData.append(contentsOf: intToBytes(AddressBookEncryptionKeys.Constants.version))
+        
+        let dataForEncryption = AddressBookClient.serializeContacts(abContacts)
         
         // Generate a fresh one-time sub-key for encrypting the address book.
         let salt = SymmetricKey(size: SymmetricKeySize.bits256)
@@ -84,15 +90,15 @@ extension AddressBookClient {
         var offset = 0
         
         // Deserialize `encryption version`
-        let encryptionVersionBytes = encryptedData.subdata(in: offset..<(offset + MemoryLayout<Int>.size))
-        offset += MemoryLayout<Int>.size
+        let encryptionVersionBytes = try AddressBookClient.subdata(of: encryptedData, in: offset..<(offset + Constants.int64Size))
+        offset += Constants.int64Size
         
         guard let encryptionVersion = AddressBookClient.bytesToInt(Array(encryptionVersionBytes)) else {
             return .empty
         }
         
         if encryptionVersion == AddressBookEncryptionKeys.Constants.version {
-            let encryptedSubData = encryptedData.subdata(in: offset..<encryptedData.count)
+            let encryptedSubData = try AddressBookClient.subdata(of: encryptedData, in: offset..<encryptedData.count)
             
             // Derive the sub-key for decrypting the address book.
             let salt = encryptedSubData.prefix(upTo: 32)
@@ -101,42 +107,8 @@ extension AddressBookClient {
             // Unseal the encrypted address book.
             let sealed = try ChaChaPoly.SealedBox.init(combined: encryptedSubData.suffix(from: 32))
             let data = try ChaChaPoly.open(sealed, using: subKey)
-            offset = 0
             
-            // Deserialize `address book version`
-            let abVersionCountBytes = data.subdata(in: offset..<(offset + MemoryLayout<Int>.size))
-            guard let abVersion =  AddressBookClient.bytesToInt(Array(abVersionCountBytes)) else {
-                return .empty
-            }
-            offset += MemoryLayout<Int>.size
-            
-            // Deserialize `lastUpdated`
-            guard let lastUpdated = AddressBookClient.deserializeDate(from: data, at: &offset) else {
-                return .empty
-            }
-            
-            // Deserialize `contactsCount`
-            let contactsCountBytes = data.subdata(in: offset..<(offset + MemoryLayout<Int>.size))
-            offset += MemoryLayout<Int>.size
-            
-            guard let contactsCount = AddressBookClient.bytesToInt(Array(contactsCountBytes)) else {
-                return .empty
-            }
-            
-            var contacts: [Contact] = []
-            for _ in 0..<contactsCount {
-                if let contact = AddressBookClient.deserializeContact(from: data, at: &offset) {
-                    contacts.append(contact)
-                }
-            }
-            
-            let abContacts = AddressBookContacts(
-                lastUpdated: lastUpdated,
-                version: abVersion,
-                contacts: IdentifiedArrayOf(uniqueElements: contacts)
-            )
-            
-            return abContacts
+            return try contactsFrom(unencryptedData: data)
         } else {
             throw AddressBookClientError.encryptionVersionNotSupported
         }
@@ -150,8 +122,8 @@ extension AddressBookClient {
         var offset = 0
         
         // Deserialize `version`
-        let versionBytes = unencryptedData.subdata(in: offset..<(offset + MemoryLayout<Int>.size))
-        offset += MemoryLayout<Int>.size
+        let versionBytes = try AddressBookClient.subdata(of: unencryptedData, in: offset..<(offset + Constants.int64Size))
+        offset += Constants.int64Size
         
         // Deserialize and check `address book version`
         guard let version = AddressBookClient.bytesToInt(Array(versionBytes)), version == AddressBookContacts.Constants.version else {
@@ -159,13 +131,13 @@ extension AddressBookClient {
         }
         
         // Deserialize `lastUpdated`
-        guard let lastUpdated = AddressBookClient.deserializeDate(from: unencryptedData, at: &offset) else {
+        guard let lastUpdated = try AddressBookClient.deserializeDate(from: unencryptedData, at: &offset) else {
             return .empty
         }
         
         // Deserialize `contactsCount`
-        let contactsCountBytes = unencryptedData.subdata(in: offset..<(offset + MemoryLayout<Int>.size))
-        offset += MemoryLayout<Int>.size
+        let contactsCountBytes = try AddressBookClient.subdata(of: unencryptedData, in: offset..<(offset + Constants.int64Size))
+        offset += Constants.int64Size
         
         guard let contactsCount = AddressBookClient.bytesToInt(Array(contactsCountBytes)) else {
             return .empty
@@ -173,7 +145,7 @@ extension AddressBookClient {
         
         var contacts: [Contact] = []
         for _ in 0..<contactsCount {
-            if let contact = AddressBookClient.deserializeContact(from: unencryptedData, at: &offset) {
+            if let contact = try AddressBookClient.deserializeContact(from: unencryptedData, at: &offset) {
                 contacts.append(contact)
             }
         }
@@ -210,19 +182,19 @@ extension AddressBookClient {
         return data
     }
     
-    private static func deserializeContact(from data: Data, at offset: inout Int) -> Contact? {
+    private static func deserializeContact(from data: Data, at offset: inout Int) throws -> Contact? {
         // Deserialize `lastUpdated`
-        guard let lastUpdated = AddressBookClient.deserializeDate(from: data, at: &offset) else {
+        guard let lastUpdated = try AddressBookClient.deserializeDate(from: data, at: &offset) else {
             return nil
         }
         
         // Deserialize `address`
-        guard let address = readString(from: data, at: &offset) else {
+        guard let address = try readString(from: data, at: &offset) else {
             return nil
         }
         
         // Deserialize `name`
-        guard let name = readString(from: data, at: &offset) else {
+        guard let name = try readString(from: data, at: &offset) else {
             return nil
         }
         
@@ -242,7 +214,7 @@ extension AddressBookClient {
     }
     
     private static func bytesToInt(_ bytes: [UInt8]) -> Int? {
-        guard bytes.count == MemoryLayout<Int>.size else {
+        guard bytes.count == Constants.int64Size else {
             return nil
         }
         
@@ -259,10 +231,10 @@ extension AddressBookClient {
         return AddressBookClient.intToBytes(timestamp)
     }
     
-    private static func deserializeDate(from data: Data, at offset: inout Int) -> Date? {
+    private static func deserializeDate(from data: Data, at offset: inout Int) throws -> Date? {
         // Extract the bytes for the timestamp (assume it's stored as an Int)
-        let timestampBytes = data.subdata(in: offset..<(offset + MemoryLayout<Int>.size))
-        offset += MemoryLayout<Int>.size
+        let timestampBytes = try AddressBookClient.subdata(of: data, in: offset..<(offset + Constants.int64Size))
+        offset += Constants.int64Size
         
         // Convert the bytes back to an Int
         guard let timestamp = AddressBookClient.bytesToInt(Array(timestampBytes)) else { return nil }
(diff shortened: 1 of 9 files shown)


modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -150,30 +122,30 @@ extension AddressBookClient {
         var offset = 0
         
         // Deserialize `version`
-        let versionBytes = unencryptedData.subdata(in: offset..<(offset + MemoryLayout<Int>.size))
-        offset += MemoryLayout<Int>.size
+        let versionBytes = try AddressBookClient.subdata(of: unencryptedData, in: offset..<(offset + Constants.int64Size))
```
daira, line 125: Non-blocking: 

The name `unencryptedData` is less apt now that this can be either data that has never been encrypted, or data that has been decrypted. Maybe change it to just `data`.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -272,15 +244,25 @@ extension AddressBookClient {
...
     
     // Helper function to read a string from the data using a length prefix
-    private static func readString(from data: Data, at offset: inout Int) -> String? {
+    private static func readString(from data: Data, at offset: inout Int) throws -> String? {
         // Read the length first (assumes the length is stored as an Int)
-        let lengthBytes = data.subdata(in: offset..<(offset + MemoryLayout<Int>.size))
-        offset += MemoryLayout<Int>.size
+        let lengthBytes = try AddressBookClient.subdata(of: data, in: offset..<(offset + Constants.int64Size))
+        offset += Constants.int64Size
         guard let length = AddressBookClient.bytesToInt(Array(lengthBytes)), length > 0 else { return nil }
```
daira, line 251: Should a string of length 0 be accepted?

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -272,15 +244,25 @@ extension AddressBookClient {
...
         
         // Read the string bytes
-        let stringBytes = data.subdata(in: offset..<(offset + length))
+        let stringBytes = try AddressBookClient.subdata(of: data, in: offset..<(offset + length))
         offset += length
         return AddressBookClient.bytesToString(Array(stringBytes))
     }
+    
+    private static func subdata(of data: Data, in range: Range<Data.Index>) throws -> Data {
+        guard data.count >= range.lowerBound.distance(to: range.upperBound) else {
```
daira, line 260: Blocking. 

```suggestion
        guard data.count >= range.upperBound else {
```
(`upperBound` is exclusive and relative to the start of `data` in this case. It's not necessary for the check to involve `lowerBound`, because `range.upperBound >= range.lowerBound` for any `range`.)

LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -69,7 +77,7 @@ extension AddressBookClient: DependencyKey {
                         let unencryptedFileURL = documentsDirectory.appendingPathComponent(Constants.unencryptedFilename)
                         
                         if let contactsData = try? Data(contentsOf: unencryptedFileURL) {
-                            // file exists, ensure data are parsed, re-saved and file deteled
+                            // file exists; ensure data are parsed, re-saved and file deteled
```
daira, line 80:
```suggestion
                            // Unencrypted file exists; ensure data are parsed, re-saved as encrypted, and
                            // the unencrypted file deleted.
```

nuttycom (approved):
utACK 9441a06b09d8fc9d896e008c68ca1bf0b545552f

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -146,34 +118,34 @@ extension AddressBookClient {
...
             return .empty
         }
         
         var contacts: [Contact] = []
         for _ in 0..<contactsCount {
-            if let contact = AddressBookClient.deserializeContact(from: unencryptedData, at: &offset) {
+            if let contact = try AddressBookClient.deserializeContact(from: data, at: &offset) {
                 contacts.append(contact)
             }
         }
```
daira, line 151: On line 155, pass `version` instead of `AddressBookContacts.Constants.version`. They are currently the same, but passing `version` will just work once this changes.

LukasKorba: Updated

nuttycom (approved):
utACK

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -242,7 +214,7 @@ extension AddressBookClient {
     }
     
     private static func bytesToInt(_ bytes: [UInt8]) -> Int? {
-        guard bytes.count == MemoryLayout<Int>.size else {
+        guard bytes.count == Constants.int64Size else {
             return nil
         }
```
daira, line 220: I believe the `.load` call on line 122 has Undefined Behaviour when the bytes being accessed are not aligned, which they may not be. For discussion of this at the Swift layer see https://stackoverflow.com/questions/63282018/how-can-you-create-an-aligned-array-of-bytes-and-read-from-it . The actual behaviour is dependent on the platform (both  architecture and operating system). In particular both x86-64 and [ARM](https://developer.arm.com/documentation/102376/0200/Alignment-and-endianness/Alignment) support unaligned accesses to normal memory, which is why you wouldn't have seen the problem in testing. An example of a still-relevant architecture that does not natively support unaligned addresses (the OS [might handle it via a trap](https://www.kernel.org/doc/Documentation/unaligned-memory-access.txt)) is RISC-V. In any case it's Undefined Behaviour at the Swift level on *all* platforms.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -54,13 +61,14 @@ extension AddressBookClient: DependencyKey {
                     let encryptedFileURL = documentsDirectory.appendingPathComponent(try AddressBookClient.filenameForEncryptedFile())
 
                     if let contactsData = try? Data(contentsOf: encryptedFileURL) {
-                        // file exists, try to find the unencrypted file in case the delete failed
+                        let contacts = try AddressBookClient.contactsFrom(encryptedData: contactsData)
+                        
+                        // file exists and was successfuly decrypted;
```
daira, line 66:
```suggestion
                        // file exists and was successfuly decrypted and parsed;
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -69,16 +77,16 @@ extension AddressBookClient: DependencyKey {
                         let unencryptedFileURL = documentsDirectory.appendingPathComponent(Constants.unencryptedFilename)
                         
                         if let contactsData = try? Data(contentsOf: unencryptedFileURL) {
-                            // file exists, ensure data are parsed, re-saved and file deteled
-                            let contacts = try AddressBookClient.contactsFrom(unencryptedData: contactsData)
+                            // Unencrypted file exists; ensure data are parsed, re-saved as encrypted, and file deteled
```
daira, line 80:
```suggestion
                            // Unencrypted file exists; ensure data are parsed, re-saved as encrypted, and file deleted
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -69,16 +77,16 @@ extension AddressBookClient: DependencyKey {
...
+                            let contacts = try AddressBookClient.contactsFrom(contactsData)
 
                             // try to encrypt and store the data
                             do {
                                 try AddressBookClient.storeContacts(contacts, remoteStorage: remoteStorage, remoteStore: false)
```
daira, line 85: Why is this not also trying to store remotely? I don't understand why the `remoteStore` flag needs to exist. It's easier to reason about if we always try to do a remote store whenever we do a local store.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -54,13 +61,14 @@ extension AddressBookClient: DependencyKey {
                     let encryptedFileURL = documentsDirectory.appendingPathComponent(try AddressBookClient.filenameForEncryptedFile())
```
daira, line 61: Nit: on line 54, "locale storage" should be "local storage".

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -206,11 +214,11 @@ extension AddressBookClient: DependencyKey {
         return syncedContacts
```
daira, line 214: On line 211, it's only necessary to store the contacts locally if the merge added anything relative to the original contents of the local file. More importantly (because remote storage has higher latency), it's only necessary to store the contacts remotely if the merge added anything relative to the original contents of the remote file.

It might sound as though this optimization makes things more complicated. But actually it makes them simpler, because then you no longer need the `remoteStore` flag. In most cases, the remote storage will be synced with the local storage already, and we won't do a remote store. If it happens not to be synced, then this approach (where we remove the `remoteStore` flag) means that in the case of a mutation, we might do two remote stores — one to merge in the local contacts and one to make the update. But that case should be so rare that it doesn't matter for efficiency.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -69,16 +77,16 @@ extension AddressBookClient: DependencyKey {
...
-                            // file exists, ensure data are parsed, re-saved and file deteled
-                            let contacts = try AddressBookClient.contactsFrom(unencryptedData: contactsData)
+                            // Unencrypted file exists; ensure data are parsed, re-saved as encrypted, and file deteled
+                            let contacts = try AddressBookClient.contactsFrom(contactsData)
 
                             // try to encrypt and store the data
                             do {
                                 try AddressBookClient.storeContacts(contacts, remoteStorage: remoteStorage, remoteStore: false)
                             } catch {
                                 // the store of the new file failed, skip the file remove
```
daira, line 87:
```suggestion
                                // the store of the new file failed locally, skip the file remove
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -206,11 +214,11 @@ extension AddressBookClient: DependencyKey {
         return syncedContacts
     }
     
-    private static func storeContacts(
+    @discardableResult private static func storeContacts(
```
daira, line 217: If you take my suggestion to remove the `remoteStore` flag, then in that case the result should not be discardable. The reason for it to be discardable currently is that we know it will be `StoreResult.success` whenever `remoteStore` is false. But that reason will no longer apply.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -206,11 +214,11 @@ extension AddressBookClient: DependencyKey {
...
         _ abContacts: AddressBookContacts,
         remoteStorage: RemoteStorageClient,
         remoteStore: Bool = true
```
daira, line 220: It's clearer for this not to have a default.
```suggestion
        remoteStore: Bool
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -114,12 +122,12 @@ extension AddressBookClient: DependencyKey {
                 
                 syncedContacts.contacts.append($0)
                 
-                try storeContacts(syncedContacts, remoteStorage: remoteStorage)
+                let storeResult = try storeContacts(syncedContacts, remoteStorage: remoteStorage)
```
daira, line 125:
```suggestion
                let storeResult = try storeContacts(syncedContacts, remoteStorage: remoteStorage, remoteStore: true)
```
(See my suggestion below for this not to have a default.)

modules/Sources/Features/AddressBook/AddressBookStore.swift
```diff
@@ -210,7 +210,7 @@ public struct AddressBook {
                             .send(.updateDestination(nil))
                         )
                     } catch {
-                        // TODO: FIXME
+                        // TODO: [#1408] erro handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```
daira, line 213:
```suggestion
                        // TODO: [#1408] error handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```

modules/Sources/Features/AddressBook/AddressBookStore.swift
```diff
@@ -238,14 +238,17 @@ public struct AddressBook {
 
             case .saveButtonTapped:
                 do {
-                    let abContacts = try addressBook.storeContact(Contact(address: state.address, name: state.name))
+                    let result = try addressBook.storeContact(Contact(address: state.address, name: state.name))
+                    let abContacts = result.contacts
+                    if result.storeResult == .remoteFailed {
+                        // TODO: [#1408] erro handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```
daira, line 244:
```suggestion
                        // TODO: [#1408] error handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -114,12 +122,12 @@ extension AddressBookClient: DependencyKey {
...
                 
                 // update the latest known contacts
                 latestKnownContacts = syncedContacts
                 
-                return syncedContacts
+                return (syncedContacts, storeResult)
             },
             deleteContact: {
```
daira, line 132: This should also handle the error when we fail to store remotely.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Features/SendFlow/SendFlowStore.swift
```diff
@@ -253,8 +253,7 @@ public struct SendFlow {
                         .send(.fetchedABContacts(abContacts))
                         )
                 } catch {
-                    // TODO: FIXME
-                    print("__LD fetchABContactsRequested Error: \(error.localizedDescription)")
+                    // TODO: [#1408] erro handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```
daira, line 256:
```suggestion
                    // TODO: [#1408] error handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```

modules/Sources/Features/SendConfirmation/SendConfirmationStore.swift
```diff
@@ -117,8 +117,7 @@ public struct SendConfirmation {
                     let abContacts = try addressBook.allLocalContacts()
                     return .send(.fetchedABContacts(abContacts))
                 } catch {
-                    print("__LD fetchABContactsRequested Error: \(error.localizedDescription)")
-                    // TODO: FIXME
+                    // TODO: [#1408] erro handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```
daira, line 120:
```suggestion
                    // TODO: [#1408] error handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```

modules/Sources/Features/AddressBook/AddressBookStore.swift
```diff
@@ -274,8 +276,7 @@ public struct AddressBook {
                             let syncedContacts = try await addressBook.syncContacts(abContacts)
                             await send(.fetchedABContacts(syncedContacts, false))
                         } catch {
-                            print("__LD syncContacts Error: \(error.localizedDescription)")
-                            // TODO: FIXME
+                            // TODO: [#1408] erro handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```
daira, line 279:
```suggestion
                            // TODO: [#1408] error handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```

modules/Sources/Features/AddressBook/AddressBookStore.swift
```diff
@@ -261,8 +264,7 @@ public struct AddressBook {
                     let abContacts = try addressBook.allLocalContacts()
                     return .send(.fetchedABContacts(abContacts, true))
                 } catch {
-                    print("__LD fetchABContactsRequested Error: \(error.localizedDescription)")
-                    // TODO: FIXME
+                    // TODO: [#1408] erro handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```
daira, line 267:
```suggestion
                    // TODO: [#1408] error handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```

modules/Sources/Features/AddressBook/AddressBookStore.swift
```diff
@@ -238,14 +238,17 @@ public struct AddressBook {
...
+                        // TODO: [#1408] erro handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
+                    }
                     return .concatenate(
                         .send(.fetchedABContacts(abContacts, false)),
                         .send(.contactStoreSuccess)
                     )
                 } catch {
-                    // TODO: FIXME
-                    print("__LD saveButtonTapped Error: \(error.localizedDescription)")
+                    // TODO: [#1408] erro handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```
daira, line 251:
```suggestion
                    // TODO: [#1408] error handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```

modules/Sources/Features/TransactionList/TransactionListStore.swift
```diff
@@ -71,8 +71,7 @@ public struct TransactionList {
                 let abContacts = try addressBook.allLocalContacts()
                 state.addressBookContacts = abContacts
             } catch {
-                print("__LD fetchABContactsRequested Error: \(error.localizedDescription)")
-                // TODO: FIXME
+                // TODO: [#1408] erro handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```
daira, line 74:
```suggestion
                // TODO: [#1408] error handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
daira, line 217:
```suggestion
    private static func storeContacts(
```

daira (changes requested):
Reviewed, changes needed.

modules/Sources/Features/AddressBook/AddressBookStore.swift
LukasKorba: Updated

modules/Sources/Features/AddressBook/AddressBookStore.swift
LukasKorba: Updated

modules/Sources/Features/SendFlow/SendFlowStore.swift
LukasKorba: Updated

modules/Sources/Features/AddressBook/AddressBookStore.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Actually, we need to sync contacts and re-store instead of store, that would replace remote file with local one. We need merge strategy to be applied. Code updated.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
daira: Thanks for catching my error.

modules/Sources/Dependencies/AddressBookClient/AddressBookInterface.swift
```diff
@@ -17,8 +17,8 @@ extension DependencyValues {
 
 @DependencyClient
 public struct AddressBookClient {
-    public let allLocalContacts: () throws -> AddressBookContacts
-    public let syncContacts: (AddressBookContacts?) async throws -> AddressBookContacts
-    public let storeContact: (Contact) throws -> AddressBookContacts
-    public let deleteContact: (Contact) throws -> AddressBookContacts
+    public let allLocalContacts: () throws -> (contacts: AddressBookContacts, remoteStoreResult: RemoteStoreResult?)
```
daira, line 20: I think this can return `RemoteStoreResult.notAttempted` rather than `nil`, meaning that it does not need to be optional.
```suggestion
    public let allLocalContacts: () throws -> (contacts: AddressBookContacts, remoteStoreResult: RemoteStoreResult)
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -41,10 +49,10 @@ extension AddressBookClient: DependencyKey {
             allLocalContacts: {
                 // return latest known contacts or load ones for the first time
                 guard latestKnownContacts == nil else {
-                    return latestKnownContacts ?? .empty
+                    return (latestKnownContacts ?? .empty, nil)
```
daira, line 52:
```suggestion
                    return (latestKnownContacts ?? .empty, RemoteStoreResult.notAttempted)
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -54,39 +62,46 @@ extension AddressBookClient: DependencyKey {
...
                         let unencryptedFileURL = documentsDirectory.appendingPathComponent(Constants.unencryptedFilename)
                         if FileManager.default.fileExists(atPath: unencryptedFileURL.path) {
                             try? FileManager.default.removeItem(at: unencryptedFileURL)
                         }
-                        
-                        let contacts = try AddressBookClient.contactsFrom(encryptedData: contactsData)
 
                         latestKnownContacts = contacts
-                        return contacts
+                        return (contacts, nil)
```
daira, line 75:
```suggestion
                        return (contacts, RemoteStoreResult.notAttempted)
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -54,39 +62,46 @@ extension AddressBookClient: DependencyKey {
...
                     } else {
                         // Fallback to the unencrypted file check and resolution
                         let unencryptedFileURL = documentsDirectory.appendingPathComponent(Constants.unencryptedFilename)
                         
                         if let contactsData = try? Data(contentsOf: unencryptedFileURL) {
-                            // file exists, ensure data are parsed, re-saved and file deteled
-                            let contacts = try AddressBookClient.contactsFrom(unencryptedData: contactsData)
+                            // Unencrypted file exists; ensure data are parsed, re-saved as encrypted, and the original file deleted.
+
+                            var contacts = try AddressBookClient.contactsFrom(contactsData)
```
daira, line 83: When I suggested changing the `unencryptedData:` label in `AddressBookClient.contactsFrom(unencryptedData:)`, I had not seen that there was also a `contactsFrom(encryptedData:)` method. The two methods are now more confusable. Please either revert that change or rename `contactsFrom(_)` to something like `AddressBookClient.decodeContacts`. Sorry for going back and forth on this.

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -146,41 +118,41 @@ extension AddressBookClient {
     ///     [Unencrypted data]        `address book version`
     ///     [Unencrypted data]        `timestamp`
     ///     [Unencrypted data]        `contacts`
-    static func contactsFrom(unencryptedData: Data) throws -> AddressBookContacts {
+    static func contactsFrom(_ data: Data) throws -> AddressBookContacts {
```
daira, line 121: When I suggested changing this label, it hadn't registered with me that it would increase the confusability with `contactsFrom(encryptedData:)`.

If the label were `plainData:`, that would clearly distinguish them while being accurate for both the case where the data has been decrypted, and the case where it was never encrypted.
```suggestion
    static func contactsFrom(plainData: Data) throws -> AddressBookContacts {
```

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -101,42 +107,8 @@ extension AddressBookClient {
...
-            }
-            
-            let abContacts = AddressBookContacts(
-                lastUpdated: lastUpdated,
-                version: abVersion,
-                contacts: IdentifiedArrayOf(uniqueElements: contacts)
-            )
-            
-            return abContacts
+            return try contactsFrom(data)
```
daira, line 111:
```suggestion
            return try contactsFrom(plainData: data)
```

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
```diff
@@ -242,12 +214,27 @@ extension AddressBookClient {
...
+                
+                withUnsafeMutableBytes(of: &value) { valuePtr in
+                    // Copy bytes manually to handle unaligned access
+                    for i in 0..<Swift.min(ptr.count, MemoryLayout<Int>.size) {
+                        valuePtr[i] = ptr[i]
+                    }
+                }
+                return value.bigEndian
+            }
         }
```
daira, line 238: Simply use [loadUnaligned(fromByteOffset:as:)](https://developer.apple.com/documentation/swift/unsaferawbufferpointer/loadunaligned(frombyteoffset:as:)-2r9sy):
```suggestion
        return bytes.withUnsafeBytes {
            Int.init?(exactly: $0.loadUnaligned(fromByteOffset: 0, as: Int64.self).bigEndian)
        }
```
(As a bonus, this also returns `nil` for an out-of-range value on platforms where `Int` is less than 64 bits. We probably don't support such platforms for other reasons, but since it's easy to get right here we might as well.)

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
daira: Yes, I was asking whether it's intended to reject them.

modules/Sources/Dependencies/AddressBookClient/AddressBookInterface.swift
```diff
@@ -17,8 +17,8 @@ extension DependencyValues {
...
+    public let syncContacts: (AddressBookContacts?) async throws -> (contacts: AddressBookContacts, remoteStoreResult: RemoteStoreResult)
+    public let storeContact: (Contact) throws -> (contacts: AddressBookContacts, remoteStoreResult: RemoteStoreResult)
+    public let deleteContact: (Contact) throws -> (contacts: AddressBookContacts, remoteStoreResult: RemoteStoreResult?)
```
daira, line 23: Similarly here.
```suggestion
    public let deleteContact: (Contact) throws -> (contacts: AddressBookContacts, remoteStoreResult: RemoteStoreResult)
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -114,31 +130,32 @@ extension AddressBookClient: DependencyKey {
...
                 
-                var syncedContacts = try syncContacts(contacts: abContacts, remoteStorage: remoteStorage, storeAfterSync: false)
-                
+                let result = try syncContacts(contacts: abContacts, remoteStorage: remoteStorage, storeAfterSync: false)
+                var syncedContacts = result.contacts
+
                 // if it doesn't exist, do nothing
                 guard syncedContacts.contacts.contains($0) else {
-                    return syncedContacts
+                    return (syncedContacts, nil)
```
daira, line 148:
```suggestion
                    return (syncedContacts, RemoteStoreResult.notAttempted)
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -199,34 +216,48 @@ extension AddressBookClient: DependencyKey {
...
         }
-        
-        let filenameForEncryptedFile = try AddressBookClient.filenameForEncryptedFile()
 
         let fileURL = documentsDirectory.appendingPathComponent(filenameForEncryptedFile)
         try encryptedContacts.write(to: fileURL)
-        
+
         // store encrypted data to the remote storage
+        var isRemoteSuccess = remoteStore
```
daira, line 246: I think it's clearer to set this to true only if the call to `storeAddressBookContacts` succeeds.
```suggestion
        var isRemoteSuccess = false
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -199,34 +216,48 @@ extension AddressBookClient: DependencyKey {
...
+        
         if remoteStore {
-            try? remoteStorage.storeAddressBookContacts(encryptedContacts, filenameForEncryptedFile)
+            do {
+                try remoteStorage.storeAddressBookContacts(encryptedContacts, filenameForEncryptedFile)
+            } catch {
+                isRemoteSuccess = false
```
daira, line 252:
```suggestion
                isRemoteSuccess = true
            } catch {
                // we initialized isRemoteSuccess to false
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
daira, line 246: The `(false, true)` case in the switch isn't needed. I think it's clearer to do this (suggestion split over two comments to work around a GitHub limitation).
```suggestion
        var isRemoteSuccess = false
```
```suggestion
        // store encrypted data to the remote storage if requested
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -199,34 +216,48 @@ extension AddressBookClient: DependencyKey {
...
+            }
+        }
+
+        switch (remoteStore, isRemoteSuccess) {
+        case (true, true): return .success
+        case (true, false): return .failure
+        case (false, true): return .notAttempted
+        case (false, false): return .notAttempted
         }
```
daira, line 261:
```suggestion
            do {
                try remoteStorage.storeAddressBookContacts(encryptedContacts, filenameForEncryptedFile)
                return .success
            } catch {
                return .failure
            }
        } else {
            return .notAttempted
        }
```

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -147,7 +164,7 @@ extension AddressBookClient: DependencyKey {
         contacts: AddressBookContacts,
         remoteStorage: RemoteStorageClient,
         storeAfterSync: Bool = true
-    ) throws -> AddressBookContacts {
+    ) throws -> (contacts: AddressBookContacts, remoteStoreResult: RemoteStoreResult) {
         // Ensure remote contacts are prepared
         var remoteContacts: AddressBookContacts = .empty
         var storeData = true
```
daira, line 170: I'm not convinced about the use of this flag, particularly line 177. Why should the remote file not be stored if it doesn't exist and `storeAfterSync` is true?

daira (approved):
utACK with comments (some blocking).

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
nuttycom: Why is that distinction relevant here? This is a static function; it doesn't care about the source of the data or its history.

modules/Sources/Dependencies/AddressBookClient/AddressBookInterface.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -54,39 +62,46 @@ extension AddressBookClient: DependencyKey {
...
                             }
                             
                             try? FileManager.default.removeItem(at: unencryptedFileURL)
                             
                             latestKnownContacts = contacts
-                            return contacts
+                            return (contacts, remoteStoreResult)
                         } else {
-                            return .empty
+                            return (.empty, nil)
```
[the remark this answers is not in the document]
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookInterface.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookEncryption.swift
LukasKorba: Updated

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -114,31 +130,32 @@ extension AddressBookClient: DependencyKey {
...
                 // if it doesn't exist, do nothing
                 guard syncedContacts.contacts.contains($0) else {
-                    return syncedContacts
+                    return (syncedContacts, nil)
                 }
                 
                 syncedContacts.contacts.remove($0)
                 
-                try storeContacts(syncedContacts, remoteStorage: remoteStorage)
+                let remoteStoreResult = try storeContacts(syncedContacts, remoteStorage: remoteStorage)
```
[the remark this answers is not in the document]
LukasKorba: Updated #1408

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -54,39 +62,46 @@ extension AddressBookClient: DependencyKey {
...
+                            // Unencrypted file exists; ensure data are parsed, re-saved as encrypted, and the original file deleted.
+
+                            var contacts = try AddressBookClient.contactsFrom(contactsData)
 
                             // try to encrypt and store the data
+                            var remoteStoreResult: RemoteStoreResult
                             do {
-                                try AddressBookClient.storeContacts(contacts, remoteStorage: remoteStorage, remoteStore: false)
+                                remoteStoreResult = try AddressBookClient.storeContacts(contacts, remoteStorage: remoteStorage, remoteStore: false)
+
```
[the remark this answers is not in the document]
LukasKorba: I agree it's redundant in some scenarios but syncContacts is done in a way that it could throw an error and terminate the store completely. The current code ensure there is at least local store done and sync "can" fail. We might want to improve the order and robustness of the solution but I put it to follow up work in #1408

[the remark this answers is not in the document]
daira: If we want `syncContacts` to always store locally, there's an easy way to do that: set `cannotUpdateRemote = true` in those error cases (i.e. treat them the same way as `RemoteStorageError.containerURL`), rather than propagating the error directly. If we want to record the reason then that can be done as well. Anyway this has been merged as-is now, but please consider that approach for #1408.

modules/Sources/Dependencies/AddressBookClient/AddressBookLiveKey.swift
```diff
@@ -173,60 +194,70 @@ extension AddressBookClient: DependencyKey {
...
                 if $0.id == contact.id {
                     notFound = false
                     
-                    if $0.lastUpdated > contact.lastUpdated {
-                        indexToUpdate = i
+                    // If the timestamps are equal, the local entry takes priority.
+                    if $0.lastUpdated >= contact.lastUpdated {
+                        syncedContacts.contacts[i].name = $0.name
+                        syncedContacts.contacts[i].lastUpdated = $0.lastUpdated
+                        shouldUpdateRemote = true
```
daira, line 208: This is a more serious bug: the `shouldUpdateRemote` logic absolutely depended on merging from local into remote. This way around it will be set to true only when the remote has an entry that should be updated locally, which is completely wrong for the way this variable is used.

daira: Actually, we determined that mutations to the address book cause an unconditional remote store, so this has much smaller impact than I originally thought (although it's still a bug relative to the intended behaviour of `syncContacts` with `storeAfterSync: true`).

PR #1680, opened 2026-04-05
Render ZAP1 attestation memos in transaction details

Zk-nd3r:
ZAP1 memos in shielded transactions currently display as raw hex.  This PR parses them into typed event cards.

**What it does:**
- Detects ZAP1 prefix (`ZAP1:` or `0x5a41503109`) in memo content
- Parses event type byte into human-readable name (21 types)
- Extracts structured fields (wallet hash,  amount,  timestamp,  serial number)
- Renders as a typed event card instead of hex dump

**Files:**
- `Zap1MemoParser.swift` - parser (matches the Kotlin implementation in zodl-android PR #2173)
- `Zap1MemoParserTests.swift` - XCTest coverage
- 1-line patch to `TransactionDetailsStore.swift` at line 434

**Verify any attestation:** frontiercompute.cash/verify.html

Protocol spec: frontiercompute.cash/protocol.html
ZIP 1243: github.com/zcash/zips/pull/1243

### DIFF
--- a/modules/Sources/Features/TransactionDetails/TransactionDetailsStore.swift
+++ b/modules/Sources/Features/TransactionDetails/TransactionDetailsStore.swift
@@ -431,7 +431,7 @@ public struct TransactionDetails {
             case .memosLoaded(let memos):
                 state.areMessagesResolved = true
                 state.$transactionMemos.withLock {
-                    $0[state.transaction.id] = memos.compactMap { $0.toString() }
+                    $0[state.transaction.id] = memos.compactMap { $0.toString() }.map { Zap1Attestation.format($0) ?? $0 }
                 }
                 state.messageStates = state.memos.map {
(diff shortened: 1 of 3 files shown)


Zk-nd3r:
Cross-platform conformance suite added at github.com/Frontier-Compute/zap1/tree/main/contrib/zodl-conformance.  Same 20+ test vectors for both platforms.

Swift parser passes all vectors.  Matches the Kotlin implementation in zodl-android #2173 exactly.  Handles null-byte padded memos (real Zcash 512-byte format).

Zk-nd3r:
Closing this stale integration. My April claim that both wallet PRs were conflict-free is no longer true: the iOS paths have moved, and this patch transforms shared memo state rather than presentation-only output. The ZAP1 wire and type registry also need canonicalization first. I will only resubmit against current paths with raw-memo preservation, the current empty-memo behavior, and current tests.

PR #1690, opened 2026-04-12
Use @preconcurrency on import of packages which aren't migrated to swift concurrency yet

Chlup:
This is first PR from list of PRs that convert Zodl project to Swift 6. In this PR packages which are not converted to Swift 6 are imported with `@preconcurrency`. It shouldn't affect how Zodl works in any way. It is purely technical change and it is prerequisite for futures Swift 6 related changes.


LukasKorba (approved):
LGTM

PR #1691, opened 2026-04-13
Resolve swift 6 concurrency warnings in AppDelegate

Chlup:
Another PR in a series aimed at converting the project to Swift 6. This PR addresses the mutability of `isConnectedToWifi` in `AppDelegate`. I added the `swift-atomics` library to the dependencies (it will be used in more places in later PRs), and `ManagedAtomic` is now used to make `isConnectedToWifi` Sendable. Atomics behave similarly to locks, but they can only be used with primitive types (Bool, Int, Float, etc.). They rely on CPU-level atomic instructions to ensure thread safety of a variable.


LukasKorba (approved):
LGTM

PR #1692, opened 2026-04-13
Make AddressBookClient Swift 6 compliant

Chlup:
Another PR in a series aimed at converting the project to Swift 6. This PR makes `AddressBookClient` (and all related code) Swift 6 compliant. 

Main changes in this PR:
- Logic from this dependency is extracted from closures to it's own class `AddressBookImpl`. This class is private so rest of the codebase still see only `AddressBookClient` interface. This is just refactoring logic is not changed it is just moved.
- `AddressBookClient.latestKnownContacts` variable is now protected by `OSAllocatedUnfairLock`. It makes this variable Sendable even when it's mutated. And thanks to this whole `AddressBookImpl` is Sendable.


LukasKorba (approved):
LGTM

PR #1694, opened 2026-04-13
Make AudioServicesClient swift 6 compliant

Chlup:
Another PR in a series aimed at converting the project to Swift 6. This PR makes `AudioServicesClient` Swift 6 compliant. This is very small dependency so not much is happening here. `AudioServicesTestKey.swift` can be removed because when using `var` closures for dependency client then `@DependencyClient` macro can synthesize "unimplemented" code for us.


LukasKorba (approved):
LGTM

PR #1696, opened 2026-04-13
Make AutolockHandlerClient Swift 6 compliant

Chlup:
Another PR in a series aimed at converting the project to Swift 6. This PR makes `AutolockHandlerClient ` Swift 6 compliant. This is very small dependency so not much is happening here. API of this dependency had to be changed to async because it used `UIDevice` and that must be isolated by @MainActor.


LukasKorba (approved):
LGTM

PR #1713, opened 2026-04-13
Make SupportDataGeneratorClient Swift 6 compliant

Chlup:
Another PR in a series aimed at converting the project to Swift 6. This PR makes `SupportDataGeneratorClient ` Swift 6 compliant. Previously `SupportDataGeneratorClient` used `UIDevice` to get system version. But UIDevice must be used from main thread and that means isolation to @MainActor and that means that API of `SupportDataGeneratorClient` would have to change to async. So instead of `UIDevice` I used `ProcessInfo` to get system version. It gives same results as UIDevice.


LukasKorba (approved):
LGTM

PR #1714, opened 2026-04-13
Make UserMetadataProviderClient Swift 6 compliant

Chlup:
Another PR in a series aimed at converting the project to Swift 6. This PR makes `UserMetadataProviderClient ` Swift 6 compliant. 

`UserMetadataStorage` mutates fews variables inside. One of the solutions would be to use actor and make `UserMetadataStorage`'s API async. But that would require bigger changes outside of this depedency. So instead of this I put all the mutating data to one structure and this structure is now protected by `OSAllocatedUnfairLock`. That makes everything save and Sendable.


LukasKorba (approved):
LGTM

PR #1715, opened 2026-04-13
Make WalletConfigProviderClient Swift 6 compliant

Chlup:
Another PR in a series aimed at converting the project to Swift 6. This PR makes `WalletConfigProviderClient ` Swift 6 compliant. 

I simply removed Combine based API from `WalletConfigProviderClient` and kept only async one. Changes for the app were minimal. It was super hard to rewrite Combine API to be Swift 6 compliant. Not because of use of Combine but because how it was used in this specific case. To use async only was easier.


LukasKorba (approved):
LGTM

PR #1731, opened 2026-04-22
Set voting tx fee and amount to 0

greg0x:
## Summary
- The Confirm & Submit screen showed a hardcoded `0.0001 ZEC` fee and `0.00000001 ZEC` amount. Neither was sourced from the store, and both misrepresent voting txs — there is no user-visible amount or fee on a vote.
- Both rows now render `0 ZEC`.

Linear: ZCA-246

## Test plan
- [ ] Reach the Confirm & Submit screen in a voting flow
- [ ] Verify Amount and Fee both show `0 ZEC`


greg0x:
accidental PR

PR #1750, opened 2026-05-12
[#1718] Redesign server selection: Automatic/Manual connection mode

### LINKED ISSUE #1718: Redesign server selection: Automatic/Manual connection mode
### Motivation

The current server selection UI forces users to pick a single server from a list. Most users don't know which server is best and shouldn't have to care. There's no "just work" option.

A simpler Automatic/Manual toggle gives users a clear mental model: either the app picks the best server for you, or you choose one explicitly.

### Proposed Solution

- **Automatic mode**: App evaluates the best server for wallet sync via `evaluateBestOf` benchmarking at app launch and when the user opens Server settings. The fastest server is selected automatically.
- **Manual mode**: User selects a single server used for sync.
- Migration: users with a custom server default to Manual (preserving privacy); all others default to Automatic.
- New `SelectedServersConfig` / `ConnectionMode` model persisted in UserDefaults (`ups_selectedServers` key).
- `isKnownEndpoint` helper to distinguish built-in servers from custom ones.
- Consistent streaming timeout applied to all known endpoints.

czarcas7ic:
Closes: #1718
Resolves MOB-1039 and MOB-1098.

## Summary

- Redesigned Server Setup around Automatic and Manual connection modes.
- Automatic mode benchmarks known endpoints and keeps using the best active sync endpoint.
- Manual mode pins a single endpoint. Switching from Automatic to Manual preselects the current active endpoint so Save pins it without requiring another tap.
- Migration preserves custom, unknown, and non-default known endpoints in Manual mode. Missing or default endpoints start in Automatic.
- Applies the shared streaming timeout to known endpoints.

## Test plan

- Added reducer and migration coverage for Automatic to Manual preselection, Save without a row tap, custom and unknown active endpoint handling, and no-active-endpoint selection requirements.
- Added migration coverage for default, non-default known, custom, unknown, already migrated, and new-user cases.
- Manual UI testing covered the default Automatic path and custom Manual preservation.


secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerInterface.swift
```diff
@@ -18,6 +18,34 @@ extension DependencyValues {
     }
 }
 
+actor EndpointSwitchCoordinator {
```
Chlup: I don't think that this works as intended. Because of actor reentrancy there is nothing in this code that protects `switchToEndpoint` closure to be called in parallel at the same time. Can you please describe what is the intention behind this?

secant/Sources/Features/ServerSetup/ServerSetupStore.swift
```diff
@@ -23,71 +23,107 @@ extension LightWalletEndpoint: @retroactive Equatable {
 struct ServerSetup {
     let streamingCallTimeoutInMillis = ZcashSDKEnvironment.ZcashSDKConstants.streamingCallTimeoutInMillis
 
+    private enum ServerEvaluationDefaults {
```
Chlup: Almost similar enum is in `RootStore`. But it has little bit different some values there. Is this intended?

secant/Sources/Dependencies/ZcashSDKEnvironment/ZcashSDKEnvironmentLiveKey.swift
```diff
@@ -95,6 +98,92 @@ extension ZcashSDKEnvironment {
...
+        @Dependency(\.userStoredPreferences) var userStoredPreferences
+
+        guard userStoredPreferences.selectedServers() == nil else { return }
+
+        if let existing = userStoredPreferences.server() {
+            let normalizedServer = migrationServerConfig(existing, network: network)
+
+            if shouldPreserveAsManualSelection(normalizedServer, network: network) {
+                do {
+                    try userStoredPreferences.setSelectedServers(.init(mode: .manual, servers: [normalizedServer]))
```
Chlup: Please don't use `.init(...)` but full type name when creating instance. It hurts the readability of code. Thanks.

secant/Sources/Dependencies/ZcashSDKEnvironment/ZcashSDKEnvironmentLiveKey.swift
```diff
@@ -95,6 +98,92 @@ extension ZcashSDKEnvironment {
...
+                } catch {
+                    LoggerProxy.error("[Migration] Failed to persist manual server selection: \(error)")
+                }
+                return
+            }
+        }
+
+        do {
+            try userStoredPreferences.setSelectedServers(.init(mode: .automatic, servers: []))
```
Chlup: Please don't use `.init(...)` but full type name when creating instance. It hurts the readability of code. Thanks.

secant/Sources/Features/Root/RootStore.swift
```diff
@@ -409,16 +435,138 @@ struct Root {
...
+                try await EndpointSwitching.coordinator.switchToEndpoint(
+                    endpoint,
+                    previousEndpoint: currentEndpoint,
+                    switchToEndpoint: sdkSynchronizer.switchToEndpoint
+                )
+
+                // Re-check after async switch. If the user saved manual mode while
+                // switchToEndpoint was in flight, restore their chosen manual server.
+                // Read from selectedServers because the manual save path writes it first.
+                guard userStoredPreferences.selectedServers()?.mode == .automatic else {
```
Chlup, line 530: This is strange. And I don't completely get it. This looks like the code does async endpoint switch and then test if the user didn't do switch to manual during the async switch. Do I get it right? And what if the user does switch to automatic during async switch to manual? 

Maybe this should be handled a little better. When user does switch in UI current "switch node" operation should be either canceled or code should wait until it's done and then do new switch. Or maybe block UI for user when the actual switch is happening?

Chlup: It looks that this code doesn't even run when user does changed. It runs automatically after start.

secant/Sources/Features/ServerSetup/ServerSetupStore.swift
```diff
@@ -152,86 +224,209 @@ struct ServerSetup {
...
+                return .none
+
             case .setServerTapped:
-                guard state.initialServer != state.selectedServer || state.selectedServer == String(localizable: .serverSetupCustom) else {
+                guard state.hasChanges else {
                     return .none
                 }
-                
+
                 state.isUpdatingServer = true
```
Chlup, line 261: This disables only save button. So during save of server selection user can switch to another server and end up in situation when user can't save new selection. This flag should disable all the UI on the screen.

https://github.com/user-attachments/assets/ca3c526e-4734-4722-bf77-6e74418d6a65

secant/Sources/Features/ServerSetup/ServerSetupView.swift
```diff
@@ -11,264 +11,274 @@ import ComposableArchitecture
...
-                                                .background {
-                                                    RoundedRectangle(cornerRadius: Design.Radius._md)
-                                                        .fill(Design.Surfaces.bgPrimary.color(colorScheme))
-                                                }
-                                                .overlay {
-                                                    RoundedRectangle(cornerRadius: Design.Radius._md)
-                                                        .stroke(Design.Inputs.Default.stroke.color(colorScheme), lineWidth: 1)
-                                                }
-                                                .padding(.vertical, 8)
+                                            TextField(
```
Chlup, line 221: I think this text field could have .autocorrectionDisabled() and .keyboardType(.URL). Autocorrection for URL doesn't make sense.

secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerInterface.swift
czarcas7ic: https://github.com/zodl-inc/zodl-ios/commit/0ea125cad7564e2876eae2c9be0afd11411fd064

Changed the coordinator to queue endpoint switches so only one SDK switch runs at a time. Automatic switches now also recheck that we’re still in automatic mode before they run.

secant/Sources/Dependencies/ZcashSDKEnvironment/ZcashSDKEnvironmentLiveKey.swift
czarcas7ic: https://github.com/zodl-inc/zodl-ios/commit/7f7cae3951e9de6179d91d477ac169fd5c824927
https://github.com/zodl-inc/zodl-ios/commit/c9aba3ece1df9c4d070f7081c63d70bd76c6cd1f

Fixed here and making note to do this in future PRs, thanks.

secant/Sources/Dependencies/ZcashSDKEnvironment/ZcashSDKEnvironmentLiveKey.swift
czarcas7ic: https://github.com/zodl-inc/zodl-ios/commit/7f7cae3951e9de6179d91d477ac169fd5c824927
https://github.com/zodl-inc/zodl-ios/commit/c9aba3ece1df9c4d070f7081c63d70bd76c6cd1f

Addressed here and for all other locations.

secant/Sources/Features/ServerSetup/ServerSetupStore.swift
czarcas7ic: The difference is that `RootStore` uses a lightweight automatic refresh, while `ServerSetupStore` uses the deeper setup screen pass.

For Root, the SDK first does cheap checks across the full endpoint list, then runs the short 1-block fetch check on the top 3 candidates and uses the best returned endpoint. For Server Setup, it uses a 60 second window, 100 blocks, and returns 3 recommendations because the user is already on that screen.

I added comments to make that distinction clearer. If you think we should use the same benchmark everywhere, I can collapse them.

https://github.com/zodl-inc/zodl-ios/commit/7c4a0feec09b519e9525b6ec2037d3516b7767f0

secant/Sources/Features/Root/RootStore.swift
czarcas7ic: https://github.com/zodl-inc/zodl-ios/commit/4a77e250949c83284c760c97246f859d52025f65

> It looks that this code doesn't even run when user does changed. It runs automatically after start.

This was intended since this Root path is automatic endpoint refresh work, not the user save handler. It can run after startup, after foregrounding, or when Root reuses completed Server Setup evaluation results. I renamed the action and helper methods to `automaticEndpointRefresh...` and added comments to make that clearer.

> Maybe this should be handled a little better. When user does switch in UI current "switch node" operation should be either canceled or code should wait until it's done and then do new switch.

That part is handled by the earlier concurrency changes here:
https://github.com/zodl-inc/zodl-ios/commit/0ea125cad7564e2876eae2c9be0afd11411fd064

Server Setup save now cancels pending automatic benchmark and refresh work, automatic refresh rechecks persisted mode before switching, and SDK endpoint switches go through the serialized coordinator so they do not run in parallel.

I didn’t treat this Root path as the place to block the UI. That belongs to the Server Setup save flow, since that is where user initiated mode changes happen.

secant/Sources/Features/ServerSetup/ServerSetupStore.swift
czarcas7ic: https://github.com/zodl-inc/zodl-ios/commit/ae26de092588af9cd27d0369301ddc564df9b5b0

Server Setup now disables the form while `isUpdatingServer` is true, so mode selection, server selection, refresh, and the custom URL field can’t be changed during save. I also added reducer guards for the direct actions and a regression test for that state.

secant/Sources/Features/ServerSetup/ServerSetupView.swift
czarcas7ic: https://github.com/zodl-inc/zodl-ios/commit/ec944eebbb9c2b96652594b3387617d916cb1f5c

Disabled autocorrection on the custom server field and switched it to the URL keyboard.

secant/Sources/Dependencies/ZcashSDKEnvironment/ZcashSDKEnvironmentLiveKey.swift
Chlup: 👍

secant/Sources/Dependencies/ZcashSDKEnvironment/ZcashSDKEnvironmentLiveKey.swift
Chlup: 👍

secant/Sources/Features/ServerSetup/ServerSetupStore.swift
Chlup: 👍

secant/Sources/Features/ServerSetup/ServerSetupView.swift
Chlup: 👍

secant/Sources/Features/ServerSetup/ServerSetupStore.swift
Chlup: Ok. Let's go with this. Thanks for explanation.

secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerInterface.swift
Chlup: Can you please put `EndpointSwitchCoordinator` to it's own file? `SDKSynchronizerInterface.swift` should be really just of `SDKSynchronizerInterface`. 

I was thinking about `EndpointSwitchCoordinator` and I tried to come up with something little bit less complex. What do you think about this?
```
actor EndpointSwitchCoordinator {
    private var pendingTask: Task<Void, Error>?

    /// Serializes endpoint switches so only one runs at a time.
    ///
    /// Each invocation chains behind the previous one in submission order. Inside its slot,
    /// the queued task:
    ///
    /// 1. Waits for its predecessor to finish (predecessor's outcome doesn't propagate to us).
    /// 2. Re-runs `shouldProceed`; returning `false` short-circuits cleanly — no error, no restore.
    /// 3. Calls `performSwitch(endpoint)`.
    /// 4. Surfaces any cancellation that arrived while the SDK call ran.
    ///
    /// If any step throws — including `CancellationError` — `previousEndpoint` is resolved
    /// (when provided) and `performSwitch` is called once more with that endpoint to roll
    /// back. A failed restore is logged; the original error always propagates. Restoring
    /// inside a cancelled task is best-effort: the SDK call may itself throw cancellation,
    /// in which case the failure is logged and we leave the SDK in whatever state it landed.
    ///
    /// `previousEndpoint` is a closure so the restore target is computed at the time of
    /// failure rather than at submission time, which would otherwise go stale when other
    /// switches ran ahead of us in the queue.
    ///
    /// Caller cancellation is forwarded into the queued task via `withTaskCancellationHandler`.
    /// Note: waiting on the predecessor (`await previous.value`) doesn't observe our own
    /// cancellation, so a hung predecessor delays our cancellation too.
    ///
    /// - Parameters:
    ///   - endpoint: Endpoint to switch to.
    ///   - previousEndpoint: Closure returning the endpoint to restore on failure. Evaluated
    ///     in the catch block. Pass `nil` (default) to skip restoration.
    ///   - shouldProceed: Re-check executed after acquiring the slot. Returning `false` exits
    ///     cleanly without throwing. Defaults to always proceeding.
    ///   - performSwitch: The actual switching work — typically the SDK's `switchToEndpoint`.
    /// - Throws: `CancellationError` if the caller's task is cancelled, or any error thrown
    ///   by `performSwitch`. By the time the caller sees the error, restore has already been
    ///   attempted (if `previousEndpoint` was provided).
    func switchToEndpoint(
        _ endpoint: LightWalletEndpoint,
        previousEndpoint: (@Sendable () async -> LightWalletEndpoint)? = nil,
        shouldProceed: @escaping @Sendable () async -> Bool = { true },
        performSwitch: @escaping @Sendable (LightWalletEndpoint) async throws -> Void
    ) async throws {
        try Task.checkCancellation()

        let previous = pendingTask
        let task = Task {
            // Wait for our predecessor to release the slot. Its outcome doesn't gate us.
            _ = try? await previous?.value

            do {
                // Cancellation may have arrived while we waited for the predecessor.
                try Task.checkCancellation()
                guard await shouldProceed() else { return }
                // shouldProceed suspended; re-check before the SDK call.
                try Task.checkCancellation()
                try await performSwitch(endpoint)
                // Surface late cancellation so the caller sees CancellationError even
                // if the SDK call ignored it.
                try Task.checkCancellation()
            } catch {
                if let resolved = await previousEndpoint?() {
                    do {
                        try await performSwitch(resolved)
                    } catch {
                        LoggerProxy.error("[EndpointSwitch] Failed to restore endpoint: \(error)")
                    }
                }
                throw error
            }
        }
        pendingTask = task

        try await withTaskCancellationHandler {
            try await task.value
        } onCancel: {
            task.cancel()
        }
    }
}
```

secant/Sources/Features/Root/RootStore.swift
Chlup: Ok. Thanks for explanation. Let's go with this.

secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerInterface.swift
czarcas7ic: I took this direction in https://github.com/zodl-inc/zodl-ios/commit/f613c57168dc43f34f2d31ee576f2fda53ddb03a. I split `EndpointSwitchCoordinator` into its own file and switched it to the `pendingTask` chain.

I kept the restore guard so we only roll back after this task actually completed the target SDK switch, and made `previousEndpoint` resolve lazily so the restore target is fresh. I also added coverage for queued cancellation, `shouldProceed == false`, post-switch cancellation restore, and no restore when the target switch throws before the simulated endpoint mutation.

Chlup: 👍

Chlup:
Finale code for this feature is in #1804

PR #1751, opened 2026-05-12
[MOB-1039] Add multi-server transaction submission

czarcas7ic:
Closes: #1677

## Summary

- Creates transactions locally and submits them through the shared multi-server submission path.
- Automatic mode broadcasts each created transaction to all known endpoints. Manual mode submits only to the selected endpoint.
- The first successful endpoint wins, and remaining endpoint calls get a short grace period for propagation.
- If no endpoint responds before the timeout and drain window, the send is treated as pending with timeout-specific copy because the transaction may still have been broadcast.
- First non-gRPC rejection details are preserved, partial multi-transaction submission routes to the support/error flow, and regular send, Keystone send, auto-shielding, Flexa, and swap-and-pay all use the same create-and-submit infrastructure.

## Dependency Note

This still depends on the unreleased Swift SDK submit-plan retry work. The project is pinned to a commit until the released SDK tag is available. Before merge, replace that pin with the released tag and repeat final build validation.

## Testing

- Physical-device validation covered regular and Keystone send paths, automatic mode with all known servers, and manual mode with one selected server.
- Added automated coverage for multi-transaction submission, all-server rejection, endpoint selection, direct submission helper behavior, first-success behavior, timeout pending behavior, partial failure support routing, Swap/Pay timeout pending routing, and the PCZT broadcaster/reset path.
- `jq empty secant/Resources/Localizable.xcstrings`
- `git diff --check`
- `xcodebuild build -project secant.xcodeproj -scheme secant-testnet -destination 'generic/platform=iOS Simulator' -skipMacroValidation -skipPackagePluginValidation CODE_SIGN_IDENTITY=- CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO IPHONEOS_DEPLOYMENT_TARGET=16.0`
- GitHub `Pull Request / unit_tests` passed on `4962c186`.

Video from testing: https://screen.studio/share/1mgC0NYH

The video logs show why this helps in practice. One endpoint returned an Orchard anchor error while other endpoints successfully submitted the transaction.


czarcas7ic:
The one merge blocking thing here is that we still need the [Swift SDK](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1737) PR merged and tagged and replaced in this PR, but am opening it up for initial review anyway.

secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerLive.swift
```diff
@@ -338,6 +385,332 @@ extension SDKSynchronizerClient: DependencyKey {
...
+                if let server = config.servers.first {
+                    return [server.endpoint(streamingCallTimeoutInMillis: streamingTimeout)]
+                }
+            }
+        }
+
+        return [zcashSDKEnvironment.endpoint()]
+    }
+
+    static func submitToAllEndpoints(
```
Chlup, line 588: There is lot of things happening in this function that are either wrong or not ideal. Or maybe is there something that I don't see?

 I think this function can look like this:
```
static func submitToAllEndpoints(
    rawTx: Data,
    endpoints: [LightWalletEndpoint],
    logPrefix: String,
    responseTimeout: Duration = MultiServerSubmissionTiming.firstResponseTimeout,
    timeoutDrainDelay: Duration = MultiServerSubmissionTiming.timeoutDrainDelay,
    recordSubmitFailure: @escaping @Sendable (String, Int, String) async -> Void = { _, _, _ in },
    recordTimeout: @escaping @Sendable () async -> Void = {},
    submit: @escaping @Sendable (Data, LightWalletEndpoint) async throws -> Void
) async -> String? {
    guard !endpoints.isEmpty else { return nil }

    let acceptedSubmitStore = AcceptedSubmitStore()

    return await withTaskGroup(of: SubmitResult.self) { group in
        for (offset, endpoint) in endpoints.enumerated() {
            let endpointLabel = "endpoint \(offset + 1)"
            group.addTask {
                do {
                    try await submit(rawTx, endpoint)
                    await acceptedSubmitStore.record(endpointLabel)
                    return .endpoint(endpointLabel)
                } catch TransactionEncoderError.submitError(let code, let message) {
                    await recordSubmitFailure(endpointLabel, code, message)
                    return .endpoint(nil)
                } catch {
                    return .endpoint(nil)
                }
            }
        }

        group.addTask {
            try? await Task.sleep(for: responseTimeout + timeoutDrainDelay)
            return .timedOut
        }

        var failedCount = 0
        for await result in group {
            switch result {
            case .endpoint(let winner?):
                group.cancelAll()
                return winner
            case .endpoint(nil):
                failedCount += 1
                if failedCount >= endpoints.count {
                    group.cancelAll()
                    return nil
                }
            case .timedOut:
                let recorded = await acceptedSubmitStore.recordedEndpointLabel()
                if recorded == nil { await recordTimeout() }
                group.cancelAll()
                return recorded
            case .graceExpired:
                break // remove this case
            }
        }
        return nil
    }
}
```

secant/Sources/Features/Root/RootDestination.swift
```diff
@@ -115,12 +115,18 @@ extension Root {
                         let network = zcashSDKEnvironment.network.networkType
                         let spendingKey = try derivationTool.deriveSpendingKey(seedBytes, zip32AccountIndex, network)
                         
-                        let result = try await sdkSynchronizer.createProposedTransactions(proposal, spendingKey)
+                        let result = try await sdkSynchronizer.createAndSubmitProposedTransactions(proposal, spendingKey)
                         
                         switch result {
-                        case .partial:
+                        case let .grpcFailure(txIds, _, _):
```
Chlup: I think when using new code there is a difference between `.grpcFailure` without reason and with reason. And this handling can result in situation when transactions are not send to any backend and we still tell Flexa that transactions went through.

Concrete scenario:
Preconditions:
  - User is in Flexa flow, taps Pay
  - User is in .automatic mode (8 mainnet endpoints) — but same outcome
    occurs in .manual mode with 1 endpoint that's unreachable
  - Phone has just lost Wi-Fi (no IP route)

Sequence:
  1. flexaOnTransactionRequest fires
  2. sdkSynchronizer.proposeTransfer  → succeeds (pure local computation)
  3. sdkSynchronizer.createAndSubmitProposedTransactions(proposal, key):
     a. synchronizer.broadcaster.createProposedTransactions:
        - rustBackend.createProposedTransactions inserts tx into DataDb
        - tx is now in DB with txId T
     b. SDKSynchronizerClient.createAndSubmitTransactions:
        - submitToAllEndpoints with 8 endpoints
        - Each broadcaster.submit:
            try await ... .submit(spendTransaction: rawTx)
              → grpc connect fails (no network) → throws
              → wrapped in ZcashError.serviceSubmitFailed(...)
          → falls into generic catch in submitToAllEndpoints
          → returns .endpoint(nil), no record
        - All 8 return .endpoint(nil) within ~1-3s
        - failedCount==8 → cancelAll → continuation.resume(nil)
        - Timer task cancelled before 30s sleep finishes
     c. createAndSubmitTransactions sees:
        - accepted=nil
        - submitTimeoutStore.didTimeOut()=false   (timer never fired)
        - submitFailureStore.recordedFailure()=nil
        - returns .grpcFailure(txIds:[T], description:nil, reason:nil)
  4. RootDestination case .grpcFailure(txIds, _, _):
     - txIds.last = T
     - txIdExists(T) → checks allTransactions() → tx is in DataDb → true
     - flexaHandler.transactionSent(commerceSessionId, T) is called

Result: Flexa is told the payment went through. Merchant sees a green
"paid" status. User walks out with merchandise. Zero servers in the
Zcash network ever saw the transaction.

secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerLive.swift
czarcas7ic: My suggested implementation is doing something slightly different than race all tasks and return the first winner, which would stop as soon as one endpoint accepts the transaction. That is what your suggested version does when it calls `group.cancelAll()` immediately after the first success.

For this PR, my intent was to make it that when the first endpoint succeeds, it resumes the caller immediately so the UI can move forward, but it does not cancel the other endpoint submissions right away. Instead, it starts a grace timer and only cancels the remaining endpoint tasks after that grace window expires:

https://github.com/zodl-inc/zodl-ios/blob/a893c64ab23aaaac55705ad0b7f23ea51c48c909/secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerLive.swift#L660-L683

This was to separate two concerns:

- return the winning endpoint to the caller as soon as we have one
- keep already started submissions alive briefly so other servers still get a chance to receive and respond to the transaction

Your suggestion is much simpler, but it changes that behavior. It makes the first success cancel the remaining endpoint submissions immediately, which removes the best effort propagation window this PR is trying to achieve.

What are your thoughts here? I agree that I need to add more explanation in code since it was not clear during review what I was going for, but will wait until we come to a conclusion on what the end goal should be. Personally my fear is some faulty endpoint immediately responding with success before getting to submit to the other endpoints, but maybe this fear is unfounded and taking your simpler approach is just the right move.

secant/Sources/Features/Root/RootDestination.swift
czarcas7ic: I pushed https://github.com/zodl-inc/zodl-ios/commit/b350edb7f40749d72e838aa3f635b738637aca3f so the Flexa path now only calls `transactionSent` after `.success` from the submit path, and it fails `.grpcFailure`, `.failure`, and `.partial` even if the txid exists locally. I also added coverage for the exact unreasoned `.grpcFailure(txIds:)` case, timeout `.grpcFailure`, success, and success with a missing local txid.

I don’t think we can conclude from this repo alone that this would let someone walk out with merchandise, since Flexa side validation is outside this codebase. But the app still should not report a payment as sent without an accepted submit result.

secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerLive.swift
Chlup: Thanks for explanation. Now I understand your logic. I also now understand why you used so complicated way how to structure logic and cancelation support (task group inside task inside continuation inside cancellation handler). So please add some documentation to the final form of this function. It is bit complex and not clear on the first sight. That may be applied to some other methods as well. 

I am fine with internal grace period logic. I just didn't understand it. I didn't like `var hasResumed = false` and `SubmitTaskCancellation` more. That variable isn't Swift 6 compliant and we are one PR from switching to Swift 6. And `SubmitTaskCancellation` can be replaced with one `OSAllocatedUnfairLock`. If you would like to keep `SubmitTaskCancellation` for some reason just at least please replace `NSLock` with `OSAllocatedUnfairLock`.

To keep your logic and support Swift 6 I think code could look something like this:

```
static func submitToAllEndpoints(
    rawTx: Data,
    endpoints: [LightWalletEndpoint],
    logPrefix: String,
    graceDelay: Duration = MultiServerSubmissionTiming.postAcceptanceGraceDelay,
    responseTimeout: Duration = MultiServerSubmissionTiming.firstResponseTimeout,
    timeoutDrainDelay: Duration = MultiServerSubmissionTiming.timeoutDrainDelay,
    recordSubmitFailure: @escaping @Sendable (String, Int, String) async -> Void = { _, _, _ in },
    recordTimeout: @escaping @Sendable () async -> Void = {},
    submit: @escaping @Sendable (Data, LightWalletEndpoint) async throws -> Void
) async -> String? {
    guard !endpoints.isEmpty else { return nil }

    let endpointCount = endpoints.count
    let acceptedSubmitStore = AcceptedSubmitStore()
    let workerTask = OSAllocatedUnfairLock<Task<Void, Never>?>(initialState: nil)

    return await withTaskCancellationHandler {
        await withCheckedContinuation { (continuation: CheckedContinuation<String?, Never>) in
            let task = Task {
                await withTaskGroup(of: SubmitResult.self) { group in
                    for (offset, endpoint) in endpoints.enumerated() {
                        let label = "endpoint \(offset + 1)"
                        group.addTask {
                            do {
                                try await submit(rawTx, endpoint)
                                await acceptedSubmitStore.record(label)
                                LoggerProxy.event("\(logPrefix) \(label) SUCCESS.")
                                return .endpoint(label)
                            } catch TransactionEncoderError.submitError(let code, let message) {
                                await recordSubmitFailure(label, code, message)
                                LoggerProxy.warn("\(logPrefix) \(label) REJECTED: \(code).")
                                return .endpoint(nil)
                            } catch {
                                LoggerProxy.warn("\(logPrefix) \(label) FAILED.")
                                return .endpoint(nil)
                            }
                        }
                    }

                    group.addTask {
                        try? await Task.sleep(for: responseTimeout + timeoutDrainDelay)
                        LoggerProxy.error("\(logPrefix) Timed out waiting for any server to respond.")
                        return .timedOut
                    }

                    var winner: String?
                    var failedCount = 0

                    drain: for await event in group {
                        switch event {
                        case .endpoint(let label?):
                            guard winner == nil else { continue }
                            winner = label
                            continuation.resume(returning: label)
                            group.addTask {
                                try? await Task.sleep(for: graceDelay)
                                return .graceExpired
                            }

                        case .endpoint(nil):
                            failedCount += 1
                            guard winner == nil else { continue }
                            if failedCount >= endpointCount {
                                continuation.resume(returning: nil)
                                break drain
                            }

                        case .timedOut:
                            guard winner == nil else { continue }
                            let recorded = await acceptedSubmitStore.recordedEndpointLabel()
                            if recorded == nil { await recordTimeout() }
                            continuation.resume(returning: recorded)
                            break drain

                        case .graceExpired:
                            break drain
                        }
                    }

                    group.cancelAll()
                }
            }
            workerTask.withLock { $0 = task }
            if Task.isCancelled { task.cancel() }
        }
    } onCancel: {
        workerTask.withLock { $0?.cancel() }
    }
}
```

secant/Sources/Features/Root/RootDestination.swift
Chlup: 👍

secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerLive.swift
czarcas7ic: Thanks, that makes sense. Since you're fine with the internal grace period logic, I kept that behavior intact and pushed https://github.com/zodl-inc/zodl-ios/commit/4962c1860d6eb50d9db95675abc2151076a5fa54 with the structure changes.

That removes `hasResumed` and the custom `SubmitTaskCancellation`, uses `OSAllocatedUnfairLock` for the task holder, marks `submit` as `@Sendable`, and separates timer cancellation from a real timeout so canceled sleeps don't log or record false timeouts. I also added coverage for the timeout and cancellation paths.

Chlup: 👍

Chlup:
This was finally implemented in #1828

PR #1756, opened 2026-05-14
Show Keystone authorization failures

czarcas7ic:
Keystone delegation failures were not being surfaced to the vote submission UI, so users could get stuck on “Authorizing vote” with no error. This marks the submission as failed when that happens, clears the loading state, and shows the existing authorization failure message.

Validation:
- `git diff --check`
- `xcodebuild build -project secant.xcodeproj -scheme zashi-internal -destination 'platform=iOS Simulator,name=iPhone 17 Pro' -skipMacroValidation -skipPackagePluginValidation IPHONEOS_DEPLOYMENT_TARGET=16.0`
- Focused `secant-testnet` test invocation did not run because the existing `secantTests` target fails to compile on `VotingSessionTests.swift` importing missing `VotingModels`.

### DIFF
--- a/secant/Sources/Features/Voting/VotingStore+Delegation.swift
+++ b/secant/Sources/Features/Voting/VotingStore+Delegation.swift
@@ -976,6 +976,14 @@ extension Voting {
             }
             state.delegationProofStatus = .failed(userMessage)
             state.isDelegationProofInFlight = false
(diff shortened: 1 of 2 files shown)


LukasKorba (approved):
LGTM

PR #1761, opened 2026-05-18
Update voting SDK alpha2 API integration

greg0x:
## Why

The voting flow was pinned to an older `zcash-swift-wallet-sdk` revision. Moving to `909c7c5` picks up the libzcashlc 2.6.0-alpha.2 artifact, which changes the hotkey API: `generateHotkey` no longer takes a `roundId`. Hotkeys are now derived from the seed alone, independent of the voting round, so the call sites have to be adapted to compile against the new SDK.

## What changed

- Bump the `ZcashLightClientKit` pin to `zcash/zcash-swift-wallet-sdk@909c7c5743f14a211f9f69f97ed08c05b55c2cd5` (`secant.xcodeproj/project.pbxproj` and `Package.resolved`).
- Adapt the two `generateHotkey` call sites in `VotingCryptoClientLiveKey.swift` to the new `generateHotkey(seed:)` signature — in the `generateHotkey` dependency closure and in `buildVotingPczt`. The closure's now-unused `roundId` argument is dropped to `_`.

## Scope

Dependency pin + call-site adaptation only. No bundle ID, entitlements, signing/development team, iCloud container, staging config, archive output, or local package wiring changes.

## Verification

```
xcodebuild -resolvePackageDependencies \
  -project secant.xcodeproj \
  -scheme zashi-internal \
  -skipPackagePluginValidation \
  -skipMacroValidation
```

### DIFF
--- a/secant.xcodeproj/project.pbxproj
+++ b/secant.xcodeproj/project.pbxproj
@@ -9465,7 +9465,7 @@
 			repositoryURL = "https://github.com/zcash/zcash-swift-wallet-sdk.git";
 			requirement = {
 				kind = revision;
-				revision = b3675ba9ecadd2b777f21d8b770e88cba294d26e;
+				revision = 909c7c5743f14a211f9f69f97ed08c05b55c2cd5;
 			};
 		};
 /* End XCRemoteSwiftPackageReference section */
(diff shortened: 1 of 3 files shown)


LukasKorba (approved):
utACK

PR #1765, opened 2026-05-19
Add smoke test suite and CI workflow

zcash-harry:
## Summary

Adds automated unit tests covering four critical areas of the app, plus a GitHub Actions CI workflow to run tests on every pull request.

## CI Workflow

- New `.github/workflows/pull-request.yml` runs `secantTests` on every PR
- Uses `macos-15` runner with Xcode 16.4, targeting iPhone 17 Pro simulator
- Uploads test results as artifacts for easy debugging
- Skips CI for docs-only changes (README, LICENSE, docs/, issue templates)

## Known Issue: Xcode 26 Beta Compatibility

Local builds on Xcode 26 beta fail in `SwapAndPayStore.swift` with:
- "Compiling for iOS 15.0, but module 'ComposableArchitecture' was compiled for iOS 16.0"
- "'sleep(for:tolerance:clock:)' is only available in iOS 16.0 or newer"

This is a pre-existing issue — the project deployment target is iOS 15.0 but TCA (ComposableArchitecture) now requires iOS 16.0+. CI uses Xcode 16.4 and is unaffected. The team will need to either bump the deployment target to iOS 16.0 or pin TCA to an iOS 15-compatible version.

### QRCodeGeneratorTests (23 tests)
- **Basic generation**: valid input returns correct pixel array size, output contains both black and white pixels, empty string throws error
- **Determinism**: same input produces identical output, different inputs produce different output
- **Size variations**: small (64px), large (1024px), minimum viable (25px), size 1 edge case
- **Content types**: Zcash t-address, ZIP 321 URI, URI with multiple params, long unified address (300+ chars), special characters
- **Edge cases**: size 0 returns empty array, negative size throws error
- **Round-trip decode** (7 tests): generates QR → decodes with CIDetector → verifies decoded string matches original input. Covers simple address, ZIP 321 URI, URI with all params, URL-encoded content, long unified address, single character, numeric string

### CurrencyConversionTests (24 tests)
- **Init & rounding**: fiat amount rounds to 2 decimal places using banker's rounding (HALF_EVEN), zatoshi preserved exactly
- **Zatoshi → fiat conversion**: 1 ZEC at $150 = $150.00, fractional ZEC converts correctly, zero zatoshi = $0.00
- **Fiat → zatoshi conversion**: $1.00 at $150/ZEC = correct zatoshi, fractional dollar converts correctly, zero fiat = 0 zatoshi
- **Round-trip**: zatoshi → fiat → zatoshi preserves original value within rounding tolerance
- **Edge cases**: very small zatoshi (1 zatoshi), very large amounts (1B ZEC), high exchange rate ($99,999), low exchange rate ($0.01)
- **CurrencyId enum**: ZEC and USD have correct raw values, all cases accounted for
- **Equatable**: identical states are equal, different fiat amounts are not equal

### DeeplinkURLParsingTests (9 tests)
- **Basic parsing**: `zcash:` URI extracts correct address, full URI with amount and memo parses all fields
- **Query parameters**: amount parsed as Decimal, memo preserved exactly, label and message fields extracted
- **Defaults**: missing params default to nil, address-only URI works
- **Encoding**: URL-encoded memo (`%20`) decoded correctly
- **Network**: mainnet address detected as mainnet type

### MemoByteCountingTests (12 tests)
- **Byte counting**: empty text = 0 bytes, ASCII "Hello" = 5 bytes, emoji "🎉" = 4 bytes, 10 emojis = 40 bytes, Japanese "日本語" = 9 bytes, Spanish "café" = 5 bytes (accented é is 2 bytes), mixed content "Hi 🎉 日本" = 14 bytes
- **Validation**: exactly at 512-byte limit is valid, emoji pushing over limit (510 ASCII + 🎉) is invalid, emoji exactly at limit (508 ASCII + 🎉) is valid, empty text is valid
- **Display**: `charLimitText` shows remaining bytes (e.g. "507/512" for "Hello", "508/512" for "🎉", "0/512" at limit)

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,0 +1,47 @@
+name: Pull Request
+
+on:
+  pull_request:
+    paths-ignore:
+      - '.github/ISSUE_TEMPLATE/*'
+      - '.github/pull_request_template.md'
+      - 'LICENSE'
+      - 'README.md'
+      - 'docs/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  unit_tests:
+    runs-on: macos-15
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Resolve packages
+        run: |
+          xcodebuild -project secant.xcodeproj \
+            -scheme secant-testnet \
+            -resolvePackageDependencies
+
+      - name: Run unit tests
+        run: |
+          xcodebuild test \
+            -project secant.xcodeproj \
+            -scheme secant-testnet \
+            -destination 'platform=iOS Simulator,name=iPhone 17 Pro,OS=latest' \
+            -resultBundlePath TestResults.xcresult \
+            -only-testing:secantTests \
+            CODE_SIGNING_ALLOWED=NO | xcpretty --color
+
+      - name: Upload test results
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: TestResults
+          path: TestResults.xcresult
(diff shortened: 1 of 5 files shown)


LukasKorba (approved):
utACK

PR #1766, opened 2026-05-20
[MOB-1105] Replace os.Logger with LoggerProxy in voting/background paths

LukasKorba:
## Summary
Mechanical swap of direct `os.Logger` usage in voting and background-task code paths over to the project's `LoggerProxy`, so logs are gated by DEBUG (consistent with the rest of the codebase) and don't leak voting subsystem details in release builds.

This is **PR 1 of 4** in the decomposition of the original [#1764](https://github.com/zodl-inc/zodl-ios/pull/1764) per reviewer request. Target base: `MOB-1105-integration`.

## Files touched
- `BackgroundTaskClient/BackgroundTaskClientLiveKey.swift`
- `VotingAPIClient/ServerHealthTracker.swift`
- `VotingAPIClient/VotingAPIClientLiveKey.swift`
- `Voting/VotingStore.swift` + `VotingStore+{Delegation,Navigation,Session,Submission}.swift`

Net: -126/+115 lines, all mechanical (declarations removed, call sites renamed, `, privacy: .public` modifiers stripped since `LoggerProxy` doesn't need them).

## Test plan
- [ ] Build passes (verified locally)
- [ ] No behavioral change expected — pure logging refactor

PR #1764, opened 2026-05-19
MOB-1105 Coinholder Polling fixes

LukasKorba:
This code review checklist is intended to serve as a starting point for the author and reviewer, although it may not be appropriate for all types of changes (e.g. fixing a spelling typo in documentation).  For more in-depth discussion of how we think about code review, please see [Code Review Guidelines](../blob/main/CODE_REVIEW_GUIDELINES.md).

# Author

- [ ] Self-review: Did you review your own code in GitHub's web interface? Code often looks different when reviewing the diff in a browser, making it easier to spot potential bugs.
- [ ] Does the code abide by the [Coding Guidelines](../blob/main/docs/CODING_GUIDELINES.md)?
- [ ] Automated tests: Did you add appropriate automated tests for any code changes?
- [ ] Code coverage: Did you check the code coverage report for the automated tests?  While we are not looking for perfect coverage, the tool can point out potential cases that have been missed.
- [ ] Documentation: Did you update Docs as appropiate? (E.g [README.md](../blob/main/README.md), etc.)
- [ ] Run the app: Did you run the app and try the changes? 
- [ ] Did you provide Screenshots of what the App looks like before and after your changes as part of the description of this PR? (only applicable to UI Changes)
- [ ] Rebase and squash: Did you pull in the latest changes from the main branch and squash your commits before assigning a reviewer? Having your code up to date and squashed will make it easier for others to review. Use best judgement when squashing commits, as some changes (such as refactoring) might be easier to review as a separate commit.

# Reviewer

- [ ] Checklist review: Did you go through the code with the [Code Review Guidelines](../blob/main/CODE_REVIEW_GUIDELINES.md) checklist?
- [ ] Ad hoc review: Did you perform an ad hoc review?  _In addition to a first pass using the code review guidelines, do a second pass using your best judgement and experience which may identify additional questions or comments. Research shows that code review is most effective when done in multiple passes, where reviewers look for different things through each pass._
- [ ] Automated tests: Did you review the automated tests?
- [ ] Manual tests: Did you review the manual tests?_You will find manual testing guidelines under our [manual testing section](../blob/main/docs/testing/manual_testing)_
- [ ] How is Code Coverage affected by this PR? _We encourage you to compare coverage before and after your changes and when possible, leave it in a better place. [Learn More...](../blob/main/docs/testing/local_coverage.md)_
[...]

PR #1767, opened 2026-05-20
[MOB-1105] Encrypted voting metadata + Keychain hotkey + Reset wipe

LukasKorba:
## Summary
Moves voting drafts and per-round vote records out of plaintext UserDefaults and into an encrypted, per-account local file. The voting hotkey moves into the iOS Keychain, scoped per AccountUUID. Reset Zashi now wipes everything voting-related so no state survives across resets.

This is **PR 2 of 4** in the decomposition of the original [#1764](https://github.com/zodl-inc/zodl-ios/pull/1764).

**Depends on:** PR #1766 (logger swap). This branch is stacked on top of `MOB-1105-pr-a-logger`. Until #1766 merges, this PRs diff will include both PR A + PR B commits; after #1766 merges into integration, this PR will auto-update to show only PR B.

## What's in this PR
- **Encrypted voting metadata file** — VotingMetadataProvider* + VotingMetadata* storage. ChaChaPoly-AEAD, key derived via HKDF from the wallets user-metadata encryption material. Excluded from iCloud sync.
- **Keychain hotkey** — WalletStorage/WalletStorageInterface/WalletStorageLiveKey gain methods to read/write/delete the per-account voting hotkey. Stored with kSecAttrAccessibleWhenUnlockedThisDeviceOnly.
- **Reset Zashi wipe** — RootInitialization removes voting drafts, vote records, votingConfigOverrideURL, votingCustomChains, the on-disk voting.sqlite3, plus the encrypted per-account votingMetadata file. Belt-and-suspenders sweep of legacy plaintext UserDefaults keys.
- **Submission cancel** — VotingStore+Submission cancels in-flight vote submission effects on flow dismiss so a backgrounded submission cant keep running against a wallet the user just walked away from.

## What's NOT in this PR
- Tor routing for voting traffic (PR C)
- HTTPS-only dynamic config URL enforcement (PR C)
- ServerHealthTracker Tor-aware fetcher + force-unwrap hardening (PR C)
- Navigation rewrite + UI polish (PR D)

## Test plan
- [ ] Manual: vote in one wallet, switch accounts, vote in another — drafts must NOT leak across accounts
- [ ] Manual: Reset Zashi, restore — voting state must be empty (no drafts, no hotkey, no DB rows)
- [ ] Manual: kill the app mid-submission — no orphaned background tasks


greg0x:
Noting here, that `voting hotkey moves into the iOS Keychain,`: The voting hotkey was already stored in the iOS Keychain; this branch changes it to be keyed/scoped by AccountUUID.

secant/Sources/Dependencies/VotingMetadataProvider/VotingMetadataStorage.swift
```diff
@@ -0,0 +1,286 @@
...
+
+        let filename = try filenameForEncryptedFile(account: account)
+        let fileURL = documentsDirectory.appendingPathComponent(filename)
+
+        try? FileManager.default.removeItem(at: fileURL)
+    }
+
+    func store(account: Account) throws {
+        guard let documentsDirectory = FileManager.default
+            .urls(for: .documentDirectory, in: .userDomainMask)
```
greg0x, line 82: I believe this file is included in iCloud Backup by default because it is written under the app’s Documents directory. Is that intended, or should we mark the voting metadata file with `isExcludedFromBackupKey` after each write?

secant/Sources/Features/Voting/VotingStore+Helpers.swift
```diff
@@ -23,89 +25,103 @@ extension Voting {
...
-              let votedAtUnix = raw["votedAt"] as? Double,
-              let weight = (raw["votingWeight"] as? NSNumber)?.uint64Value else {
-            return nil
+    // MARK: - Vote records
+
+    static func persistVoteRecord(_ record: VoteRecord, roundId: String, account: Account?) {
+        @Dependency(\.votingMetadata) var votingMetadata
+        votingMetadata.setRecord(record.persisted, roundId)
+        if let account {
+            try? votingMetadata.store(account)
```
greg0x: This updates in-memory first, and If the persistence fails silently — has an increased chance because of the encryption — this could cause in memory state + UI and persisted state to diverge.

greg0x (commented):
See comments about the changes. 

I'm having some issues when testing the flow. Debugging root cause at the moment.
Will update when I know more.

greg0x (approved):
Done a quick smoke test and the PR works

Haven't done an exhaustive test

secant/Sources/Dependencies/VotingMetadataProvider/VotingMetadataStorage.swift
LukasKorba: URLResourceValues.isExcludedFromBackup = true set after every atomic write. Keeps the voting metadata file truly local-only.

secant/Sources/Features/Voting/VotingStore+Helpers.swift
LukasKorba: 4 helpers now throws. Internal rollback preserves the in-memory/disk invariant on failure. Caller catches via handlePersistFailure(_:state:) which logs the error + sets state.$toast = .top(...) so the user sees "Your voting changes couldn't be saved. Please try again."

LukasKorba: Given the nature or read/write operation on iOS and us never experiencing the issue, I consider this change good enough.

PR #1768, opened 2026-05-20
[MOB-1105] Tor routing for voting + HTTPS config + ServerHealthTracker hardening

LukasKorba:
## Summary
Routes all voting network traffic through Tor when the user has Tor enabled in Settings, hardens ServerHealthTracker against stale-entry crashes, and locks dynamic voting config fetches to HTTPS.

This is **PR 3 of 4** in the decomposition of the original [#1764](https://github.com/zodl-inc/zodl-ios/pull/1764).

**Depends on:** PRs #1766 + #1767. Diff includes all three until earlier ones merge.

## What's in this PR
- **Tor routing** — new performVotingRequest helper in VotingAPIClientLiveKey routes through sdkSynchronizer.httpRequestOverTor when swapAPIAccess == .protected, otherwise standard URLSession. Three call sites updated.
- **ServerHealthTracker Tor-aware probes** — actor now takes a caller-supplied @Sendable fetcher captured during initialize. Background probes respect the current Tor preference. No fallback to a plain URLSession that would leak the device IP every 60s.
- **HTTPS-only dynamic config URL** — StaticVotingConfig.validate now requires dynamicConfigURL.scheme == https. Static-config rotations can no longer silently downgrade.
- **VotingConfigSettings custom-chain validation** — also Tor-aware via the same dependency.
- **retryWithBackoff** — replace fatalError(unreachable) with a guarded rethrow.
- **ServerHealthTracker force-unwrap removal** — defensive against stale server map entries.

## What's NOT in this PR
- Navigation rewrite + UI polish (PR D)

## Test plan
- [ ] Manual: enable Tor, enter voting, observe traffic routes through Tor circuit
- [ ] Manual: disable Tor, enter voting, observe traffic routes through standard URLSession
- [ ] Manual: paste a custom poll URL with Tor on — validation request also routes through Tor
- [ ] Try a custom dynamic_config_url with scheme=http — must be rejected


secant/Sources/Dependencies/VotingAPIClient/VotingAPIClientLiveKey.swift
```diff
@@ -193,6 +190,25 @@ private let fastHttpSession: URLSession = {
...
+    _ request: URLRequest,
+    fast: Bool = false
+) async throws -> (Data, URLResponse) {
+    @Dependency(\.sdkSynchronizer) var sdkSynchronizer
+    @Shared(.inMemory(.swapAPIAccess)) var swapAPIAccess: WalletStorage.SwapAPIAccess = .direct
+
+    if swapAPIAccess == .protected {
+        let (data, response) = try await sdkSynchronizer.httpRequestOverTor(request)
+        return (data, response as URLResponse)
+    }
```
greg0x, line 222: Is it intentional for TOR to not use the fast policy?

With Tor off, `fast: true` still selects `fastHttpSession`, but with Tor on this branch always calls `httpRequestOverTor(request)`. That means health probes and share/status calls may no longer get the previous short timeout unless the Tor transport applies one itself.

That’s concise and framed as a question, which is appropriate if we have not confirmed the SDK Tor timeout behavior.

greg0x (approved):
tested and working ✨

secant/Sources/Dependencies/VotingAPIClient/VotingAPIClientLiveKey.swift
LukasKorba: Good catch, addressed [here](https://github.com/zodl-inc/zodl-ios/pull/1768/changes/3617a6f9b8fb3481e8efb38bb4ecd8fb10c370c4)

PR #1769, opened 2026-05-20
[MOB-1105] Voting navigation rewrite + UI polish + Settings entry

LukasKorba:
## Summary
Replaces the legacy `screenStack: [Screen]` state machine with a proper TCA `VotingCoordFlow` coordinator (`@Reducer enum Path` + `StackState` + SwiftUI `NavigationStack`). Adds a per-round session cache so navigation pops no longer flash the loading screen. Ships the Figma-aligned shimmer loading + "No polls" sheet.

This is **PR 4 of 4** in the decomposition of the original [#1764](https://github.com/zodl-inc/zodl-ios/pull/1764).

**Depends on:** PRs #1766 + #1767 + #1768. Diff includes all four until earlier ones merge.

## What's in this PR
- **VotingCoordFlow** — new coordinator (`Store`/`View`/`Coordinator` + `RoundSession` cache value type)
- **New screens** — `PollsList`, `ProposalList`, `ProposalDetail`, `ConfirmSubmission`, `Tallying`, `Results`, `Ineligible`, `HowToVote`, `NoRounds`, `WalletSyncing`, `VotingError`, `DelegationSigning` (skeleton)
- **Per-round cache** — witness pipeline, hotkey, voting weight, draft votes, tally results survive navigation pops; evicted on flow dismiss, wallet account switch, voting service config change
- **Legacy screens preserved** — pre-existing views moved to `LegacyXxxView.swift` so the legacy code keeps building until we delete it post-QA
- **Settings entry** — legacy "Coinholder Polling" hidden, new flow exposed as **"Beta: Coinholder Polling"** in all build configurations
- **UI polish** — shimmer skeleton loading screen, "No polls" blocking sheet over shimmer backdrop, focus-ring clipping fix on Poll Data Source card, double-back-arrow fix on Voting Settings, perception-tracking warning fixes
- **Submission pipeline** — placeholder alert; real submission lands in a follow-up
- **Endorsement filter re-enabled** in `PollsListView` (was temporarily disabled during development)
- **Reverted bundled config URL** back to the prod Valar source

## What's NOT in this PR
- Real submission pipeline (Phase 5 follow-up)
- Production entry swap (delete `Legacy*` files, route Settings to new flow as the primary entry) — blocked on Phase 5 + QA sign-off

## Test plan
- [ ] Settings shows only **"Beta: Coinholder Polling"** entry
- [ ] Tapping opens new flow as fullScreenCover, NavigationStack works
- [ ] Loading state shows shimmer card (not iOS spinner)
- [ ] Empty state shows "No polls right now" sheet over dimmed shimmer
- [ ] Vote / Review / Results navigation does NOT flash loading on back-pop
- [ ] Reviewer-flagged TODO in `ConfirmSubmissionView` + `VotingCoordFlowCoordinator` are intentional Phase 5 placeholders surfacing an alert


secant/Sources/Features/Voting/PollsList/PollsListView.swift
```diff
@@ -79,7 +91,7 @@ struct PollsListView: View {
         }
     }
 
-    private var visiblePolls: [Voting.State.RoundListItem] {
+    private var visiblePolls: [RoundListItem] {
```
greg0x, line 94: This re-enables the default-config endorsement filter, but `VotingCoordFlow` does not appear to fetch `fetchZodlEndorsedRoundIds()` after loading rounds. That means the bundled/default source can successfully load `allRounds` and still render an empty polls list because `zodlEndorsedRoundIds` stays empty. If this wiring is intentionally deferred to #1770, it would be worth making that explicit there; otherwise this PR should fetch endorsements before or alongside showing the polls list.

greg0x (approved):
Not blocking on these points, given this is a partial PR and we have the follow-up #1770 
But if these are not addressed in #1770 we should fix them here.

greg0x (approved):

secant/Sources/Features/Voting/PollsList/PollsListView.swift
greg0x: I wouldn’t block on this, but please preserve the endorsement fetch when #1770 is rebased. This PR now wires the default-source `fetchZodlEndorsedRoundIds()` happy path, while current #1770 is still based on the older coordinator and does not include that call. If the follow-up stack drops it, the default-source polls list can regress to filtering against an empty `zodlEndorsedRoundIds` set.

secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -0,0 +1,595 @@
...
+                // VotingConfigSettings emits `.delegate(.dismiss)` from its
+                // back button; pop the settings push so the user returns to
+                // the polls list. Re-fetch is handled by `.delegate(.saved)`
+                // separately (Phase 4+).
+                if !state.path.isEmpty {
+                    state.path.removeLast()
+                }
+                return .none
+
+            case .path(.element(id: _, action: .configSettings(.delegate(.saved)))):
```
greg0x, line 31: This reinitializes after a config-source save, but it keeps the previous voting state alive: `roundCache`, `allRounds`, `voteRecords`, `zodlEndorsedRoundIds`, `pendingPipelineRoundId`, and any in-flight pipeline. If the next fetch fails, or a round id collides across sources, the new UI can reuse stale per-round data from the previous source. Please clear the loaded voting state and cancel `cancelPipelineId` before sending `.initialize` (or narrow the cache-eviction claim if this is intentionally not wired here).

secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -0,0 +1,595 @@
...
+                    // wallet must land on IneligibleView instead of
+                    // marching through hotkey generation only to fail
+                    // submission later.
+                    let notes = try await votingCrypto.getWalletNotes(
+                        walletDbPath,
+                        snapshotHeight,
+                        networkId,
+                        accountUUID
+                    )
+                    let bundleResult = notes.smartBundles()
```
greg0x, line 445: This caches the raw note sum as voting power and then continues to hotkey generation. The legacy flow rejected empty notes and used `notes.smartBundles().eligibleWeight`, routing wallets with no eligible weight or weight below `ballotDivisor` to `IneligibleView`. Since this PR wires the active-round UI path, please preserve that eligibility behavior here before marking the round usable.

secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -0,0 +1,595 @@
...
+                    state.path.append(.proposalList(ProposalList.State(roundId: roundId)))
+                    // Cache check: skip pipeline if hotkey is already
+                    // populated for this round. Re-entry within the same
+                    // session is instant.
+                    if let cached = state.roundCache[roundId], cached.hotkeyAddress != nil {
+                        return .none
+                    }
+                    return .send(.startActiveRoundPipeline(roundId: roundId))
+                case .tallying:
+                    state.path.append(.tallying(Tallying.State(roundId: roundId)))
```
greg0x, line 285: A tallying round is pushed to `TallyingView`, but the new coordinator does not start any status polling or refresh action from here. The UI text says results appear once the helpers finish tallying, but this path returns `.none`, so users can sit on the tallying screen forever until they manually leave and reload after finalization. Please start/reuse round-status polling for tallying rounds or provide a manual refresh path.

secant/Sources/Features/Voting/ProposalDetail/ProposalDetailView.swift
```diff
@@ -0,0 +1,137 @@
...
+            .screenTitle(String(localizable: .coinVoteCommonScreenTitle))
+            .zashiBack()
+        }
+    }
+
+    @ViewBuilder
+    private func optionRow(_ option: VoteOption, selected: VoteChoice?) -> some View {
+        let isSelected = selected == .option(option.index)
+        let isReviewing = mode == .review
+        Button {
```
greg0x, line 85: When a submitted round is opened through `reviewVotes`, the list uses `.review` mode but proposal detail is still fully editable: tapping an option sends `.draftVoteSet`. That can create a new draft for a completed round and invalidate the completed-vote marker on the next load because completed records are only valid when drafts are empty. Please carry read-only mode into proposal detail, or suppress option actions when opened from review mode.

secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -0,0 +1,595 @@
...
+                }
+
+            case let .tallyResultsLoaded(roundId, results):
+                state.roundCache[roundId, default: RoundSession(roundId: roundId)]
+                    .tallyResults = results
+                state.roundCache[roundId, default: RoundSession(roundId: roundId)]
+                    .tallyFetched = true
+                return .none
+
+            case let .tallyResultsFailed(roundId, message):
```
greg0x, line 360: This leaves the results screen in a permanent loading state on tally fetch failure. `ResultsView` only stops showing “Loading results…” when `tallyFetched == true`, but this failure path neither marks the fetch complete nor stores an error/retry state. Please either render an error/retry path or track failed completion separately from successful results.

PR #1773, opened 2026-05-21
Restore VotingCoordFlow feature parity after Phase 5D

greg0x:
## Why

Phase 5D deleted the legacy `Voting` reducer and swapped the production entry to `VotingCoordFlow`. The coordinator-backed flow shipped without several behaviors the legacy reducer carried — most were no-ops or one-shot loads instead of lifecycle-scoped loops. This PR restores those gaps so the new flow is at feature parity before the broader MOB-1105 integration lands.

## Changes

Each commit is one discrete restoration:

- **Restore voting flow polling lifecycle** — proving-cache warm-up, opened-round status poll, poll-closed alert handling, post-finalization new-round poll. Cancel IDs wired into config changes and flow dismissal so loops do not outlive the flow.
- **Restore Keystone signing recovery on round entry** — loads persisted partial Keystone bundle signatures after the active-round pipeline rebuilds bundle state, validates indices, resumes the next signing request or the all-signed delegation proof path. Recovers from crash/restart mid-signing.
- **Restore Zashi delegation precompute** — starts background PIR precompute once hotkey and voting weight are ready, caches delegation inputs per bundle, resumes pending submission through the existing auth path. Avoids paying the PIR setup cost inline at submission time.
- **Restore share confirmation recovery** — per-round share tracking state, DB load, confirmation polling, and overdue resubmit through the existing helper APIs. Started after submitted votes load or a batch completes, cancelled with the flow lifecycle.
- **Reset voting state when wallets change** — votes/hotkeys/witnesses/recovery are wallet-scoped but the flow cached round data by round id only. Now listens for selected-wallet changes from the view, clears account-scoped state, cancels in-flight work, and reinitializes for the new wallet id when the intro has already been seen.
- **Persist Keystone skipped bundle metadata** — completed vote records now carry the original eligible Keystone weight and bundle count alongside submitted weight, so review surfaces can show submitted-vs-total. Legacy/Zashi records decode through optional fields.

## Scope

Concentrated in `VotingCoordFlowCoordinator.swift` (+944), with small touches to `RoundSession.swift`, `VotingCoordFlowStore.swift`, `VotingCoordFlowView.swift`, `VotingHelpers.swift`, and `VotingMetadata.swift`. No public API changes outside the voting flow.

## Test plan

- [ ] Enter an active round → confirm proving cache warms and opened-round poll runs; close round externally → poll-closed alert fires.
- [ ] Keystone: sign part of a bundle set, force-kill app, re-enter round → signing resumes from the right bundle; skip a bundle → completed record reflects submitted-vs-total weight.
- [ ] Zashi: submit a vote → precompute is already cached (no inline PIR setup wait).
- [ ] Submit shares → confirm DB row persists; force overdue → resubmit fires.
- [ ] Switch selected wallet mid-flow → state clears and reinitializes for the new account.

### DIFF
--- a/secant/Sources/Features/CoordFlows/VotingCoordFlow/RoundSession.swift
+++ b/secant/Sources/Features/CoordFlows/VotingCoordFlow/RoundSession.swift
@@ -139,6 +139,12 @@ struct RoundSession: Equatable {
     /// proposal count) into the encrypted voting metadata file. The Results
     /// screen uses this to render "Voted MMM d - Voting Power X.XXX ZEC".
     var voteRecord: Voting.VoteRecord?
+
+    /// DB-backed helper-server share confirmation tracking. The UI
+    /// workstream owns presentation; the coordinator keeps this state
+    /// current so My Votes/review surfaces can read it.
+    var shareTrackingStatus: ShareTrackingStatus = .idle
+    var shareDelegations: [VotingShareDelegation] = []
 }
 
 // MARK: - Submission state machine types
@@ -227,4 +233,11 @@ struct KeystoneBundleSignature: Equatable {
     let sig: Data
     let sighash: Data
     let rk: Data // swiftlint:disable:this identifier_name
+}
+
+enum ShareTrackingStatus: Equatable {
+    case idle
+    case loading
+    case tracking
+    case fullyConfirmed
 }
--- a/secant/Sources/Features/CoordFlows/VotingCoordFlow/RoundSession.swift
+++ b/secant/Sources/Features/CoordFlows/VotingCoordFlow/RoundSession.swift
@@ -26,6 +26,12 @@ struct RoundSession: Equatable {
     /// (wallet, round) pair until a new snapshot or wallet rescan.
     var votingWeight: UInt64 = 0
 
+    /// Original eligible power before any Keystone bundle skipping. For
+    /// Zashi and full Keystone submissions this matches `votingWeight`; when
+    /// Keystone users skip unsigned bundles, `votingWeight` is reduced and
+    /// this remains the pre-skip value for persisted transparency metadata.
+    var eligibleVotingWeight: UInt64 = 0
+
     /// Eligible notes at the round's snapshot height. Cached because the
     /// SDK query is non-trivial and the snapshot is immutable.
     var walletNotes: [NoteInfo] = []
@@ -38,6 +44,10 @@ struct RoundSession: Equatable {
     /// bundling step in the active-round pipeline. Drives both the
     /// delegation proof loop and the per-bundle vote submission loop.
     var bundleCount: UInt32 = 0
+
+    /// Original eligible bundle count before any Keystone skip. Kept so a
+    /// completed vote record can explain reduced power later.
+    var eligibleBundleCount: UInt32 = 0
 
     /// Per-round hotkey address derived deterministically from the per-
     /// account hotkey mnemonic (in the Keychain) and the round id. Same
--- a/secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
+++ b/secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
@@ -51,6 +51,7 @@ extension VotingCoordFlow {
                 state.pollClosedRoundId = nil
                 return .merge(
                     .cancel(id: cancelPipelineId),
+                    .cancel(id: cancelDelegationPrecomputeId),
                     .cancel(id: cancelStatusPollingId),
                     .cancel(id: cancelNewRoundPollingId),
                     .send(.initialize)
@@ -264,6 +265,7 @@ extension VotingCoordFlow {
                     .cancel(id: cancelPipelineId),
                     .cancel(id: cancelSubmissionId),
                     .cancel(id: cancelDelegationProofId),
+                    .cancel(id: cancelDelegationPrecomputeId),
                     .cancel(id: cancelStatusPollingId),
                     .cancel(id: cancelNewRoundPollingId)
                 )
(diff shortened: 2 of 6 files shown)


LukasKorba (commented):
## utACK at 32fb3f34

Read all 6 commits one at a time. Built clean on `iPhone 17 Pro` simulator. Cross-checked against the durability invariants from #1770 — no violations. Cancel-ID discipline is exemplary: every new `.run` loop is `.cancellable(id:cancelInFlight:)` and every cancel ID is wired into `.dismissFlow`, the config-change reset, AND the wallet-change reset.

Holding off on Approve until I've worked through your test plan on a device (and since this is still draft anyway).

### Actionable nits (non-blocking)

1. **Hardcoded English in `pollClosed(status:)` alert** (`b723b738`) — `"Voting closed"`, `"View results"` / `"View status"`, `"Back to polls"`, plus the three message variants. Catalog entries can land via the UI polish branch before this hits a release, but flagging it here so it doesn't disappear into the diff.
2. **`smartBundles()` is called twice in `.votingWeightLoaded`** (`32fb3f34`) — once in `eligibleTotals(for:)`, once in `votingWeight(for:bundleCount:)`. Cheap, but a single computation + reuse would be one less re-bundling per round entry.
3. **`account?.vendor.isHWWallet() ?? false` → `isKeystoneUser`** (`31a58d94`) — implicit assumption that all HW vendors are Keystone. Consistent with usage elsewhere in the codebase; not a regression. Flagging so it surfaces the next time a non-Keystone HW vendor lands.

### Per-commit notes

**`b723b738` Restore voting flow polling lifecycle**
- Proving cache warm-up: idempotent via `hasRequestedProvingCacheWarmup`, non-fatal failure handling. ✓
- Status polling (5 s) on opened active rounds and finalization polling (30 s) for new rounds — both cancelled on config change and dismiss. ✓
- Status-transition handling cleanly distinguishes `activeVotingFlowRoundId` (in voting/review/submission → show alert) from `topPathRoundId` (in tallying/results/ineligible → swap path entry). ✓

**`fd49cc54` Restore Keystone signing recovery on round entry**
- `validKeystoneSignatures` strict-sequential validation refuses gaps/out-of-order arrays. ✓
- Pre-clear capture before `prepareFreshRound` then re-store via `storeKeystoneBundleSignature` — preserves recovery state across the fresh prep that `clearRound`s the DB. ✓
- `.keystoneSignaturesRestored` resets `pendingVotingPczt` / `pendingUnsignedDelegationPczt` / `keystoneSigningStatus` — aligns with #1770 invariant 5 (Keystone state resets on terminal events). ✓
- All-signed branch correctly transitions to `delegationProofStatus = .generating(progress: 0)` + `batchSubmissionStatus = .authorizing` + dispatches `.keystoneAllBundlesSigned`. ✓

**`4feaf7d9` Restore Zashi delegation precompute**
- `reduceMaybeStartDelegationPrecompute` guards: Keystone check, `notStarted` status, in-flight flags, `hotkeyAddress != nil`, `bundleCount > 0`, `walletNotes` non-empty, active status, service-config + seed-fingerprint + accountId present. Solid idempotency. ✓
- Per-bundle skip via `getDelegationTxHash` — won't recompute material already on-chain. ✓
- Failure path also resumes pending submission so a precompute failure doesn't strand the user (they pay inline PIR cost instead). Nice safety net. ✓
- `Task.checkCancellation()` between bundles for clean cancellation. ✓

**`040b472d` Restore share confirmation recovery**
- DB is source of truth (`votingCrypto.getShareDelegations`); in-memory `RoundSession.shareDelegations` is a cache. ✓
- Both trigger points (post `.batchVoteSubmitted` and post-`loadSubmittedVotesFromDb` with a vote record) gate on `shareTrackingStatus == .idle`. ✓
- `shouldResubmitShare` threshold scales with remaining window (`max(30, min(3600, remainingWindow / 4))`) and skips if `voteEndTime <= now + 10`. ✓
- `cancelShareTrackingIfSwitchingRound` cancels when navigating between rounds. ✓
- The `.roundStatusUpdated` refactor adding the `isCurrentRound` gate tightens scope vs. `b723b738` (no more cross-round cancellation). ✓

**`31a58d94` Reset voting state when wallets change**
- `.onChange(of: store.selectedWalletAccount?.id)` doesn't double-fire on initial mount. ✓
- Idempotent reducer guard: returns `.none` if walletId AND isKeystoneUser unchanged. ✓
- `resetAccountScopedVotingState` is comprehensive and correctly leaves `hasRequestedProvingCacheWarmup` (process-lifetime) + intro flags + override URL (cross-account) untouched. ✓
- All 7 cancel IDs cancelled in one `.merge`. ✓
- Three-way branch (nil account → loading; intro unseen → howToVote; otherwise → loading + reinitialize) is correct. ✓

**`32fb3f34` Persist Keystone skipped bundle metadata**
- `eligibleVotingWeight` / `eligibleBundleCount` snapshotted in `.votingWeightLoaded` from `notes.smartBundles()` — pre-skip, stable across recovery (because notes are the immutable snapshot view). ✓
- `reduceSkipRemainingKeystoneBundles` saves the pre-mutation values via outer-scope `session.bundleCount` / `session.votingWeight` BEFORE overwriting with reduced `signedCount` / `signedWeight`. ✓
- Idempotent guard `if roundSession.eligibleBundleCount == 0` prevents accidental overwrites on repeated invocation. ✓
- `PersistedVotingRecord` + `VoteRecord` use optional new fields; older persisted records decode cleanly. ✓
- Zashi nil-out at the persistence site: `state.isKeystoneUser ? … : nil`. ✓
- `skippedKeystoneBundleCount` / `hasSkippedKeystoneBundles` give review surfaces a clean API.

### Manual test plan (TODO on a real device)
- [ ] Force-kill mid Keystone signing → re-enter → verify resume index
- [ ] Open active round → close externally → poll-closed alert + status replacement
- [ ] Submit Zashi vote → confirm precompute already cached on the second submission
- [ ] Force a share overdue → confirm `resubmitShare` fires
- [ ] Switch wallet mid-flow → confirm state clears + reinitializes
- [ ] Skip Keystone bundles → submit → verify completed record carries submitted-vs-total

Beautifully scoped restoration — one focused commit per concern, no accidental scope creep. Will switch to Approve once tests pass and you mark this Ready.

PR #1784, opened 2026-05-26
[#MOB-1105] Voting UI fixes, polishments, and endorsement-set race fix

LukasKorba:
## Summary

Post-merge UI/copy polish and a couple of correctness fixes around the Voting flow.

- **Polls list sorting** — clusters by status (Active → Voted → Closed); within each cluster sorts by end time (soonest first for Active/Voted, most-recently-closed first for Closed).
- **Confirm Submission header icons** — disc pair now uses `destinationOut` + `compositingGroup` so the right disc bites a notch out of the left one (matches Figma).
- **Proposal Detail** — vote-option radio restyled (filled donut + stroked offBg, mirrors CurrencyConversion); bottom bar uses `safeAreaInset` so the forum-discussion row no longer overlaps scrolling content.
- **Submission Done CTA** — pops back to PollsList instead of pushing ReviewVotes; user decides where to go next.
- **Poll Closed dialog** — system `AlertState` replaced with a `votingSheet` matching Figma (red exclamation, *View results* / *Close*).
- **Voting Server Settings screen** — UI fixes and polishments (radio + chevron tap zones, Add Source popover keyboard handling, title field autofocus, dropped three-dots menu, etc).
- **Endorsement parser** — accepts both base64 and hex round-id encodings (stage returns hex; old base64-only path silently produced an empty set and hid endorsed rounds behind the filter).
- **PollsList vs NoRounds race** — coordinator now defers the root-screen decision on default config until endorsements resolve. Previously the empty-state was computed against a not-yet-loaded endorsement set, leaving users stuck on NoRounds.
- **Voting localization cleanup** — `VotingConfigSettings(View)` literal strings moved under `coinVote.configSettings.*` SwiftGen keys for consistency with the rest of voting.

## Test plan

- [ ] Polls list shows Active → Voted → Closed clusters with the expected within-cluster ordering.
- [ ] Confirm Submission screen renders the disc-pair with the notch effect.
- [ ] Proposal Detail radio matches the Figma; forum-discussion row clears content when scrolling.
- [ ] After a successful submit, Done returns to PollsList (the round is now marked Voted).
- [ ] Tapping a finalized/tallying round surfaces the Poll Closed sheet (icon, copy, buttons).
- [ ] In Voting Server Settings: radio area selects; rest of a custom row opens Edit popover; default row tap selects from anywhere; Add Source popover keeps the toolbar visible when keyboard opens; title field auto-focuses for both Add and Edit.
- [ ] On bundled (default) config with endorsed stage rounds: list populates without flicker through NoRounds.


secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -194,7 +193,18 @@ extension VotingCoordFlow {
...
-                state.rootScreen = visibleRoundCount(state: state) == 0 ? .noRounds : .pollsList
+                // On default config, defer the pollsList-vs-noRounds decision
+                // until `.zodlEndorsementsLoaded` (or `.zodlEndorsementsFailed`)
+                // resolves; otherwise this branch would compute the empty-state
+                // against an as-yet-unpopulated endorsement set and lock the
+                // user on noRounds even when endorsed rounds arrive a moment
+                // later. On custom config there's no endorsement filter, so we
+                // can decide immediately from `allRounds`.
+                if state.isOnDefaultConfig {
+                    state.rootScreen = .loading
```
greg0x: Small follow-up: this also runs on later `.allRoundsLoaded` refreshes, not just the initial load. Once endorsements have resolved and the user is already on the polls list, we should preserve `.pollsList` while refetching endorsements and reserve `.loading` for the initial unknown-endorsement case. Otherwise a slow or failed endorsement request can briefly blank/flicker the list.

greg0x (approved):
utACK 

Left a nit.

secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
LukasKorba: Addressed in a3dd3dd6 — only fall back to `.loading` when no round-derived screen is showing. Refreshes while on `.pollsList` or `.noRounds` preserve the visible state until `.zodlEndorsementsLoaded` re-evaluates.

PR #1791, opened 2026-05-27
[#MOB-1105] Draft Keystone bundle UI handoff

greg0x:
## Summary

- Matches the Keystone voting confirmation details to the Figma design.
- Reworks the Keystone bundle signing screen around the Figma layout.
- Keeps the existing per-bundle memo implementation intact, so the memo shown in
  the app should continue to match what Keystone displays for each bundle.

## Handoff note

I am sharing this draft with @LukasKorba because Lukas is taking over the
implementation. This gives him a concrete snapshot to decide whether to build on
top of this work or restart from scratch.

This is not a release blocker. It can be treated as a quick follow-up if the
current release path is already locked.

## Verification

- Built zashi-internal for the iOS simulator with code signing disabled.
- Ran focused voting tests:
  - VotingCoordFlowCoordinatorTests
  - VotingHelpersTests
- Result: 23 tests passed, 0 failures.

### DIFF
--- a/secant/Sources/Features/Voting/DelegationSigning/DelegationSigningStore.swift
+++ b/secant/Sources/Features/Voting/DelegationSigning/DelegationSigningStore.swift
@@ -35,6 +35,8 @@ struct DelegationSigningView: View {
     @Perception.Bindable var store: StoreOf<VotingCoordFlow>
     let roundId: String
 
+    @State private var isQRCodeEnlarged = false
+
     var body: some View {
         WithPerceptionTracking {
             let session = store.roundCache[roundId]
@@ -93,6 +95,19 @@ struct DelegationSigningView: View {
                 store: store.scope(state: \.$keystoneScan, action: \.keystoneScan)
             ) { scanStore in
                 ScanView(store: scanStore, popoverRatio: 1.075)
+            }
+            .enlargeQR(isPresented: $isQRCodeEnlarged) {
+                Group {
+                    if let pczt = store.roundCache[roundId]?.pendingUnsignedDelegationPczt,
+                       let encoder = sdkSynchronizer.urEncoderForPCZT(pczt) {
+                        AnimatedQRCode(urEncoder: encoder, size: UIScreen.main.bounds.width - 64)
+                            .padding()
+                            .background {
+                                RoundedRectangle(cornerRadius: Design.Radius._4xl)
+                                    .fill(Color.white)
+                            }
+                    }
+                }
             }
         }
     }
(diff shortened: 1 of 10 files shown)


secant/Sources/Features/Voting/DelegationSigning/DelegationSigningStore.swift
```diff
@@ -140,165 +152,219 @@ struct DelegationSigningView: View {
...
+        VStack(spacing: 0) {
+            signatureProgressSection(
+                current: current,
+                total: total,
+                signed: signed,
+                signedWeight: signedWeight,
+                pendingWeight: pendingWeight
+            )
+
+            if canUseSignedBundlesOnly(signed: signed, total: total) {
```
czarcas7ic: Can we only show this row while `status == .awaitingSignature`? Right now it can be tapped while the next Keystone PCZT is still being prepared. The old CTA only appeared while awaiting a signature, so matching that should avoid the race.

I doubt the window in practice would allow for this but might as well be safe like the old logic.

secant/Sources/Features/Voting/DelegationSigning/DelegationSigningStore.swift
```diff
@@ -321,6 +387,77 @@ struct DelegationSigningView: View {
         return votingAuthorizationMemo(pollTitle: pollTitle, rawWeight: bundleTotal)
     }
 
+    private static func bundleWeightSummary(session: RoundSession?) -> (signed: UInt64, pending: UInt64) {
```
czarcas7ic, line 396: Can we base this on `resolvedKeystonePrefixCount` instead of `keystoneBundleSignatures.count`?

A recovered bundle can already be completed without still being counted in `keystoneBundleSignatures`:
https://github.com/zodl-inc/zodl-ios/blob/2a4cb7de7b22dfb0363860109fe9099b10e4dcf2/secant/Sources/Features/CoordFlows/VotingCoordFlow/RoundSession.swift#L261-L277

The skip path also uses that resolved prefix:
https://github.com/zodl-inc/zodl-ios/blob/2a4cb7de7b22dfb0363860109fe9099b10e4dcf2/secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift#L2926

So in a crash/recovery case, this UI can show `0 signed` even though the reducer has a usable recovered prefix.

czarcas7ic (approved):

secant/Sources/Features/Voting/DelegationSigning/DelegationSigningStore.swift
LukasKorba: Addressed in c01e11a7 — `canUseSignedBundlesOnly` now also requires `status == .awaitingSignature`. Threaded `status` through `signingDetailsCard` so the gate matches the old standalone CTA.

secant/Sources/Features/Voting/DelegationSigning/DelegationSigningStore.swift
LukasKorba: Addressed in c01e11a7 — switched both the displayed `signedBundleCount` (line 46) and `bundleWeightSummary`'s `signedCount` cutoff to use `resolvedKeystonePrefixCount`. Recovered bundles now fall into the signed bucket immediately even if `keystoneBundleSignatures` hasn't been repopulated yet.

PR #1794, opened 2026-05-28
Cover app content on .inactive so App Switcher snapshot is hidden

Sajjon:
> [!CAUTION]
> This PR has been implemented by Claude (Opus 4.7)
> and should merely be regarded as a “detailed GH issue” - although I’ve 
> verified it works in simulator.

## Summary

The lock-screen / privacy overlay was not actually hiding the app's content in the iOS App Switcher.

> [!NOTE]
> You might not actually *want* the privacy overlay shown eagerly like this,
> but most wallets and banking apps typically hides content eagerly like this.

**Root cause:** `RootView` was only covering content when `scenePhase == .background`, but iOS captures the App Switcher snapshot during the `.inactive` transition — which fires *before* `.background`. By the time `.background` arrives, the snapshot is already taken.

**Fix:** trigger the cover whenever the scene is non-active (`scenePhase != .active`). This catches `.inactive` (the snapshot moment) as well as `.background`.

Also includes a small project-file cleanup needed to build the testnet target locally: removes a stale `IPHONEOS_DEPLOYMENT_TARGET = 15.0` override (which conflicted with ComposableArchitecture's iOS 16 minimum) and drops dead `PBXBuildFile` entries for test files that no longer exist.

## Test plan

- [ ] Build & run `secant-testnet` on simulator
- [ ] Open the app to any screen (e.g. home with balances visible)
- [ ] Swipe up to the App Switcher
- [ ] Verify the Zashi-icon cover is visible in the app's App Switcher card — no real content visible
- [ ] Tap back into the app and confirm normal UI returns
- [ ] Repeat with a physical device if available

### DIFF
--- a/secant.xcodeproj/project.pbxproj
+++ b/secant.xcodeproj/project.pbxproj
@@ -2330,10 +2330,6 @@
 		37D0F0000000000000000022 /* RoundAuthenticator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37D0F0000000000000000002 /* RoundAuthenticator.swift */; };
 		37D0F0000000000000000023 /* RoundAuthenticator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37D0F0000000000000000002 /* RoundAuthenticator.swift */; };
 		37D0F0000000000000000024 /* RoundAuthenticator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37D0F0000000000000000002 /* RoundAuthenticator.swift */; };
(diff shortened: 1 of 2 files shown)


secant.xcodeproj/project.pbxproj
LukasKorba, line 1: I beleive this is here by accident, could @Sajjon revert it, please?

Sajjon: Replied here https://github.com/zodl-inc/zodl-ios/pull/1795#discussion_r3336656852

PR #1668, opened 2026-03-30
Update TCA to 1.25.4 for Xcode 26 compatibility

greg0x:
## Summary

- Bump TCA from 1.23.1 to 1.25.4 — fixes `WritableKeyPath` Sendable errors on Xcode 26 (pointfreeco/swift-composable-architecture#3844)
- Add explicit `xctest-dynamic-overlay >= 1.9.0` floor — SPM's resolver otherwise picks 1.4.3 which is missing the `withIssueContext` isolation parameter TCA 1.25.4 needs
- Bump `IPHONEOS_DEPLOYMENT_TARGET` from 15.0 to 16.6 across all targets to match `secant-distrib` — TCA 1.25.4 requires iOS 16 minimum

Closes zodl-inc/zodl-ios#1667

## Test plan

- [x] Clean build on Xcode 26.4 (secant-mainnet)
- [x] App tested and working
- [x] Verify no runtime regressions on iOS 16.6+ device/simulator

PR #1795, opened 2026-05-28
Restore edge-swipe-to-back gesture on .zashiBack() screens

Sajjon:
> [!CAUTION]
> This PR has been implemented by Claude (Opus 4.7)
> and should merely be regarded as a “detailed GH issue” - although 
> I’ve verified that it works in simulator (unable to run on device due to the iCloud container.)

## Summary

Edge-swipe-to-back was broken on screens that use the custom \`.zashiBack()\` / \`.zashiBackV2()\` toolbar back button — most visibly **Receive → QR Code / AddressDetails**.

**Root cause:** both modifiers apply \`.navigationBarBackButtonHidden(true)\` so the custom button can replace the native one. SwiftUI/UIKit also disables \`UINavigationController.interactivePopGestureRecognizer\` whenever the back button is hidden, which kills the edge swipe.

**Fix:** small \`UIViewRepresentable\` helper (\`enableSwipeBackGesture()\`) that walks the responder chain to the host \`UINavigationController\` and re-enables the gesture with a delegate that only allows it when the stack has more than one view controller — so the root view of a \`NavigationStack\` doesn't lock up if the user tries to swipe there. Applied from both \`.zashiBack()\` and \`.zashiBackV2()\` in the visible-back branch (the \`hidden\` case stays gestureless on purpose).

Also includes the same project-file cleanup as #1794 (drops \`IPHONEOS_DEPLOYMENT_TARGET = 15.0\` so the testnet target builds against ComposableArchitecture's iOS 16 minimum, plus stale test \`PBXBuildFile\` entries removed by Xcode).

## Test plan

- [ ] Build & run \`secant-testnet\`
- [ ] From Home → Receive → tap a QR icon to reach AddressDetails
- [ ] Edge-swipe from the left → pops back to Receive
- [ ] Try edge-swiping on the **root** of the Receive flow → no lockup, no unintended pop
- [ ] Spot-check other \`.zashiBack()\` screens reachable via NavigationStack push (SendForm sub-screens, Transactions detail, etc.)
- [ ] Confirm screens using \`.zashiBack(hidden: true)\` still don't swipe back

### DIFF
--- a/secant.xcodeproj/project.pbxproj
+++ b/secant.xcodeproj/project.pbxproj
@@ -2330,10 +2330,6 @@
 		37D0F0000000000000000022 /* RoundAuthenticator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37D0F0000000000000000002 /* RoundAuthenticator.swift */; };
 		37D0F0000000000000000023 /* RoundAuthenticator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37D0F0000000000000000002 /* RoundAuthenticator.swift */; };
 		37D0F0000000000000000024 /* RoundAuthenticator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37D0F0000000000000000002 /* RoundAuthenticator.swift */; };
-		9E0C0D702AFB842B00D69A16 /* TransactionStateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E0C0D6F2AFB842B00D69A16 /* TransactionStateTests.swift */; };
-		9E139AAF2B07B3A600D104B8 /* ZatoshiStringRepresentationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E139AAE2B07B3A600D104B8 /* ZatoshiStringRepresentationTests.swift */; };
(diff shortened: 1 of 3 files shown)


secant/Sources/UIComponents/Toolbar/ZashiBack.swift
```diff
@@ -71,4 +73,62 @@ extension View {
...
+    }
+
+    required init?(coder: NSCoder) { fatalError("init(coder:) has not been implemented") }
+
+    override func didMoveToWindow() {
+        super.didMoveToWindow()
+        guard let nav = findNavigationController() else { return }
+        coordinator.navigationController = nav
+        nav.interactivePopGestureRecognizer?.delegate = coordinator
+        nav.interactivePopGestureRecognizer?.isEnabled = true
```
Sajjon, line 123: This enables the swipe to go back gesture

secant.xcodeproj/project.pbxproj
LukasKorba, line 1: I beleive this is here by accident, could @Sajjon revert it, please?

Sajjon: @LukasKorba Hey! When I branched of from `main` the project did not compile due to deployment target 15 clashing with TCA IIRC. I saw https://github.com/zodl-inc/zodl-ios/pull/1668#discussion_r3026809083

What I’ve done is I removed the **overrides** for each target, and let ONLY the project set 16.0 (which then applies for ALL targets), typically what one wants.

Claude then detected dead PBXBuildFile entries for test files that no longer exist

PR #1797, opened 2026-05-30
Add testTag IDs for maestro e2e

zcash-harry:
First entry in the maestro tag-IDs initiative. Adds .accessibilityIdentifier("home.moreButton") to the dots menu Button in GlobalNavBar.swift so the maestro smoke suite can find it deterministically (was relying on coord taps).


LukasKorba (approved):
utACK

PR #1799, opened 2026-06-01
[#MOB-1318] Voting: fix reported UI issues in the polling flow

LukasKorba:
## Summary

Bundled fixes for six UI/UX issues reported against the Coinholder Polling flow after the 3.5.0 release.

- **Review screen header** — shows "Voted <date>" once a ballot is submitted, using the locally persisted `votedAt` from the encrypted vote metadata, instead of the round's upcoming end date.
- **Review screen subtitle removed** — dropped the unused "Review your submitted votes" line to match the design.
- **Unverified poll warning sheet** — tapping Enter Poll on a poll from a custom (non-default) data source now surfaces a Figma-matched warning sheet with Go back (primary) / Proceed anyway (secondary). Brings iOS to parity with Android.
- **Poll Closed sheet gated by `hasVoted`** — the post-deadline "your vote can no longer be submitted" sheet is now suppressed for users who already submitted their ballot. Telling someone they can't vote right after voting is just noise.
- **Tally-bar colors on the answers screen** — green for Yes/Support, red for No/Oppose, gray for Abstain, blue for everything else. Matches the product/design guidance and the rest of the voting screens.
- **Specific error when the same wallet already voted from another device** — when the backend returns a nullifier-already-spent error during authorization/submission, the sheet now shows the dedicated "wallet already registered for this voting round" copy instead of the generic "Check your connection and try again." Mirrors Android: detection now uses case-insensitive `contains("nullifier") && contains("spent")`; the two Keystone-flow `delegationProofFailed` sites now pre-map through `VotingErrorMapper` so `BatchSubmissionStatus.authorizationFailed(error:)` consistently carries a user-friendly string; the view shows that string directly and falls back to the generic copy only when empty.

## Test plan

- [ ] Vote on a poll, return to the review screen — header reads "Voted <date>", no "Review your submitted votes" subtitle present.
- [ ] Configure a custom (non-default) voting data source, tap Enter Poll on any poll — warning sheet appears with Go back / Proceed anyway; Go back returns to the list; Proceed anyway enters the poll.
- [ ] Let a round transition to `.tallying`/`.finalized` while not having voted — Poll Closed sheet appears as before. After voting, transition the round — Poll Closed sheet is suppressed.
- [ ] Open the answers screen on a poll with Yes/No/Abstain options — Yes is green, No is red, Abstain is gray, custom labels render blue.
- [ ] On a Zashi wallet, vote on round X from device A, then attempt to vote on the same wallet on device B — failure sheet shows "This wallet has already been registered for this voting round..." copy (EN + ES).
- [ ] Repeat the same scenario with a Keystone wallet (covers the `:2927` Keystone delegation-proof path) — same specific copy.
- [ ] Other voting-failure paths (e.g. unreachable lightwalletd) still show their existing user-friendly messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ directly impact users rather than highlighting other crucial architectural updat
 ### Changed
 - Default server set to zec.rocks.
 
+### Added
+- Tapping Enter Poll on a poll from a custom (unverified) data source now surfaces an "Unverified Poll" warning sheet, letting you go back or proceed at your own risk.
+
+### Fixed
+- We fixed the Coinholder Polling review screen so it now shows "Voted <date>" once you've submitted your ballot, instead of the round's upcoming end date.
+- We removed an unused "Review your submitted votes" subtitle from the Coinholder Polling review screen header to match the design.
+- We suppressed the Coinholder Polling "Poll Closed" sheet for users who already submitted their ballot — telling someone they can no longer vote right after voting is just noise.
+- We corrected the tally-bar colors on the Coinholder Polling answers screen: green for Yes/Support, red for No/Oppose, gray for Abstain, blue for everything else — matching the rest of the voting screens.
+- We now show a specific message when an attempt to vote fails because the same wallet has already voted from another device, instead of the generic "Check your connection" copy.
+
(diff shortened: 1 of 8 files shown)


Chlup (approved):
utACK

PR #1804, opened 2026-06-04
[MOB-1098] Automatic server selection: Automatic/Manual connection mode

Chlup:
## Summary

Reimplements automatic server selection (MOB-1098), superseding #1750 with a simpler, single-source-of-truth design.

- **Automatic mode** benchmarks the known lightwalletd endpoints at launch and on returning to foreground and keeps the wallet on the fastest one. **Manual mode** pins a chosen server (a known one or a custom `host:port`). The only stored state is the existing `ups_server` config plus one new `automaticServerSelection` boolean flag — no dual storage and no `EndpointSwitchCoordinator`.
- **Safety (the core fix):** a small `TransactionGuard` actor makes server switches and transaction submissions mutually exclusive. Every submission — send, swap, shield, Flexa, Keystone PCZT, and the voting submit/delegation/tree-state path — is wrapped, so an automatic switch can never run while a transaction is being created or broadcast (the SDK's `switchTo(endpoint:)` is unsafe mid-submission). The auto-refresh skips while a submission is active or while Server Setup is open; a manual Save waits for any in-flight switch and wins.
- **Migration:** on first run the mode is derived from the existing server — default/none → Automatic; custom or non-default → Manual (preserves an explicit/private choice).
- **Server Setup** gets the two-mode UI (store simplified to the flag + single source of truth; the PR's two-mode view reused).
- The Root store keeps only two one-line triggers (launch / foreground); all orchestration lives in `AutoServerSelectionClient`.

## Test plan

- [x] 119 unit tests pass on the `zashi-internal` scheme — `TransactionGuard` (mutex, skip-while-submitting, manual-waits, cancelled-parked-op), `AutoServerSelectionClient`, first-run migration, `ServerSetup` store (Automatic/Manual save, no-op), network-aware endpoints. (Last full run was green; the trailing whitespace-only "remove semicolons" change was not rebuilt because the Xcode MCP was offline.)
- [ ] `secant-mainnet` scheme builds (couldn't switch the active scheme from the MCP; all new files are members of all 5 app targets and there's no testnet/mainnet-conditional code).
- [ ] Manual: Settings → Server — Automatic/Manual toggle; Save disabled when Custom is selected but blank; Auto→Manual preselects the active server; switch applies and persists.
- [ ] Manual: confirm no server switch occurs during an in-flight send / swap / shield / vote.

> Note: this branch also includes a non-feature `Project update` commit (`project.pbxproj` + scheme) added during a rebase outside the feature work.


LukasKorba (approved):
utACK

PR #1805, opened 2026-06-05
[MOB-1119] Update text on Smart banner related to unshielded funds

Chlup:
- Updated to latest alpha SDK.
- Removed Text from smart banner that used `.smartBannerContentShieldTitleShorter` because this one was using old text. And  `.smartBannerContentShieldTitle` text was already using new text. So I removed old text from strings catalogue and also the Text using the shorter title. Because we would have two same texts next to each other.

### DIFF
--- a/secant/Sources/Features/SmartBanner/SmartBannerContent.swift
+++ b/secant/Sources/Features/SmartBanner/SmartBannerContent.swift
@@ -180,13 +180,8 @@ extension SmartBannerView {
                 .padding(.trailing, 12)
             
             VStack(alignment: .leading, spacing: 2) {
-                ViewThatFits {
-                    Text(localizable: .smartBannerContentShieldTitle)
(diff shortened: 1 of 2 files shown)


secant/Sources/Features/SmartBanner/SmartBannerContent.swift
```diff
@@ -183,9 +183,6 @@ extension SmartBannerView {
                 ViewThatFits {
                     Text(localizable: .smartBannerContentShieldTitle)
                         .zFont(.medium, size: 14, color: titleStyle())
-
-                    Text(localizable: .smartBannerContentShieldTitleShorter)
-                        .zFont(.medium, size: 14, color: titleStyle())
                 }
```
LukasKorba: with just 1 Text, the ViewThatFits closure no longer makes sense and can be removed completely

LukasKorba (approved):
Pre-utACKed with 1 nit

PR #1808, opened 2026-06-05
[E2E]: e2e: enable biometric on sim + parallelize CI via build-once + phase matrix

zcash-harry:
## Summary

Three changes that move the iOS e2e smoke suite from "biometric flows
silently bypassed on the simulator + 35-minute serial CI" to "real
LAContext gates exercised + ~17-min CI wall time".

### 1. UserDefaults-gated simulator bypass

`LocalAuthenticationLiveKey.authenticate` short-circuited to `true`
on the simulator for dev convenience, which made every in-flow
biometric gate (Address Book, Recovery Phrase, Send Confirmation,
Export Private Data) untestable end-to-end.

Now gated on a `zodlE2EBiometric` UserDefaults flag:
- Default (no flag) → bypass active, dev experience unchanged.
- E2e launches with `-zodlE2EBiometric YES` → bypass lifted, real
  LAContext.evaluatePolicy fires and the gates work.

The Maestro YAMLs in zodl-qa already pass this argument on every iOS
launchApp; once this PR is in, those gates fire properly in CI.

### 2. CI matrix split: phase 1 / phase 2

Suite now runs as two parallel matrix jobs per platform:
- **Phase 1**: create-new-wallet, navigate-to-receive,
  address-book/delete-contact, currency-conversion
- **Phase 2**: restore-existing-wallet, recovery-phrase,
  export-private-data, send-transaction

Phases are state-independent so they can run on separate sims
concurrently. Wrapper picks up `ZODL_TEST_PHASE` env var to select.
Halves wall time.

### 3. Build once, share .app via artifact

Previous matrix design re-ran `xcodebuild` + SPM resolve in each
phase job (~7-10 min each). Now a single `build` job uploads the .app
bundle; both phase matrix jobs `actions/download-artifact` and
`simctl install` it. Saves ~10 min of CI minutes per run.

## Test plan

- [ ] CI green on both phase 1 and phase 2 jobs.
- [ ] Manually verify the `.app` download step finds the bundle (the
      first run after merge may need the `mv app-bundle …` heuristic
      tweaked depending on how `actions/upload-artifact` packs the
      `.app` directory).
- [ ] No UX change in normal dev launches (bypass stays active for
      Xcode-launched builds).

## Companion PRs

- zodl-qa main already passes `-zodlE2EBiometric YES` on iOS launches
  and supports `ZODL_TEST_PHASE` env var.
- zodl-android PR (`harry/ci-parallel-phases`) carries the parallel
[...]

PR #1809, opened 2026-06-08
[E2E]: bump runner to macos-26 to match local dev macOS

zcash-harry:
Stacked on top of #1808 (parallelize / biometric-toggle PR).

## What

Bump the CI iOS runner from `macos-15` to **`macos-26`** so it matches what devs actually run locally (macOS Tahoe 26). Brings the iOS sim runtime + Xcode version in line with the dev environment.

## Why

The runner version affects:
- Xcode version (newer = closer to what devs use)
- Sim runtime (iOS 26.x sims on macos-26 vs iOS 26.x on macos-15 behave differently — XCTest driver startup time, SwiftUI hit-testing on cold sim, etc.)
- The Maestro driver image's compatibility with the underlying XCTest framework

We've burned a lot of debug cycles this week chasing iOS-26-only quirks on macos-15 runners (Face ID overshoot to Springboard, SwiftUI tap not landing on cold sim). Matching the dev environment closes that gap.

## What changed

Both `runs-on: macos-15` lines (build job + e2e-phase matrix) flipped to `macos-26`.

## Out of scope (now)

This branch previously carried a "persist wallet direction across CI runs via GH repo variable" commit. We dropped that — instead each platform's wrapper now hardcodes its own direction (Android: `ANDROID_TO_IOS`, iOS: `IOS_TO_ANDROID`), each platform tests its native send path against its own seed. That's the realistic config. The wrapper change is in `zodl-qa` main; nothing in this repo depends on it.

## Test plan
- [ ] CI green on this branch
- [ ] Sim version in run log shows iOS 26.x on macos-26 host

### DIFF
--- a/.github/workflows/e2e-smoke.yml
+++ b/.github/workflows/e2e-smoke.yml
@@ -26,7 +26,7 @@ jobs:
   # via actions/upload-artifact. The previous matrix design ran
   # xcodebuild + SPM resolve in each phase job, paying ~7 min twice.
   build:
-    runs-on: macos-15
+    runs-on: macos-26
     timeout-minutes: 25
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


LukasKorba (approved):
LGTM

Chlup (approved):

PR #1772, opened 2026-05-21
Enable Swift 6 and fix all compilation errors

Chlup:
## Goal

Enable Swift 6 language mode and resolve every compilation error the switch
surfaces. The mode flip and the error fixes have to land together — the
project doesn't compile until all errors are fixed, so this work can't be
split into smaller PRs.

## Changes

- Switch Swift language mode to 6 in the project file.
- Add `@MainActor`, `@Sendable`, `Sendable` conformances, and `@preconcurrency`
  imports across ~60 view/feature files and the major dependency clients
  (`SDKSynchronizer`, `BackgroundTaskClient`, `ExchangeRate`, `AddressBookClient`,
  `ShieldingProcessor`, `VotingAPIClient`, `VotingCryptoClient`) wherever
  strict concurrency requires it.
- Change `PreferenceKey.defaultValue` from `static var` to `static let` in
  the voting views to match the rest of the codebase and satisfy Swift 6's
  "nonisolated global shared mutable state" check.

## Out of scope

Swift 6 **warnings** are not fixed here — those land in a follow-up PR.
This PR is only the change required to make the project compile.

PR #1811, opened 2026-06-08
Fix remaining Swift 6 concurrency warnings

Chlup:
## Summary

Follow-up to #1772 (Enable Swift 6). With `SWIFT_VERSION = 6.0` the project compiled but emitted strict-concurrency warnings. This PR clears them without behavior changes; one approved API change to `QRCodeGenerator.generate` is the only public-surface modification.

## Changes

- **MainActor-isolated calls in reducers**: wrap `MFMailComposeViewController.canSendMail()` and `UIImpactFeedbackGenerator(...).impactOccurred()` in `MainActor.assumeIsolated`. TCA's `Store` is `@MainActor` since 1.14 so reducer bodies always run on main, but the `Reducer` protocol itself is nonisolated, so the compiler needs the hint.
- **Static `Store` factories**: mark `Home.placeholder`, `Home.error`, and `Root.*` placeholder factories `@MainActor` to match `StoreOf<...>.init`.
- **`Binding` extension**: mark `Utils/Bindings.swift` extension `@MainActor` so captures into `Binding.init`'s `@Sendable` get/set closures are allowed.
- **Effect captures**: narrow `[state]` captures to specific fields in `SendForm` and `ScanCoordFlow` so `State` doesn't need to be `Sendable`.
- **`SplashView` timer**: read `isOn` via `MainActor.assumeIsolated` inside the `Timer.scheduledTimer` closure (timer was scheduled from `@MainActor` and fires on the main run loop).
- **`QRCodeGenerator.generate` (API change)**: converted from `Combine.Future<CGImage?, Never>` to `async () -> CGImage?`. All 6 call sites updated to use `.run { send in ... }` effects instead of `.publisher { future.map(...) }`.
- **`ZashiButton`**: replaced default-expression-based parameter inference with constrained extensions so Swift 6 isolation crossings don't trip on default values.
- **Drop redundant `await`s** on non-async calls in AddressBook and Splash.

## Out of scope

Non-concurrency warnings (TCA deprecations, iOS deprecations, dead code) — follow-up PR.

## Test plan

- [ ] Unit tests green on CI
- [ ] Manual smoke on testnet: launch, send, scan address, request ZEC, swap

### DIFF
--- a/secant/Sources/Features/AddressBook/AddressBookStore.swift
+++ b/secant/Sources/Features/AddressBook/AddressBookStore.swift
@@ -356,7 +356,7 @@ struct AddressBook {
                 if requestToSync {
                     return .run { send in
                         do {
-                            let result = try await addressBook.syncContacts(account.account, abContacts)
+                            let result = try addressBook.syncContacts(account.account, abContacts)
                             let syncedContacts = result.contacts
                             if result.remoteStoreResult == .failure {
                                 // TODO: [#1408] error handling https://github.com/Electric-Coin-Company/zashi-ios/issues/1408
--- a/secant/Sources/Features/AddressDetails/AddressDetailsStore.swift
+++ b/secant/Sources/Features/AddressDetails/AddressDetailsStore.swift
@@ -99,26 +99,27 @@ struct AddressDetails {
                 return .none
 
             case .generateQRCode:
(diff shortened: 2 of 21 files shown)


LukasKorba (approved):
LGTM

PR #1817, opened 2026-06-09
Fail archive when PartnerKeys.plist is missing or invalid

Chlup:
## Summary
- Adds `Scripts/validate-partner-keys.sh`, which fails an **archive** build (`ACTION=install`) when `secant/Resources/PartnerKeys.plist` is missing, malformed, or missing any of the 7 required non-empty `String` keys (`cbProjectId`, `flexaPublishableKey`, `flexaPublishableTestKey`, `nearKey`, `cmcKey`, `nearFeeDepositAddress`, `nearAPIKey`). It is a no-op on normal builds, Run, and tests.
- Wires a **"Validate PartnerKeys.plist"** Run Script phase (immediately after the existing "Generate PartnerKeys.plist" phase) into all three app targets: `zodl-production`, `zodl-internal`, `zodl-testnet`.
- Adds a self-contained bash test harness (`Scripts/tests/validate-partner-keys-test.sh`, 7 cases) and a cross-reference comment in `PartnerKeys.swift` to guard against key-list drift.

## Why
The existing "Generate PartnerKeys.plist" build phase writes an **empty** dummy plist (`<dict/>`) whenever the real file is absent, so the app could be archived with no partner keys — silently breaking the Coinbase, Flexa, NEAR, and CoinMarketCap features in the shipped build. `testSeed` is excluded because it is `#if DEBUG`-only and never compiled into an archive.

## Test plan
- [x] `bash Scripts/tests/validate-partner-keys-test.sh` → 7/7 pass (valid plist, empty dummy, malformed, missing file, empty value, non-String, normal-build no-op)
- [x] Normal build with the phases present → succeeds; build log shows the Validate phase running with no issues (no-op on `ACTION=build`)
- [ ] Archive with the real plist → succeeds; archive log shows `PartnerKeys.plist OK — all required keys present.`
- [ ] Archive with the plist moved aside → fails with the 7 missing-key errors, then restore

Design & plan: `docs/superpowers/specs/2026-06-09-partner-keys-archive-validation-design.md`, `docs/superpowers/plans/2026-06-09-partner-keys-archive-validation.md`.

> Note: this branch also carries the pre-existing commit "Deny Claude access to partners keys file" (`.claude/settings.json`).


LukasKorba (approved):
LGTM

PR #1819, opened 2026-06-10
Move transaction guard into dependency choke points; gate auto server switch on live navigation state

Chlup:
## Summary

Redesign of the server-switch vs. transaction-broadcast coordination (design doc: `docs/superpowers/specs/2026-06-10-transaction-guard-server-switch-design.md`, implementation plan: `docs/superpowers/plans/2026-06-10-transaction-guard-server-switch.md`):

- **Guard at choke points (correctness):** `transactionGuard.withSubmission` moves from 11 scattered call sites into the live implementations of six dependency closures — `sdkSynchronizer.createProposedTransactions` / `createTransactionFromPCZT` / `getTreeState` and `votingAPI.submitVoteCommitment` / `submitDelegation` / `delegateShares`. New broadcast paths physically cannot forget the guard; nesting (the old deadlock hazard) is impossible because feature code no longer wraps broadcasts. The `TransactionGuard` actor itself is unchanged.
- **Auto-switch defers during sensitive flows (UX):** `AutoServerSelectionClient` splits into `findBestServer()` (benchmark) and `applySwitch(_:)` (re-validates, then switches under `switchIfIdle`). Root routes candidates through a new `.autoServerCandidateReady` action that reads `Root.State.canApplyAutoServerSwitch` **live at decision time** — no event-tracked UI state anywhere. While the user is in send/scan/swap/transactions/settings (voting), Keystone signing, Server Setup (all three entry points incl. the smart-banner path), or a background task, the candidate is stashed; an `onChange` on the gate re-feeds it when the flow exits (15-minute TTL). `isSensitiveFlowActive` is an exhaustive no-default switch over `Root.State.Path`, so adding a flow case won't compile until classified.
- **Observability + docs:** defer / apply-after-exit / TTL-expired / guard-busy-skip log events; CLAUDE.md guard section rewritten for the choke-point model (including the deliberate `resubmitShare` exemption).

Motivating scenario this fixes: send → background → foreground re-runs the endpoint benchmark → the automatic switch could previously land mid-flow (e.g. between voting's commit and share delegation, or during Keystone QR signing). Now each broadcast is individually guarded at the dependency layer and the switch waits for the flow to end.

## Test Plan

- [x] Full suite green: 325/325 (scheme `zodl-internal`), verified independently in review
- [x] 16 unit tests for the new behavior: `AutoServerSelectionClientTests` (findBestServer ×4, applySwitch ×5 incl. guard-busy and thrown-switch paths) and `RootAutoServerGatingTests` (gate classification for all 12 Path cases, Keystone binding, all Server Setup entry points, TTL boundary)
- [ ] Manual QA on device/simulator (Automatic server selection ON): start a send → background → foreground → finish the flow; expect `[AutoServerSelection] Candidate deferred (… sensitiveFlow: true)` then `Applying deferred candidate after flow exit` in the console; manual Save on Server Setup still switches immediately


Chlup:
Wrong target branch.

PR #1824, opened 2026-06-12
[PRO-334] Fail pending sent tx when chain tip is past its expiry

zcash-harry:
## What

Stops the wallet from rendering an unminable sent tx as "Sending"
indefinitely once the chain tip has passed its `expiry_height`.

Adds a `currentChainTip` parameter to `TransactionState.init` and, when
both that parameter and the tx's `expiry_height` are known, treats an
unmined sent tx as failed when `chainTip >= expiry`. The single call
site in `SDKSynchronizerLive` is updated to pass
`synchronizer.latestState.latestBlockHeight` (or `nil` when the wallet
hasn't synced yet, preserving the original behaviour during fresh
launch).

## Why

Reported in [PRO-334]. Repro:

1. User sends a tx on a pre-HF version of the app (e.g. v3.4.1)
2. Tx broadcasts but is never mined — the upcoming consensus rule
   change makes it permanently unminable
3. User updates the app to a post-HF version (e.g. v3.5.3)
4. The Activity list keeps showing the tx as "Sending" forever

Root cause is upstream in the SDK: `Overview.State.init` only returns
`.expired` when its `expired_unmined` column has been flipped, and that
column doesn't reliably catch up for txs that were unmined when the
wallet migrated across a consensus-rule change. The SDK fix is filed in
[zcash/zcash-swift-wallet-sdk#1756]; this PR is the immediate user-
facing fix, scoped so it can ship before the SDK release cycle.

## Safety

- Funds are not at risk either way: the signed tx bytes are bound to
  the old consensus branch, no validator post-HF will accept them, and
  the SDK's `TxResubmissionAction` already refuses to resubmit
  (`expiry > tip` filter)
- The fallback only fires for sent txs with `minedHeight == nil` AND a
  known `currentChainTip` AND a known `expiry > 0` AND `tip >= expiry`.
  Each condition matches reality before we flip the status
- `currentChainTip` is `nil` during pre-sync, so a freshly-launched
  wallet keeps the current "render as Sending until we know better"
  behaviour

## Once the SDK fix lands

The SDK PR makes `Overview.State` itself return `.expired` in this case,
which means `transaction.state == .expired` will be true on its own and
this fallback becomes redundant (but harmless). We can revert it then
or leave the defence-in-depth.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ directly impact users rather than highlighting other crucial architectural updat
 ### Added
 - Server selection now offers an Automatic mode that benchmarks known servers and keeps your wallet on the fastest one; Manual mode still lets you pin a specific server. Automatic switching is paused while sending, swapping, shielding, or voting.
 
+### Fixed
+- A sent transaction that broadcast on an older app version and never mined now correctly shows as "Failed" in the Activity list once its expiry passes the network chain tip. Previously it could stay stuck on "Sending" indefinitely after updating the app.
+
+### Fixed
+- A sent transaction that broadcast on an older app version and never mined now correctly shows as "Failed" in the Activity list once its expiry passes the network chain tip. Previously it could stay stuck on "Sending" indefinitely after updating the app.
+
 ## 3.5.2 build 1 (20026-06-08)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


Chlup (changes requested):
Please add record to the CHANGELOG. Otherwise this looks good to go.

Chlup (approved):
utACK

PR #1826, opened 2026-06-12
[MOB-1363] Require local auth before recovery phrase is loaded and shown

Chlup:
## Linear

[MOB-1363 — [Security] Recovery phrase settings route bypasses the intended local-auth gate](https://linear.app/zodl/issue/MOB-1363/security-recovery-phrase-settings-route-bypasses-the-intended-local) (sub-issue of MOB-1276, finding ZODL-IOS-02)

## Problem

Two layers of the recovery-phrase display skipped the intended protections:

1. **Route bypass:** the Recovery Phrase row in Advanced Settings sent `.operationAccessGranted(.recoveryPhrase)` directly instead of `.operationAccessCheck(.recoveryPhrase)`, skipping the Face ID / Touch ID gate that every other sensitive row (Export private data, Tax export, Delete wallet, …) goes through.
2. **Pre-auth seed in memory and view tree:** `RecoveryPhraseDisplay` exported the wallet in `onAppear` and rendered the real seed words as SwiftUI `Text` nodes that were merely **blurred**. The words were present in the accessibility tree and in view-hierarchy snapshots before any authentication.

## Fix

- Advanced Settings row now sends `.operationAccessCheck(.recoveryPhrase)` → LocalAuthentication is required to even open the screen from Settings.
- `RecoveryPhraseDisplay` no longer loads anything in `onAppear`. The seed phrase + birthday are loaded from the keychain **only** in the reveal path, after `localAuthentication.authenticate()` succeeds (`.recoveryPhraseUnhideRequested` → auth → `.recoveryPhraseRevealed`).
- Pre-reveal, the grid renders placeholder dots (`•••••`) — no seed-word text nodes, no accessibility content, nothing snapshot-visible.
- Hiding via any path (Hide button, eye-off nav button, app backgrounded / resigns active, screen re-appears) purges the phrase and birthday from TCA state.
- Removed dead `twoColumnSeed()` and the now-unreachable "no words" fallback + its localization key.

This also hardens the wallet-backup flow (same screen): reveal there already required auth per tap, but the seed used to sit in state pre-auth; now it doesn't.

## Verification

- ✅ Full test suite passes; new `RecoveryPhraseDisplayTests` (5 tests) cover: no seed load on appear, reveal denied without auth, seed loaded only after successful auth, purge on hide and on re-appear

## Manual testing steps

1. **Settings route now gated:** Settings → Advanced Settings → Recovery Phrase. **Expected:** Face ID / Touch ID / passcode prompt appears immediately. Cancel the prompt → you stay in Advanced Settings. Approve → Recovery Phrase screen opens with all 24 words shown as blurred dots.
2. **Reveal requires its own auth:** On the Recovery Phrase screen tap **Reveal security details**. **Expected:** another biometric prompt; only after success do the real words + birthday appear.
3. **No seed behind the blur (key check):** Before tapping Reveal, enable VoiceOver (Settings → Accessibility) and swipe through the word grid. **Expected:** VoiceOver reads dots/placeholders, never real seed words. (Pre-fix it could read the real words.) Alternatively inspect with Xcode → Debug View Hierarchy: the word cells must contain `•••••` strings only.
4. **Hide purges:** Reveal the phrase, then tap **Hide security details**. **Expected:** dots again. Re-reveal requires authentication again.
5. **Backgrounding purges:** Reveal the phrase, swipe to app switcher / background the app, return. **Expected:** phrase is hidden (dots) and the app-switcher snapshot never shows the words.
6. **Backup flow regression:** Create a new wallet (or open Wallet Backup from the home banner) → security warning → continue → Reveal. **Expected:** auth prompt, then words; "I've saved it" / "Remind me later" still work.
7. **Failure path:** With Face ID covered/failed and passcode cancelled, tap Reveal. **Expected:** words stay hidden as dots; no error state corruption; tapping Reveal again re-prompts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ directly impact users rather than highlighting other crucial architectural updat
 
 ### Fixed
 - A sent transaction that broadcast on an older app version and never mined now correctly shows as "Failed" in the Activity list once its expiry passes the network chain tip. Previously it could stay stuck on "Sending" indefinitely after updating the app.
+- Opening the Recovery Phrase from Advanced Settings now always requires Face ID / Touch ID, and the seed words are only loaded and rendered after a successful authentication.
 
 ### Removed
 - A hidden legacy debug menu (reachable via a gesture on the splash screen) that could copy the seed phrase to the clipboard without Face ID / Touch ID.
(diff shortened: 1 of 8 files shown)


LukasKorba (approved):
LGTM

PR #1827, opened 2026-06-12
[MOB-1368] Delete tax CSV export from temp storage after sharing

Chlup:
## Linear

[MOB-1368 — [Security] Tax CSV export remains in the app temp directory after sharing](https://linear.app/zodl/issue/MOB-1368/security-tax-csv-export-remains-in-the-app-temp-directory-after) (sub-issue of MOB-1276, finding ZODL-IOS-08)

## Problem

`TaxExporterClient` serialized the transaction history CSV to a **fixed, predictable path** directly in the app temporary directory, with no file-protection class. The export reducer kept the URL in state and never deleted the file — the share sheet closing only reset UI flags. Transaction-history CSVs (dates, amounts, fees, account context) persisted in temp storage indefinitely.

A second subtlety: `UIShareDialogView`'s `completion` fires when the sheet finishes *presenting* (it's the `present(animated:completion:)` callback), so there was no hook at all for "user finished/cancelled sharing".

## Fix

- Each export is written to a unique `tmp/tax-exports/<UUID>/` directory created with `FileProtectionType.complete`; the CSV itself is written with `.completeFileProtection`.
- `UIShareDialogView` gained an optional `onDismiss` callback wired to `UIActivityViewController.completionWithItemsHandler`, which fires on **both** completed share and cancel. All other call sites are untouched (parameter defaults to `nil`).
- The export screen sends a new `.shareSheetClosed` action from that hook; the reducer clears `dataURL` and calls the new `taxExporter.cleanupExports()`, which removes the whole `tax-exports` directory.
- Each new export first purges stale exports from previous runs, covering the app-killed-while-sheet-open case.

## Verification

- ✅ Full suite passes (315 tests), incl. 3 new `ExportTransactionHistoryTests`: cleanup on share-sheet close, unique per-export directory + stale purge, idempotent `cleanupExports`

## Manual testing steps

> Tip: to inspect the app's tmp directory on Simulator, run the app, then `xcrun simctl get_app_container booted co.electriccoin.zcash data` (adjust bundle id for the installed variant) and look into `tmp/`.

1. **Export + complete share:** Settings → Advanced Settings → Export tax file (pass Face ID) → Download → share sheet opens → pick *Save to Files* and save. **Expected:** the saved CSV opens fine from Files; afterwards the app's `tmp/tax-exports` directory does **not** exist.
2. **Export + cancel:** Same flow, but dismiss the share sheet (swipe down / Cancel). **Expected:** `tmp/tax-exports` is removed right after the sheet closes.
3. **Share still works while sheet open:** With the sheet open, wait a few seconds, then choose AirDrop or Save to Files. **Expected:** sharing succeeds (file is only deleted when the sheet closes, not when it appears).
4. **Stale cleanup:** Start an export, and while the share sheet is open force-kill the app (the CSV stays behind — pre-existing behavior on kill). Relaunch, run another export. **Expected:** after the new export, only the new `<UUID>` directory exists under `tmp/tax-exports`, the stale one is gone; and after closing the sheet, everything is gone.
5. **Repeat exports:** Run Download twice in a row in the same session. **Expected:** second export works, share sheet shows the CSV, no accumulation of old files.
6. **Regression — other share dialogs:** Test a couple of other share flows (e.g. Receive → share address, Transaction details → share, Send feedback → share). **Expected:** they behave exactly as before (the `onDismiss` hook is opt-in).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ directly impact users rather than highlighting other crucial architectural updat
 ### Fixed
 - A sent transaction that broadcast on an older app version and never mined now correctly shows as "Failed" in the Activity list once its expiry passes the network chain tip. Previously it could stay stuck on "Sending" indefinitely after updating the app.
 - Opening the Recovery Phrase from Advanced Settings now always requires Face ID / Touch ID, and the seed words are only loaded and rendered after a successful authentication.
+- The tax CSV export is now written to a protected location and deleted as soon as the share sheet closes, instead of remaining in temporary storage.
 
 ### Removed
 - A hidden legacy debug menu (reachable via a gesture on the splash screen) that could copy the seed phrase to the clipboard without Face ID / Touch ID.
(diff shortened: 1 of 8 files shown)


LukasKorba (approved):
utACK

PR #1830, opened 2026-06-12
[MOB-1364] Update swapAPIAccess routing flag on every runtime Tor toggle

Chlup:
## Linear

[MOB-1364 — [Security] Tor enablement does not update the shared network flag until restart](https://linear.app/zodl/issue/MOB-1364/security-tor-enablement-does-not-update-the-shared-network-flag-until) (sub-issue of MOB-1276, finding ZODL-IOS-03)

## Problem

The shared `swapAPIAccess` flag decides per request whether the **exchange-rate** (`ExchangeRateLiveKey`), **swap-and-pay** (`Near1Click`) and **voting** (`VotingAPIClientLiveKey`) clients use `URLSession.shared` or the SDK's `httpRequestOverTor` path. The only write was in `SDKSynchronizerClient.live()` — once, at construction, from the stored Tor setting. Runtime Tor toggles called only `sdkSynchronizer.torEnabled(_)`, so:

- enabling Tor mid-session → those requests **kept egressing directly** (IP/network metadata exposure) until the next launch, while the user believed protected networking was active;
- disabling Tor mid-session → requests kept trying the Tor path.

## Fix

`swapAPIAccess` is now updated at every runtime toggle point:

1. **`SDKSynchronizerLive.torEnabled` (choke point):** any caller of `sdkSynchronizer.torEnabled` updates the flag automatically. The flag is set **before** the SDK call, so if Tor init throws, routing **fails closed** (stays on the protected path, requests error out) rather than silently leaking via direct networking.
2. **`TorSetup` reducer:** `.enableTapped`, `.saveChangesTapped` (both directions), `.disableTapped` update the shared flag synchronously — the next request after the tap already routes correctly.
3. **`Root.torDisableTapped`** (the "Tor init failed → disable" alert path): this path never calls `torEnabled` (call commented out), so it now sets the flag to `.direct` explicitly.

All three consumers read `@Shared(.inMemory(.swapAPIAccess))` at request time, so the change takes effect immediately, no restart needed.

## Verification

- ✅ Full suite passes (317 tests), incl. 5 new `TorSetupTests`: protected immediately on enable, **fail-closed when Tor init fails**, direct immediately on disable, and both save-changes directions

## Manual testing steps

> Observing routing: easiest with a local HTTP proxy (Proxyman/Charles/mitmproxy) configured on the Simulator, or Console.app filtering the app's network logs. Direct requests to `api.coinmarketcap`-backed rate endpoint / `1click.chaindefuser.com` (swap) are visible to the proxy; Tor-routed requests are not (they go through the SDK's Tor client).

1. **Enable Tor mid-session routes immediately:** Start the app with Tor OFF and currency conversion ON — observe the exchange-rate request hitting the proxy directly. Then Settings → Advanced Settings → Private (Tor) → Enable (or select opt-in + Save). Pull-to-refresh the rate (or open the swap flow to fetch assets). **Expected:** no new direct requests appear in the proxy from this moment — without restarting the app. (Pre-fix: requests kept appearing until relaunch.)
2. **Disable Tor mid-session routes immediately:** With Tor ON, disable it in the same screen. Refresh the rate. **Expected:** the request now appears in the proxy (direct) immediately.
3. **Fail-closed on Tor init failure:** Enable Tor in airplane mode-ish conditions (or otherwise make Tor bootstrap fail — e.g. block the network right after tapping Enable) so the "Tor failed" alert appears. Choose "Don't disable". Refresh the exchange rate. **Expected:** the request does NOT egress directly (it fails) — privacy is preserved while the user's Tor preference is ON.
4. **Failure alert → Disable:** Repeat step 3 but choose "Disable Tor" in the alert. **Expected:** requests route directly again immediately, and the Tor setting shows disabled.
5. **Restart consistency:** After each of the above, kill and relaunch the app. **Expected:** routing matches the persisted setting (same behavior as before this PR).
6. **Voting + swap surfaces:** With Tor ON, open the swap flow (asset list fetch) and the voting screen (round fetch). **Expected:** no direct hits in the proxy for these either.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ directly impact users rather than highlighting other crucial architectural updat
 - Opening the Recovery Phrase from Advanced Settings now always requires Face ID / Touch ID, and the seed words are only loaded and rendered after a successful authentication.
 - The tax CSV export is now written to a protected location and deleted as soon as the share sheet closes, instead of remaining in temporary storage.
 - Exported support logs no longer leave plaintext files behind: staging files are removed right after the ZIP is built and the ZIP is deleted when the share sheet closes.
+- Turning Tor on or off now applies to exchange-rate, swap and voting requests immediately instead of after the next app launch.
 
 ### Removed
 - A hidden legacy debug menu (reachable via a gesture on the splash screen) that could copy the seed phrase to the clipboard without Face ID / Touch ID.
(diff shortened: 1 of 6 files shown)


LukasKorba (approved):
LGTM

PR #1832, opened 2026-06-12
[MOB-1365] Verify voting proposal metadata against on-chain proposals_hash

Chlup:
## Linear

[MOB-1365 — [Security] Voting proposal metadata is not authenticated against proposals_hash before vote signing](https://linear.app/zodl/issue/MOB-1365/security-voting-proposal-metadata-is-not-authenticated-against) (sub-issue of MOB-1276, finding ZODL-IOS-05)

## Problem

`authenticateVotingSession` verified the round's signed **EA admin key**, but the proposal metadata (IDs, titles, descriptions, option labels) and the `proposals_hash` itself were parsed from the vote-server response and carried **unverified**. The vote commitment later signs a `proposalId`/`choice` derived from that metadata. A malicious or compromised vote server holding valid EA-key material could therefore display altered proposal text or swapped option labels while the wallet signs a vote for the attacker-selected numeric proposal/choice.

## Fix

- New `ProposalsHasher` recomputes the **ZIP 1244 §"Proposals Hash"** canonical form: proposals sorted by `id`, options by `index`, no whitespace, fixed key order (`id`, `title`, `description`, `options[{index,label}]`), string bytes matching Rust `serde_json::to_string` (no `\/` escaping) — then SHA-256.
- `authenticateVotingSession` (the single gate all round fetches go through: `/rounds`, `/rounds/active`, `/round/{id}`) now requires `computedHash == round.proposals_hash` **after** the EA-key check, and **fails closed** — a mismatching (or missing) hash hides the round behind the same "no active round" surface as failed EA auth, before anything is displayed or signable.
- Option descriptions, ZIP numbers and forum links are UI-only metadata outside the ZIP canonical form (verified below) — they don't affect the hash.

> **Note on the "ZIP 1244" naming.** "ZIP 1244" throughout this PR (and in the code comments) is **not a ratified/assigned ZIP number** — it is the number of the GitHub pull request [`zcash/zips#1244`](https://github.com/zcash/zips/pull/1244), which holds the *open draft* spec "Shielded Voting Wallet API" proposed by Valar Group (the vote-chain operator). The PR is literally titled **"[ZIP TBD]"**: a ZIP number has not been assigned yet, and `1244` is just the pull-request number in the `zcash/zips` repo. The `§"Proposals Hash"` canonical form implemented here is defined in that draft, and the example vector pinned in the tests (`3f9a361d…`) is the draft's own worked example. Because the draft is unmerged and still evolving, a later change to its canonical form would require a matching wallet update — the verification fails closed until then.

### Why this canonicalization is correct

1. It is the **same scheme this codebase used** to verify CDN-config proposals against the chain hash before proposals moved to chain rounds (removed only because the CDN stopped publishing proposal JSON — see `docs/voting-service-discovery.md`), including its pinned ZIP 1244 test vectors.
2. **Validated against production**: all **7 rounds** currently served by `vote-chain-primary.valargroup.org` recompute exactly to their `proposals_hash` with this scheme (including proto3 zero-index omission and a 15-proposal round).

## Verification

- ✅ Full suite passes (322 tests), incl. 10 new `ProposalsHasherTests`: ZIP 1244 example vector, serde_json slash vector, sorting/escaping, and the acceptance-criteria tamper cases — modified title, swapped option labels, changed numeric id (hash changes) vs. benign array reorder and UI-only metadata (hash unchanged)

## Manual testing steps

1. **Happy path (production data):** Open the Coinholder Polling flow with a wallet on mainnet. **Expected:** rounds list and ballots load exactly as before (all current production rounds recompute correctly — verified against the live API during development).
2. **Tampered metadata fails closed (MITM harness):** Using a local HTTPS proxy (mitmproxy/Charles with the proxy CA trusted on a dev device/Simulator and Tor disabled), intercept the response of `/shielded-vote/v1/rounds` (or `/round/{id}`) and change one proposal's `title` (or swap two option `label`s, or change a proposal `id`) **without touching `proposals_hash`**. **Expected:** the affected round disappears from the list (treated as unauthenticated; log line `Round auth failed: proposals_hash mismatch round=…` in Console). No ballot with tampered text is ever rendered.
3. **Tampered hash also fails:** Same setup, but change `proposals_hash` instead. **Expected:** same fail-closed behavior.
4. **Missing hash fails:** Remove the `proposals_hash` field from the round JSON. **Expected:** round hidden (fail closed).
5. **Vote submission regression:** On an active testnet/staging round, complete a full vote (delegate → ballot → sign → submit). **Expected:** unchanged behavior; the tally/results screens still render.
6. **Tally-phase rounds:** Confirm finished rounds with results still display (they go through the same authentication gate).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ directly impact users rather than highlighting other crucial architectural updat
 - The tax CSV export is now written to a protected location and deleted as soon as the share sheet closes, instead of remaining in temporary storage.
 - Exported support logs no longer leave plaintext files behind: staging files are removed right after the ZIP is built and the ZIP is deleted when the share sheet closes.
 - Turning Tor on or off now applies to exchange-rate, swap and voting requests immediately instead of after the next app launch.
+- Coinholder Polling now verifies the proposal titles, descriptions and option labels it displays against the round's on-chain `proposals_hash` before showing the ballot or accepting a vote, and refuses tampered rounds.
 
 ### Removed
 - A hidden legacy debug menu (reachable via a gesture on the splash screen) that could copy the seed phrase to the clipboard without Face ID / Touch ID.
--- a/secant.xcodeproj/project.pbxproj
+++ b/secant.xcodeproj/project.pbxproj
@@ -49,6 +49,10 @@
 		349CE5EA2FDC12D400E2B9CF /* MultiServerSubmitRoutingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 349CE5E92FDC12D400E2B9CF /* MultiServerSubmitRoutingTests.swift */; };
 		349CE5F82FDC167400E2B9CF /* RecoveryPhraseDisplayTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 349CE5F72FDC167400E2B9CF /* RecoveryPhraseDisplayTests.swift */; };
 		349CE65A2FDC358500E2B9CF /* AdvancedSettingsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 349CE6592FDC358500E2B9CF /* AdvancedSettingsTests.swift */; };
+		349CE63A2FDC259F00E2B9CF /* ProposalsHasher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 349CE6392FDC259F00E2B9CF /* ProposalsHasher.swift */; };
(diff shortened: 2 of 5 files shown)


Chlup:
@greg0x Can you please verify if this change makes? To be honest I am not sure. During testing all existing past votes were displayed but when we tried to create new one (with @LukasKorba) then new one was filtered out on the app side based on this new hash check.

PR #1840, opened 2026-06-16
Switch project from Classic groups to Synchronized groups (PBXFileSystemSynchronizedRootGroup)

Chlup:
## Summary

Switches the Xcode project from **Classic groups** to **Synchronized groups** (`PBXFileSystemSynchronizedRootGroup`, introduced in Xcode 16). Instead of every file being explicitly listed in `project.pbxproj`, Xcode now synchronizes the `secant/` and `zodlTests/` folders with the file system automatically.

**The migration itself changes no source code, build settings, or dependencies** — purely how the project file tracks files (~5.9k deletions / ~150 insertions in `project.pbxproj`). One small follow-up (see *PartnerKeys.plist handling* below) removes 3 dead build phases and adjusts one CI workflow that the migration surfaced.

## Why

- Removes the huge, merge-conflict-prone file/group lists from `project.pbxproj`.
- New files added under `secant/` / `zodlTests/` are picked up automatically — no more manual target-membership wrangling or forgotten files.

## What's in the diff

- `secant/` becomes a synchronized root group for the three app targets (`zodl-production`, `zodl-internal`, `zodl-testnet`).
- `zodlTests/` becomes a synchronized root group for the test target.
- Membership exceptions preserve the previous behavior — Info.plists, `Preview Assets.xcassets`, the SwiftGen `*.stencil` templates and `swiftgen.yml` are kept out of the targets/bundle where they don't belong.
- `objectVersion` bumped `56 → 70`.

## ⚠️ Requires Xcode 16+

Synchronized groups need `objectVersion = 70`. After this merges, anyone on Xcode 15 — **including CI** — won't be able to open or build the project. Please confirm the CI image and the team are on Xcode 16+.

## PartnerKeys.plist handling (follow-up)

`PartnerKeys.plist` is gitignored and was created at build time by a "Generate PartnerKeys.plist" phase, then bundled via the classic explicit file reference. Synchronized groups resolve bundle membership from the on-disk file tree *before* build phases run, so on a clean checkout the plist no longer landed in the `.app` — which broke the `e2e-smoke` workflow's post-build key injection.

Fixed here:
- Removed the 3 now-dead "Generate PartnerKeys.plist" build phases (they ran too late to affect sync-group membership).
- `e2e-smoke.yml` writes an empty `PartnerKeys.plist` stub before the build, so the sync group includes it in the bundle; the real NEAR keys are still injected post-build.

App behavior is unchanged: when the plist is absent the loader returns `nil` for every key and partner features stay disabled (no crash), exactly as before. `pull-request.yml` needs no change — it never requires the plist in the bundle.

## Verification

- Builds, runs, and the full test suite passes locally.
- A structural review of the `project.pbxproj` diff confirmed the migration is behavior-preserving:
  - **Compiled sources identical per target** — 463 files × 3 app targets (byte-for-byte equal sets) + 35 test files; nothing added or dropped, all present on disk.
  - **Copy Bundle Resources identical** — 38 resources per app target.
  - **Build settings (`XCBuildConfiguration`) byte-for-byte identical** — incl. `INFOPLIST_FILE`, `DEVELOPMENT_ASSET_PATHS`, entitlements.
  - **Shell-script build phases unchanged by the migration** — SwiftGen, SwiftLint, Validate PartnerKeys (the dead Generate PartnerKeys phases are dropped in the follow-up above).
  - **Package/framework linkage unchanged** — 14 Swift packages × 3 targets.
  - Only intentional bundle delta: `Preview Assets.xcassets` and the SwiftGen `.stencil` templates are no longer copied into the app bundle (previews still work via `DEVELOPMENT_ASSET_PATHS`; SwiftGen reads templates from disk) — a minor cleanup.


LukasKorba (approved):
utACK

PR #1841, opened 2026-06-16
Fix LoggerTests parallel-suite OSLog pollution

zcash-harry:
## Problem

\`LoggerTests.walletLoggerLogsViaProxy\` fails on CI (\`run 27640596824\`, main HEAD):

> Expectation failed: (logs.count → 7) == 1

\`LoggerProxy.info\` writes to a process-global \`walletLogger\`. While this test holds \`walletLogger\` set to category \"testWalletLogger\", *parallel* suites that indirectly call \`LoggerProxy.info\` (\`ServerHealthTracker\`, \`VotingAPIClientLiveKey\`, \`BackgroundTaskClientLiveKey\`, etc.) race-write into the same OSLog category. \`@Suite(.serialized)\` only serializes within this suite, not across the bundle. Result on CI: 1 from this test + 6 from parallel suites = 7.

## Fix

- Assert \"our message is in the bucket\" (\`count >= 1\` + \`.contains { ... == testMessage }\`) instead of demanding exclusive bucket ownership we can't enforce given process-global state.
- \`defer { walletLogger = nil }\` so the global doesn't keep landing parallel suites' logs in this category after the test exits.
- In-place comment explains the constraint so future readers don't try to tighten the assertion back to \`== 1\`.

## Verification plan

- Re-run unit tests on this PR's CI — \`walletLoggerLogsViaProxy\` should pass.
- Manual: \`xcodebuild test -scheme zodl-internal -only-testing zodlTests/LoggerTests\` locally also passes (was passing before; the race only manifested under parallel-suite load on CI).

### DIFF
--- a/.github/workflows/e2e-smoke.yml
+++ b/.github/workflows/e2e-smoke.yml
@@ -290,14 +290,22 @@ jobs:
       # Key is per-run-id so each successful run produces a fresh
       # cache entry; restore-keys falls back to the most recent
       # matching prefix.
+      #
+      # Prefix bumped v1 → v2 (2026-06-16) to invalidate existing
+      # caches. A stale v1 cache from a past run had scanned_tip at
+      # chain tip but left an unscanned range covering the wallet's
+      # spend transactions — wallet reported 0.39 ZEC total / 0
+      # spendable while the actual on-chain balance was ~0.1 ZEC.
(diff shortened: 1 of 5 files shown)


Chlup (approved):
utACK

zcash-harry:
> utACK

Thank you, still improving the E2E flakiness.

PR #1845, opened 2026-06-18
Fix 4 bugs surfaced by new unit tests

Chlup:
## Summary

New unit tests added for previously-uncovered utilities surfaced four real bugs. This PR fixes each bug and flips the corresponding tests (previously wrapped in `withKnownIssue`) to assert the corrected behaviour.

- **`TransactionsManager` `.unread` filter** — returned `true` for *every* transaction. Now matches only genuinely-unread transactions (not sent, not shielding, has a memo, and not yet marked read).
- **`ServerConfig.endpoint(for:)`** — multi-colon hosts were mis-parsed: a 3-component `host:host:port` lost its `:` separator, and any host with more colons (e.g. IPv6) produced an *empty* host. Now splits on the **last** colon and takes the host verbatim, so every colon is preserved exactly with no duplicate separator; empty-host inputs return `nil`.
- **`NetworkError.allowsRetry`** — the `501...504` check was inverted/backwards. Now retries 5xx server errors and rejects 4xx client errors.
- **`@Clamped` setter** — clamped the *current* stored value instead of the newly-assigned one, making assignments silent no-ops. Now clamps the incoming value.

## Test Plan

- [x] Full unit-test suite: **635 passed, 0 expected failures** (per commit)
- [x] The four affected tests now assert the corrected behaviour instead of documenting the bug via `withKnownIssue`.
- [x] Hardened `endpoint(for:)` host parsing (last-colon split) with 3 added tests for multi-colon / IPv6 / missing-host cases; `ServerConfigEndpointTests` (8) and the server-selection suites (16) pass via Xcode.

## Impact on the rest of the code

These are long-standing latent bugs, so the table below summarises what each fix actually touches in the shipping app. Short version: none of the four old behaviours is something other code currently relies on.

| Fix | Consumed by (production) | Effect on the rest of the code |
|---|---|---|
| `@Clamped` setter | Nothing — referenced only by `ClampedTests`. | **None.** Zero production usages, so shipping behaviour cannot change. |
| `TransactionsManager` `.unread` filter | Dormant — never added to the active filter set; `FiltersSheet` doesn't render it and its only reader `isUnreadFilterActive` is unused. | **None today.** The visible "unread" dot is driven by the *separate*, already-correct `isUnread(_:)` helpers, which are unchanged. |
| `ServerConfig.endpoint(for:)` | Saving a custom server, and reading the stored custom server on launch (`ServerSetup`, `ZcashSDKEnvironment`). | **Narrow.** Ordinary `host:port` inputs are unchanged, so existing saved servers are unaffected. Behaviour differs only for colon-containing hosts (now preserved verbatim) and host-less inputs like `:443` / `443` (now `nil` instead of a broken empty-host endpoint). |
| `NetworkError.allowsRetry` | Swap flow only — `SwapAndPayStore.refreshSwapAssets` and the `TransactionDetails` asset refresh; the quote call uses it for its error message only. | **Swap retries only.** Now retries all 5xx and no 4xx. The old logic retried 4xx (wasteful — never succeeds) and skipped 502/503/504 (gave up too early), so nothing benefited from it. |

### DIFF
--- a/secant/Sources/Dependencies/UserPreferencesStorage/UserPreferencesStorage.swift
+++ b/secant/Sources/Dependencies/UserPreferencesStorage/UserPreferencesStorage.swift
@@ -187,8 +187,8 @@ extension UserPreferencesStorage {
                     host = String(first)
                 } else if split.count == 3, let first = split.first {
                     let second = split[1]
-                    
-                    host = "\(String(first))\(String(second))"
+
+                    host = "\(String(first)):\(String(second))"
                 }
                 
                 return LightWalletEndpoint(
--- a/secant/Sources/Dependencies/UserPreferencesStorage/UserPreferencesStorage.swift
+++ b/secant/Sources/Dependencies/UserPreferencesStorage/UserPreferencesStorage.swift
@@ -178,28 +178,26 @@ extension UserPreferencesStorage {
                 input = String(input.dropFirst(http.count))
             }
             
-            let split = input.split(separator: ":")
-            
-            if let portString = split.last, let port = Int(portString) {
-                var host = ""
-                
-                if split.count == 2, let first = split.first {
-                    host = String(first)
-                } else if split.count == 3, let first = split.first {
-                    let second = split[1]
-
-                    host = "\(String(first)):\(String(second))"
-                }
-                
-                return LightWalletEndpoint(
-                    address: host,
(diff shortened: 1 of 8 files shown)


LukasKorba (approved):
utACK

PR #1848, opened 2026-06-20
[MOB-1130]: Hide sync banner by blocks remaining, not percentage

zcash-harry:
## Summary

Replaces the 98% sync-banner threshold with an absolute **3,456 blocks remaining** (~3 days at 75s/block target).

The old percentage scaled with the wallet's birthday height, so older wallets saw the banner disappear while still tens of thousands of blocks behind sync — leading to confusion when funds weren't yet spendable. A fixed block-count threshold gives every wallet the same ~10s-of-remaining-work hide point, regardless of age.

## Changes

- `SmartBannerStore.swift`: new `smartBannerSyncingBlocksThreshold = 3456`; `lastKnownBlocksRemaining` state; threshold checks swap from `< 0.98` to `>= 3456`.
- `SensitiveData.swift`: surface SDK `fullyScannedHeight` through `RedactableSynchronizerState` so the reducer can compute `latestBlockHeight - fullyScannedHeight`.

The banner still shows the percentage in its label; only the hide/show decision changes.

## Test plan

- Build `zodl-internal` ✅
- New wallet: banner hides at the very end of sync (unchanged from before).
- Old wallet (restored from far-back birthday): banner remains visible until the wallet is within ~3,456 blocks of chain tip, then hides cleanly.

### DIFF
--- a/secant/Sources/Features/SmartBanner/SmartBannerStore.swift
+++ b/secant/Sources/Features/SmartBanner/SmartBannerStore.swift
@@ -18,7 +18,7 @@ struct SmartBanner {
         static let remindMe2days: TimeInterval = 86_400 * 2
         static let remindMe2weeks: TimeInterval = 86_400 * 14
         static let remindMeMonth: TimeInterval = 86_400 * 30
-        static let smartBannerSyncingThreshold = 0.98
+        static let smartBannerSyncingBlocksThreshold: BlockHeight = 3456
     }
     
     @ObservableState
@@ -56,6 +56,7 @@ struct SmartBanner {
         var isSyncTimedOutAutoAppeareDisabled = false
         var isWalletBackupAcknowledged = false
         var isWalletBackupAcknowledgedAtKeychain = false
+        var lastKnownBlocksRemaining: BlockHeight = -1
         var lastKnownErrorMessage = ""
         var lastKnownSyncPercentage = -1.0
         var messageToBeShared: String?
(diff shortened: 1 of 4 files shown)


Chlup (changes requested):
### `lastKnownBlocksRemaining` is never reset on sync completion → stale value can spuriously re-show the syncing banner

The new state field `lastKnownBlocksRemaining` (initialized to `-1`) is assigned `max(0, latestBlockHeight - fullyScannedHeight)` only while the synchronizer status is `.syncing`, and the banner trigger guard in `evaluatePriority4` checks `lastKnownBlocksRemaining >= smartBannerSyncingBlocksThreshold` (3456).

When sync completes (`.upToDate`), `lastKnownBlocksRemaining` is never reset — it retains its last `.syncing` value. If `evaluatePriority4` is subsequently evaluated while `walletStatus` is not `.restoring` (for example when the priority ladder is re-run after a wallet-account change or a reconnect), the stale value still satisfies `>= 3456` and spuriously triggers the syncing banner even though the wallet is fully synced.

The previous percentage-based field self-invalidated on completion: its final value was always above the percentage threshold, so the old `lastKnownSyncPercentage < smartBannerSyncingThreshold` guard naturally became false. The new block-count field has no equivalent self-invalidation.

The stale value is read here:

https://github.com/zodl-inc/zodl-ios/blob/f6cf007c47cbd8493fb6ef123385e30c3f7eb616/secant/Sources/Features/SmartBanner/SmartBannerStore.swift#L434-L437

…and the `.upToDate` branch that should clear it does not:

https://github.com/zodl-inc/zodl-ios/blob/f6cf007c47cbd8493fb6ef123385e30c3f7eb616/secant/Sources/Features/SmartBanner/SmartBannerStore.swift#L372-L376

**Suggested fix:** reset `lastKnownBlocksRemaining` (to `-1` or `0`) in the `.upToDate` branch, or gate the `evaluatePriority4` check on the live synchronizer status actually being `.syncing`.

Chlup (approved):
utACK

PR #1849, opened 2026-06-22
[MOB-1348] Fail closed on multi-recipient ZIP-321 payment requests

Chlup:
## Linear

[MOB-1348 — [Security] ZIP-321: confirm iOS exposure & enumerate/display all payments in review (multi-recipient QR)](https://linear.app/zodl/issue/MOB-1348) (sub-issue of MOB-1276; Android mirror MOB-1341 / iOS security report 2026-04-16, dossier finding `#37`)

## Background

A crafted ZIP-321 URI can encode extra payments (`address.1`, `amount.1`, …). On **Android** the review collapses to `payments[0]` while the **full URI** is signed → a hidden recipient gets paid (fund theft). The dossier asserts **iOS is not affected**; this ticket's first job was to confirm that at HEAD, then add a guard + regression test so a future refactor can't regress into the Android behaviour.

## Verification — iOS is NOT exposed (confirmed at HEAD)

- The scan → proposal chokepoint `ScanCoordFlow.getProposal` reads **only `paymentRequest.payments.first`** and builds the proposal via `sdkSynchronizer.proposeTransfer(account, recipient, amount, memo)` — the **single-recipient** SDK API. The raw ZIP-321 URI and any `payments[1…]` are never handed to the SDK.
- A repo-wide search finds only `proposeTransfer` (single recipient) and `proposeShielding` — there is **no** `proposeFulfillingPaymentURI` / URI→multi-output path (the API the Android bug relied on).
- The confirmation screen's address/amount are set from the same `state.recipient`/`state.amount` fed to `proposeTransfer`, so what you review is exactly what gets signed.

(The dossier's line refs `:254,278` drifted; the relevant logic is now `ScanCoordFlowCoordinator.swift:237-279`. Severity is **not** escalated.)

The only residual was non-security: a legitimate multi-payment URI was **silently truncated** to its first payment — and that is exactly the spot a future change could "fix" into a multi-output proposal while leaving a single-recipient review in place.

## Fix

- **Fail-closed guard** at `ScanCoordFlow.getProposal`: a request with anything other than exactly one payment routes to the existing `.requestZecFailed` path instead of silently using `payments[0]`. No new UI or localized strings.
- **Clear carried-over payment fields on rejection**: the guard fails *before* the parse block writes `state.recipient`/`amount`/`memo`, and those fields are never reset elsewhere — so it now clears all three before bouncing back. Otherwise a multi-recipient scan that follows a valid single-payment scan re-pre-filled the send form with the *previous* scan's stale recipient/amount. The reset lives only in the guard's `else`; the other `.requestZecFailed` paths (parse `catch`, `proposeTransfer` failure) intentionally restore the freshly-parsed values.
- **Regression tests** (`zodlTests/ScanTests/ScanCoordFlowZip321Tests.swift`, Swift Testing):
  - multi-payment request → never calls `proposeTransfer`, bounces to the send form (fail closed);
  - single-payment request → calls `proposeTransfer` exactly once for that recipient (regression);
  - multi-payment scan *after* a successful single-payment scan → bounce-back leaves `recipient`/`amount`/`memo` cleared (no stale prefill) and never re-proposes.

## Verification

- ✅ Build succeeds (zodl-internal scheme).
- ✅ New tests pass (3/3).

## Manual testing steps

1. **Single-recipient QR (regression).** Generate a normal `zcash:<addr>?amount=0.001` request QR. Send → Scan it. **Expected:** review shows that address + 0.001 ZEC; confirm + biometrics signs and broadcasts exactly that (unchanged from today).
2. **Multi-recipient QR (the attack).** Build a QR for `zcash:?address=<addrA>&amount=0.001&address.1=<addrB>&amount.1=1.0`. Send → Scan it. **Expected:** the request is rejected (bounces back to the send form with nothing pre-filled into a signable proposal); there is **no** confirmation sheet for 0.001 with a hidden 1.0, and nothing is signed.
3. **Multi-recipient QR right after a valid single-recipient one (stale-state regression).** Scan a normal single-recipient QR (step 1), back out to the send form, then scan the multi-recipient QR (step 2). **Expected:** the rejection leaves the send form clean — the first scan's address/amount do **not** reappear.
4. **Address-only QR (regression).** `zcash:<addr>` (no amount) → Scan. **Expected:** address pre-fills into the send form as before; user enters the amount manually.

Design: `docs/superpowers/specs/2026-06-22-mob-1348-zip321-multi-recipient-guard-design.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ directly impact users rather than highlighting other crucial architectural updat
 
 ## [Unreleased]
 
+### Changed
+- QR / payment-request (ZIP-321) codes that contain more than one recipient are now rejected instead of silently processing only the first recipient, so what you review is always exactly what gets signed.
+
 ## 3.7.0 build 1
 
 ### Changed
--- a/secant/Sources/Features/CoordFlows/ScanCoordFlowCoordinator.swift
+++ b/secant/Sources/Features/CoordFlows/ScanCoordFlowCoordinator.swift
@@ -235,10 +235,20 @@ extension ScanCoordFlow {
                 return .none
                 
             case .getProposal(let paymentRequest):
+                // SECURITY (MOB-1348 / dossier #37, Android mirror MOB-1341): iOS signs only a single
+                // explicit recipient/amount/memo (payments.first) via `proposeTransfer` — the raw
+                // ZIP-321 URI is never handed to the SDK, so hidden multi-recipient payments cannot be
+                // signed. Fail closed on any non-single-payment request instead of silently collapsing
+                // to payments[0]; honouring multiple payments would first require enumerating and
+                // displaying every recipient in review (otherwise it reintroduces the Android bug).
+                guard paymentRequest.payments.count == 1 else {
+                    return .send(.requestZecFailed)
+                }
+
                 guard let account = state.selectedWalletAccount else {
                     return .none
                 }
-                
+
                 do {
                     if let payment = paymentRequest.payments.first {
                         var textMemo = ""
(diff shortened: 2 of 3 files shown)


LukasKorba (approved):
utACK

PR #1851, opened 2026-06-22
[MOB-1352] Guard Flexa against Keystone accounts; end Flexa session on account switch

Chlup:
## Linear

[MOB-1352 — [Security] Flexa: guard against Keystone account; fix session binding](https://linear.app/zodl/issue/MOB-1352) (sub-issue of MOB-1276; dossier "Flexa session binding, no vendor guard")

## Problem

`Root.flexaOnTransactionRequest` resolved the **currently-selected** account, checked only for a `zip32AccountIndex` (no vendor check), then `exportWallet()` → derived a **local** spending key → signed. Flexa's local-seed signing path only makes sense for **Zodl (local-seed)** accounts; a **Keystone** (hardware) account has no on-device seed, so a Keystone selection would sign with the wrong key. The Flexa subscription also wasn't torn down on a wallet-account switch.

## Fix (Tier 1 + 2 of the ticket)

- **Keystone vendor guard** (`RootDestination.flexaOnTransactionRequest`): fail closed when `account.vendor != .zcash` — block and prompt the user to switch to a Zodl account (new localized message), before any proposal/seed-export runs. Plus an `assertionFailure` **backstop** immediately before key derivation so a future refactor can't reintroduce signing a Keystone payment with the local seed.
- **End Flexa session on account switch** (`RootCoordinator.home(.walletAccountTapped)`): `.cancel(id: CancelFlexaId)` so a request from a session opened under the previous account can't bind to the newly-selected one. (The account is already re-resolved at callback time; this removes the stale session entirely.)

## Deferred to a follow-up (Tier 3)

The 2026-06-08 finding **ZODL-IOS-04** (native amount/address/fee confirmation before signing, to defend against a malicious Flexa service) is a larger, distinct UX feature and is tracked separately — see the linked follow-up issue. This PR closes the "wrong key / wrong account" fund-safety hole; the native-confirmation work is additive.

## Tests

- `zodlTests/FlexaTests/FlexaSecurityTests.swift`:
  - **Keystone blocked** (`keystoneAccountIsBlockedAndNeverProposes`) — a **Keystone** account is blocked: an alert is shown, nothing is reported sent, and `proposeTransfer` / the seed path is **never** entered.
  - **Account switch ends the session** (`switchingWalletAccountCancelsOpenFlexaSession`) — with a Flexa session open, switching the wallet account cancels the subscription (`CancelFlexaId`); a request delivered after the switch is dropped (the handler never re-enters). Verified red/green — removing `.cancel(id: CancelFlexaId)` makes this test fail.
- Existing `MultiServerSubmitFlexaRoutingTests` (6) still pass — the Zodl (`.zcash`) happy path is unaffected.

## Verification

- ✅ Build succeeds (zodl-internal).
- ✅ New tests pass (2/2); existing Flexa routing tests pass (6/6).

## Manual test plan

1. **Keystone blocked.** Add/select a **Keystone** account, open Pay with Flexa, start a payment. **Expected:** blocked with a clear "switch to a Zodl account" message; nothing signed.
2. **Zodl works (regression).** Select a Zodl account; complete a Flexa payment as before.
3. **Account switch ends session.** Open Flexa under a Zodl account, switch to another account, return. **Expected:** the prior session can't drive a signed payment against the new account.

Design: `docs/superpowers/specs/2026-06-22-mob-1352-flexa-keystone-guard-design.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ directly impact users rather than highlighting other crucial architectural updat
 ### Changed
 - QR / payment-request (ZIP-321) codes that contain more than one recipient are now rejected instead of silently processing only the first recipient, so what you review is always exactly what gets signed.
 
+### Fixed
+- Paying with Flexa now requires a Zodl (mobile) account and is blocked for Keystone (hardware) accounts, which have no on-device key to sign with. Switching accounts also ends any open Flexa session so a payment can't bind to the wrong account.
+
 ## 3.7.0 build 1
 
 ### Changed
--- a/secant/Resources/Localizable.xcstrings
+++ b/secant/Resources/Localizable.xcstrings
@@ -9624,13 +9624,13 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Flexa payments require a Zodl (mobile) account. Switch to a Zodl account to pay with Flexa."
+            "value" : "Flexa payments require a ZODL account. Switch to a ZODL account to pay with Flexa."
           }
         },
         "es" : {
           "stringUnit" : {
-            "state" : "translated",
-            "value" : "Los pagos con Flexa requieren una cuenta Zodl (móvil). Cambia a una cuenta Zodl para pagar con Flexa."
+            "state" : "needs_review",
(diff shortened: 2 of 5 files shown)


LukasKorba (approved):
utACK

PR #1855, opened 2026-06-22
Fix Linear release tracking: run on version-tag push

zcash-harry:
## What
Make Linear release tracking actually fire on iOS by running the Linear action **inline off the version-tag push**, with a manual `workflow_dispatch` escape hatch.

## Why
The workflow had never run. Originally `push: tags` used the invalid assumption that the file was on the tagged commit — for `push: tags`, GitHub uses the workflow file **as it exists on the tagged commit**, and our release tags are cut from a release branch that predated this workflow, so it wasn't there (`3.6.0` produced zero runs).

I first switched this to `release: published`, but verified that **won't work on iOS**: iOS publishes **no GitHub Releases** (0 in repo history) and has **no GitHub App token**, so a release made with the default `GITHUB_TOKEN` wouldn't cascade to this workflow anyway. That path would need new infrastructure (a GitHub App) to function.

So this runs the Linear action **inline** instead — needs only the existing `LINEAR_ACCESS_KEY`, no GitHub Release and no App token. The tagged-commit limitation is handled the same way Android's `create-release.yml` handles it: a `workflow_dispatch` escape hatch that runs from `main` for any tag, plus the fact that release branches cut from `main` after this merges will carry the file and fire automatically.

(Also corrects the earlier premise: GitHub tag filters aren't minimatch globs — `+` means "one or more", so `[0-9]+.[0-9]+.[0-9]+` was valid all along. Thanks @chlup.)

## Changes
- Trigger on `push: tags: ['[0-9]+.[0-9]+.[0-9]+']` **and** `workflow_dispatch` (with a `tag` input).
- Resolve + validate the tag (`^[0-9]+\.[0-9]+\.[0-9]+$`), checkout it with `fetch-depth: 0`, run `linear/linear-release-action`.
- `permissions: contents: read`; `concurrency` guard keyed on the tag.

## How it runs going forward
- Normal: push `3.7.0` → workflow fires automatically (file is on the tagged commit once release branches are cut from `main` after this merges).
- Old/transition tags: `gh workflow run linear-release.yml -f tag=3.6.0` (runs from `main`).

## Note vs Android
Android (zodl-android#2337) uses `release: published` because it already creates GitHub Releases and has a GitHub App token (the cascade is proven). iOS lacks both, so it uses this simpler inline approach. If you later set up a release-bot GitHub App on iOS, we can converge both to the same model.

### DIFF
--- a/.github/workflows/linear-release.yml
+++ b/.github/workflows/linear-release.yml
@@ -1,43 +1,81 @@
 name: Linear release tracking
 
-# Runs when a GitHub Release is published. The single job runs Linear's release
-# action, which scans commits since the previous release tag for `MOB-###`
-# references and creates a matching release under the iOS Linear pipeline
-# (Settings → Releases → zodl-ios in Linear).
+# Records a release in Linear when a version tag is pushed. Runs Linear's release
+# action, which scans commits since the previous tag for `MOB-###` references and
+# creates a matching release under the iOS Linear pipeline (Settings → Releases →
+# zodl-ios in Linear), with those issues attached.
 #
-# Why `release: published` and not `push: tags`? For tag-push events GitHub runs
-# the workflow file *as it exists on the tagged commit*. Our release tags are cut
-# from a release branch that predates this workflow, so the file isn't present
-# there and nothing fires. `release` events run the workflow from the default
-# branch (main) instead, so this file is always used regardless of what the tag
-# points at — while still releasing exactly the tag's commit range.
+# The Linear action runs INLINE here — no GitHub Release and no GitHub App token
+# are involved. (We can't use an `on: release: published` trigger like Android:
+# iOS publishes no GitHub Releases, and a release made with the default
+# GITHUB_TOKEN would not cascade to other workflows anyway without an App token.)
 #
-# Requires the LINEAR_ACCESS_KEY repo secret, generated against the iOS
-# pipeline in Linear's UI.
+# Two trigger paths:
+#
+#   * push:tags — fires automatically when a `<major>.<minor>.<patch>` tag is
+#     pushed. NOTE: GitHub runs this file *as it exists on the tagged commit*,
+#     not on main. A release branch that forked from main BEFORE this file was
+#     added won't contain it, so the push trigger silently does nothing for that
+#     tag — this is exactly why tag `3.6.0` never fired. Release branches cut
+#     from main after this merges will carry the file and work automatically.
+#
+#   * workflow_dispatch — runs this file *from the selected branch* (default
+#     main), so it works for ANY existing tag regardless of whether the tagged
+#     commit carries this file. Escape hatch for the transition and for tags on
+#     older branches:
+#       gh workflow run linear-release.yml -f tag=3.6.0
+#
+# Requires the LINEAR_ACCESS_KEY repo secret, generated against the iOS pipeline
+# in Linear's UI.
 
 on:
-  release:
-    types: [published]
+  push:
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Existing version tag to track in Linear (e.g. 3.6.0)'
+        required: true
+        type: string
 
 permissions:
   contents: read
 
-# Serialize runs for a given tag. Editing/re-publishing a release can emit
-# `release: published` more than once; this prevents overlapping Linear syncs
-# for the same release. cancel-in-progress: false so a second event waits for
-# the first to finish rather than aborting it mid-sync.
+# Serialize runs for a given tag so a push and a manual dispatch for the same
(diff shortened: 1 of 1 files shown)


Chlup (changes requested):
### Requesting changes

Thanks for chasing this down. Heads up, though: the change rests on an incorrect premise about how GitHub Actions tag filters work. The new `*.*.*` does match real release tags, so the workflow isn't broken by this PR, but I'd like the rationale corrected before merge. Two issues:

**1. The added comment is factually incorrect.**

GitHub Actions `on.push.tags` filters are **not** minimatch globs where `+` is literal. They use GitHub's filter-pattern syntax, in which `+` means "Matches one or more of the preceding character" and `[0-9]` is a character range (see the [filter pattern cheat sheet](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)). GitHub's own documented semver example is `v[12].[0-9]+.[0-9]+`, which matches `v1.10.1` and `v2.0.0`.

So the previous pattern `[0-9]+.[0-9]+.[0-9]+` would in fact have matched `3.5.3` / `3.6.0` — it was a valid (and tighter) semver matcher. The comment's claim that `+` is "a literal `+`" and that the old pattern "only ever matched strings like `3+.5+.2+`, never real release tags like `3.5.3`" is wrong and will mislead future maintainers. Please correct or remove it.

https://github.com/zodl-inc/zodl-ios/blob/63627020c936901016e32c61562f3619d7de77d8/.github/workflows/linear-release.yml#L14-L19

**2. The commit message's evidence does not support "the workflow had never fired."**

The message cites `3.5.0`, `3.5.2`, and `3.5.3` as proof of zero runs, but all three tags were pushed **before** this workflow was added (#1818, merged 2026-06-10). Only `3.6.0` (2026-06-17) postdates the workflow, so those tags cannot show the trigger was broken.

Since the original pattern was actually valid, the real question is whether the workflow ran for `3.6.0`. Worth checking that tag's Actions run history: if it ran, the trigger was never broken (and this PR replaces a correct, tighter pattern with a looser one on a false premise); if it did not, the root cause is elsewhere and this change likely will not fix it.

zcash-harry:
> ### Requesting changes
> 
> Thanks for chasing this down. Heads up, though: the change rests on an incorrect premise about how GitHub Actions tag filters work. The new `*.*.*` does match real release tags, so the workflow isn't broken by this PR, but I'd like the rationale corrected before merge. Two issues:
> 
> **1. The added comment is factually incorrect.**
> 
> GitHub Actions `on.push.tags` filters are **not** minimatch globs where `+` is literal. They use GitHub's filter-pattern syntax, in which `+` means "Matches one or more of the preceding character" and `[0-9]` is a character range (see the [filter pattern cheat sheet](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)). GitHub's own documented semver example is `v[12].[0-9]+.[0-9]+`, which matches `v1.10.1` and `v2.0.0`.
> 
> So the previous pattern `[0-9]+.[0-9]+.[0-9]+` would in fact have matched `3.5.3` / `3.6.0` — it was a valid (and tighter) semver matcher. The comment's claim that `+` is "a literal `+`" and that the old pattern "only ever matched strings like `3+.5+.2+`, never real release tags like `3.5.3`" is wrong and will mislead future maintainers. Please correct or remove it.
> 
> https://github.com/zodl-inc/zodl-ios/blob/63627020c936901016e32c61562f3619d7de77d8/.github/workflows/linear-release.yml#L14-L19
> 
> **2. The commit message's evidence does not support "the workflow had never fired."**
> 
> The message cites `3.5.0`, `3.5.2`, and `3.5.3` as proof of zero runs, but all three tags were pushed **before** this workflow was added (#1818, merged 2026-06-10). Only `3.6.0` (2026-06-17) postdates the workflow, so those tags cannot show the trigger was broken.
> 
> Since the original pattern was actually valid, the real question is whether the workflow ran for `3.6.0`. Worth checking that tag's Actions run history: if it ran, the trigger was never broken (and this PR replaces a correct, tighter pattern with a looser one on a false premise); if it did not, the root cause is elsewhere and this change likely will not fix it.
> 

A potential issue flagged by Linear AI 

The most likely problem is the trigger.
tags: in GitHub Actions uses glob patterns, not regex, so this:
'[0-9]+.[0-9]+.[0-9]+'probably never matches your version tags. If the workflow never runs, Linear never gets a release, which fits what I’m seeing in zodl-ios.
Use a glob such as:
on:
  push:
    tags:
      - '*.*.*'A couple of other checks:
make sure the pushed tag is on the commit range that includes issue refs like MOB-1125confirm LINEAR_ACCESS_KEY was generated from zodl-iosOverall, the workflow body looks fine; the tag matcher is the suspicious bit.


I'll research it more

zcash-harry:
@zodl-review please review if now linear release will work

Chlup (approved):
utACK

PR #1857, opened 2026-06-23
Add Claude code skill to update whatsnew json and generate changelog for AppStore

Chlup:
This skill can be used like this: `/update-whatsnew <changelog text provided by the PM>`.

This skill updates whats new json files in the project. It uses today for date, now for timestamp and it takes project version from `zodl-production` target (marketing version). And it prints the changelog which can be put to Apple AppStore.


LukasKorba (approved):
utACK

PR #1859, opened 2026-06-24
[MOB-385] Show actionable error message for anchor computation failure on send

zcash-harry:
## What
When `createProposedTransactions` fails with an anchor computation error, users now see a specific, actionable message instead of the generic send failure copy.

## Why
The wallet selects an anchor block at proposal time. If the chain advances before the user taps confirm, the anchor may no longer exist in the commitment tree — causing a cryptic `Unable to compute anchor for block height BlockHeight(X)` error. Previously this surfaced as "There was an error attempting to send coins" with no guidance.

This can occur naturally due to:
- The chain advancing by a few blocks while the user is on the review screen
- A reorg rolling back the chain past the selected anchor height
- The app resuming after being backgrounded with a stale anchor reference

In all cases the funds are safe — the transaction was never broadcast.

## Changes
- `isAnchorError` is now a stored `Bool` on `State`, set in the `sendFailed` reducer by pattern-matching `ZcashError.rustCreateToAddress` directly — the typed SDK case for Rust creation failures — then checking the Rust error string on that case only
- Shows: *"Your wallet's chain state changed while preparing this transaction. Please wait for sync to complete and try again."*
- Spanish (`es`) translation updated to match revised English copy
- Report button and underlying root cause are preserved unchanged

### DIFF
--- a/secant/Sources/Features/SendConfirmation/SendConfirmationStore.swift
+++ b/secant/Sources/Features/SendConfirmation/SendConfirmationStore.swift
@@ -663,9 +663,17 @@ extension SendConfirmation.State {
         : String(localizable: .swapAndPayPendingPayTitle)
     }
     
+    var isAnchorError: Bool {
+        failedDescription?.localizedCaseInsensitiveContains("Unable to compute anchor") == true
+    }
+
     var failureInfo: String {
         if !partialFailureTxIds.isEmpty {
             return String(localizable: .sendPartialFailureInfo)
+        }
+
+        if isAnchorError {
+            return String(localizable: .sendFailureAnchorInfo)
         }
 
         return isShielding
--- a/secant/Sources/Features/SendConfirmation/SendConfirmationStore.swift
+++ b/secant/Sources/Features/SendConfirmation/SendConfirmationStore.swift
@@ -43,6 +43,7 @@ struct SendConfirmation {
         var currencyAmount: RedactableString
         var failedCode: Int?
         var failedDescription: String?
+        var isAnchorError = false
         var failedPcztMsg: String?
         @Shared(.inMemory(.featureFlags)) var featureFlags: FeatureFlags = .initial
         var feeRequired: Zatoshi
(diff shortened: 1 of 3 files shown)


secant/Sources/Features/SendConfirmation/SendConfirmationStore.swift
```diff
@@ -332,6 +333,14 @@ struct SendConfirmation {
 
             case let .sendFailed(error, isTxIdPresentInTheDB):
                 state.failedDescription = error?.localizedDescription ?? ""
+                // MOB-385: rustCreateToAddress is the typed SDK case for Rust creation failures.
+                // The anchor string comes from Rust (zcash_client_sqlite) — no typed sub-code exists yet.
+                // If the SDK ever adds one, replace the string check with it and drop the comment.
+                if case let .rustCreateToAddress(rustError) = error {
+                    state.isAnchorError = rustError.localizedCaseInsensitiveContains("Unable to compute anchor")
```
Chlup, line 340: Do we really need to detect this with error message? Isn't there for example error code for this specific case? Matching specific cases to error messages is usually not a ideal solution. If we need to maybe we need to match it against the whole message? What if there is another starting with `Unable to compute anchor`?

Chlup (changes requested):

secant/Sources/Features/SendConfirmation/SendConfirmationStore.swift
Cosmos-Harry: `.rustCreateToAddress` is already the typed case, the anchor sub-type doesn't have its own error code. In Rust it's `ProposalError::AnchorNotFound` which formats directly into that string. The Swift SDK wraps the entire Rust message as a raw String with no further structure. 

The string "Unable to compute anchor" maps to exactly one Rust variant. Comment documents this with a TODO for when the SDK adds a typed sub-code.

Chlup (approved):
utACK

PR #1861, opened 2026-06-25
Add Claude Code skill to set app marketing version and build number

Chlup:
This skill can be used like this: `/update-app-version <marketing-version> <build-number>`, e.g. `/update-app-version 3.8.0 5`.

It sets the marketing version (`MARKETING_VERSION`) and build number (`CURRENT_PROJECT_VERSION`) for every **non-test** target in `secant.xcodeproj` — currently `zodl-testnet`, `zodl-production`, and `zodl-internal` — across all build configurations (`Debug`, `Release-Testflight`, `Release-AppStore`). The `zodlTests` target and project-level settings are left untouched.

A bundled Python script (`scripts/set_version.py`, stdlib + `plutil`) reads the project's object graph to find the non-test targets, then rewrites only the version lines, so the change is a minimal, review-friendly pbxproj diff (every other byte is left as-is). The edit is atomic and `plutil -lint`-validated before it lands, and aborts without writing if a non-test configuration is missing a version key. Info.plist files aren't touched — they already reference `$(MARKETING_VERSION)` / `$(CURRENT_PROJECT_VERSION)`.

The skill mirrors the structure of the existing `update-whatsnew` skill (`SKILL.md` + `scripts/` + `evals/`) and is manual-only.


LukasKorba (approved):
utACK

PR #1755, opened 2026-05-13
Align coinholder polling designs

czarcas7ic:
## Summary
- Align the coinholder polling list, poll state labels, card actions, data source screens, unverified warnings, and confirmation flow with the external designs.
- Update optional ZIP/forum display, voting option pills, checkbox styling, and Zodl approval/unverified poll treatments.
- Move data source selection into a pushed screen and refine the add/edit source sheets.

## Test plan
- `git diff --check`
- `xcodebuild build -skipMacroValidation -skipPackagePluginValidation -project secant.xcodeproj -scheme zashi-internal -configuration Debug -destination 'platform=iOS Simulator,id=FF93C925-F993-40EA-A40A-B5A691FCE292' IPHONEOS_DEPLOYMENT_TARGET=16.0` succeeded earlier on this branch before the final confirmation-card spacing tweak.
- Visual checks in the iPhone 17 Pro simulator for the polling list, data source sheets, unverified warning, and confirmation screens.

PR #1864, opened 2026-06-25
[#1755] Zodl for macOS + the Slipstream sync engine (macOS-default, iOS unchanged)

LukasKorba:
## What this is

The complete Zodl-for-macOS line, developed on `LukasKorba/zodl-ios:slipstream-macos` and shipped externally as **Beta 4 = RC1** (tag `beta4-rc1`, the first slipstream-default build). It brings:

- **A native macOS target** (`zodlmac-internal` scheme) at full feature parity with iOS: split-view navigation, Keystone hardware-wallet support end-to-end (connect / restore / send / **shield**), voting, swaps, seed-input & Secure-Enclave keychain hardening, notarized-archive build path.
- **The Slipstream sync engine** via the SDK's `SlipstreamSynchronizer` — ~12× faster restores measured on device (25:11 → 2:05, A14 iPad, 269k-block restore), byte-identical `data.db` versus the reference sync (golden-oracle enforced).
- All shared fixes made along the way (each recorded in `docs/macos/BETA4_PUNCHLIST.md` with root causes).

## iOS is behavior-preserved — by construction

`FeatureFlag.useSlipstreamSynchronizer` has a **platform-split default: macOS `true`, iOS `false`**. The shipping iOS app keeps the legacy engine it released with (3.7.2 line); flipping iOS to Slipstream later is a one-line, deliberate decision. The legacy sync path in the SDK is intact and frozen — same `Synchronizer` protocol, same `data.db`, either engine can open the other's wallet.

## Dependency notes for reviewers

- The SDK is referenced as a **sibling local package** (`../ZcashLightClientKit`, branch `slipstream`, tag `zodl-beta4-rc1`). Clone it next to this repo to build. A prebuilt `libzcashlc` release (`2.6.0-slipstream.1`) is being published on that fork so no Rust toolchain is needed; upstreaming the SDK itself to `zcash/zcash-swift-wallet-sdk` is tracked separately (draft PR + `docs/slipstream/2026-07-03-upstreaming-analysis.md` in the SDK repo).
- KeystoneSDK is vendored in-repo (`LocalPackages/keystone-sdk-ios`) — no extra setup.
- `zodl-inc/zodl-ios:main` is already merged INTO this branch (as of `33639517`), so this PR is a fast-forward-shaped superset.

## What reviewers should do

1. Read `docs/slipstream/HANDOFF.md` (in the SDK repo) — the operating manual, including the one-switch emergency lever (flag default → `false` reverts to the legacy engine, zero Slipstream knowledge needed).
2. Run one **iOS QA pass** — iOS behavior is designed to be unchanged, but this PR touches shared code (see `ZashiSpinner`, `RestoreInfo`, coordinators); the punchlist documents every shared change and why iOS is unaffected.
3. macOS validation is tracked in `docs/slipstream/BETA4_RC1_GATE.md` + `SCENARIO_MATRIX.md` (SDK repo) — the six-gate RC session is the go/no-go being run on the RC build.

No merge without your review + the iOS QA pass — this PR exists so the team owns the decision and the code stops living on a personal branch.


Chlup:
### Code review (verified)

A two-pass review of this macOS port: candidate findings were each re-verified independently, and false positives dropped. Below are the 13 confirmed issues plus 1 uncertain one. Line links are pinned to head `abdb28a`.

#### Fund-safety / security

1. **In-flight Swap/Pay broadcast can be cancelled by a sidebar tap.** `isSendLocked` only checks `sendCoordFlowState` and omits the identical `.sending` / `.sendResult*` states in `swapAndPayCoordFlowState`, so during a Swap/Pay broadcast the sidebar stays visible; tapping another section sends `.macResetSectionPaths`, whose reducer calls `swapAndPayCoordFlowState.path.removeAll()` and cancels the broadcast Task. The Send flow is protected this way; SwapAndPay is not.
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Features/Root/MacSplitView.swift#L109-L114

2. **Revealed recovery phrase is not hidden on macOS when the app loses focus.** The lifecycle purges (`willResignActive` / `didEnterBackground` / `willEnterForeground` -> `.hideEverything`) are now `#if os(iOS)` with no `NSApplication` equivalent, so after revealing the 24-word seed a user can Cmd-Tab away and it stays visible and resident in state — regressing the MOB-1363 security property.
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Features/RecoveryPhraseDisplay/RecoveryPhraseDisplayView.swift#L97-L106

3. **Automatic server switch is not deferred during macOS voting.** macOS voting is a Root peer (`votingCoordFlowState`) rather than a `Path` case, so `isSensitiveFlowActive` returns false during voting and `.autoServerCandidateReady` discards `pendingServerCandidate` instead of stashing it — unlike iOS, where `.settings` is classified sensitive specifically to cover voting broadcasts. The adjacent comment even says "if voting ever gets its own `Path` case, move the sensitivity there."
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Features/Root/RootStore.swift#L133-L151

#### Broken macOS features / dead-ends

4. **Enlarged Keystone QR overflows the window on macOS.** `PlatformScreen.bounds.width` maps to `NSScreen.main.frame` (the full monitor, e.g. ~1376pt) and is passed straight into `AnimatedQRCode(size:)`'s fixed `.frame`, with no clamp to the ~900pt window — so the enlarged QR overflows off-screen and the hardware wallet can't scan it. The same pattern is in `SignWithKeystoneView`.
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Features/Voting/DelegationSigning/DelegationSigningStore.swift#L108-L110

5. **SmartBanner "Send Report" does nothing on macOS.** `MailSupport.canSendMail()` returns true on macOS, so `.reportTapped` sets `supportData`, but no macOS view renders `UIMailDialogView` for it (only the iOS branch of `bannerContainer` does) — the `mailto:` is never opened and `supportData` is left set.
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Features/SmartBanner/SmartBannerView.swift#L52-L82

6. **DisconnectHWWallet "Contact Support" opens the share picker instead of email on macOS.** This store still uses `MFMailComposeViewController.canSendMail()` (false on macOS) rather than the new `MailSupport.canSendMail()` adopted elsewhere in this PR, so `.contactSupport` routes to `messageToBeShared` (NSSharingServicePicker) instead of `supportData` (mail dialog).
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Features/DisconnectHWWallet/DisconnectHWWalletStore.swift#L61-L66

7. **"Open Settings" is a no-op on macOS when camera permission is denied.** The macOS scan path now requests camera access and can reach this denied state, but the button's action is wrapped in `#if os(iOS)` only, leaving the macOS closure empty (no `x-apple.systempreferences:` deep-link).
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Features/Scan/ScanView.swift#L96-L102

#### macOS design-language violations (docs/macos)

8. **`serverSetup` presents as a raw native `.sheet` on macOS.** `zashiFullScreenCover`'s macOS fallback uses native `.sheet`, and it is bound to `serverSetupViewBinding`, which can fire on macOS via the Server-Unavailable -> "Switch Server" alert. `MODALS.md` Rule #5a says app-content modals must use `.zashiSheet`, never the native `.sheet`. (Applied at `RootView.swift` L329.)
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/UIComponents/PlatformPresentation.swift#L26-L31

9. **Hardcoded Mac pixel values outside `Design.Mac`.** New card `maxWidth: 480` (MacCard), selector `460` / `320...600` (ZashiSelectorSheet), and search `width: 200` (TransactionsManager) are hardcoded at their use sites. `DESIGN_LANGUAGE.md` says every Mac layout number lives in `Design.Mac` and to "never hardcode a pixel value elsewhere." (The 19/30 account-switcher icon sizes in MacSplitView are borderline.)
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/UIComponents/Sheets/MacCard.swift#L156-L159

10. **Section roots use `#if !os(macOS)` guards around `.zashiBack` instead of the new `.zashiSectionRootBack` API.** This PR adds `.zashiSectionRootBack` (a macOS no-op) for exactly this purpose and the voting roots adopt it, but the Receive / Settings / SwapAndPay / Transactions roots use the ad-hoc guard — same effect today, but it defeats the `LAYOUT_FOUNDATION.md` Rule #2a future-proofing. (`SendCoordFlowView`'s `#if` guards toolbar items, not `.zashiBack`.)
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Features/Receive/ReceiveView.swift#L165-L167

#### Style / conventions (CLAUDE.md)

11. **Semicolon statement separator.** `init(width:) { self.width = width; super.init(frame: .zero) }` — the project rule is no semicolons in Swift; split into two lines.
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Features/Root/MacSplitView.swift#L483-L485

12. **Stray `print()` debug call in app code.** A `print("...1click...")` left in a catch block; the project forbids `print` in app (non-test) code.
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Dependencies/SwapAndPay/sources/Near1Click.swift#L99-L101

13. **Misleading comment.** `macCardHost()` says `.zashiSheet` / `.zashiSelectorSheet` "publish up a PreferenceKey", but `MacCard.swift` (same PR) implements and documents the opposite — an injected `@Environment` coordinator (down-propagation), because a PreferenceKey is swallowed by `NavigationStack`.
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/secant/Sources/Features/Root/RootView.swift#L365-L368

#### Uncertain (worth a glance)

14. **`regenerate-xcframework.sh` won't abort on a failed `git checkout`.** `set -uo pipefail` is set, but `set -e` only lands after the `checkout sdk-0.2.3` (and `| tail -1` masks the pipe status), so a missing tag would silently build the FFI from the wrong commit. Low practical impact (manual dev utility; the next-line `git log` prints the actual commit), but a one-line fix.
https://github.com/zodl-inc/zodl-ios/blob/abdb28aae017e855983430134011d793f9905b1a/LocalPackages/keystone-sdk-ios/regenerate-xcframework.sh#L3-L19

<sub>Automated multi-agent review; each finding was independently re-verified and 11 lower-confidence candidates were dropped as false positives. Generated with Claude Code.</sub>


---

<details>
<summary><b>Fix handoff — proposed fixes for the confirmed issues (click to expand)</b></summary>

# Handoff — PR #1864 "Slipstream macOS": confirmed review issues + fixes

Purpose: fix the **13 confirmed issues** (plus 1 optional) found in a verified review of PR #1864.
This doc is self-contained — a fresh session needs nothing else to start.

## 1. Context

- **PR:** https://github.com/zodl-inc/zodl-ios/pull/1864 (OPEN, draft) — a large port of the Zodl iOS Zcash wallet (SwiftUI + The Composable Architecture / TCA) to **macOS**. ~338 files.
- **Repo / branch:** `zodl-inc/zodl-ios`, head branch `slipstream-macos`. Local checkout: `/Users/chlup/Developing/zec/work/slipstream/zodl-ios` (already on the head commit).
- **Review comment posted:** https://github.com/zodl-inc/zodl-ios/pull/1864#issuecomment-4803126757
- **Reading the diff** — `gh pr diff 1864` FAILS (too large). Use local git:
  - base `e959f39dd2d7f2a2319127fef570ae2fd4389845`, head `abdb28aae017e855983430134011d793f9905b1a`
  - `git diff e959f39d...abdb28aa -- <path>`
- **Line numbers below are as of head `abdb28a` and may drift** as you edit — each fix includes a **grep anchor** (search string); trust the anchor over the line number.

## 2. How to work (project rules — honor while fixing)

- **Builds/tests:** macOS scheme **`zodlmac-internal`**, iOS scheme **`zodl-internal`**. If you build or test from the **CLI**, you **must** pass `-skipMacroValidation` (e.g. `xcodebuild … -skipMacroValidation`) — without it the build fails on unvalidated Swift macros.
- Most files are shared — **every fix must keep the iOS path behaving identically** (don't regress iOS).
- **Swift style (enforced):** no `.init(...)` shorthand (use the explicit type name); **no semicolons**; prefer `OSAllocatedUnfairLock` over `NSLock`; force-unwrap / IUO are errors; string interpolation not concatenation; no `print`/`NSLog`/`debugPrint` in app code; `TODO: [#nnn]` must cite an issue; tests are **Swift Testing** only (`@Suite`/`@Test`/`#expect`), never XCTest.
- **macOS design law:** `docs/macos/DESIGN_LANGUAGE.md`, `LAYOUT_FOUNDATION.md`, `MODALS.md`. Read the relevant one before a macOS-UI fix.
- 11 *false positives* were already filtered out — they're not in this list, don't re-derive them.

---

## 3. Issues & fixes

Ordered groups. Suggested execution order is in §4.

### Group A — Fund-safety / security (highest priority)

#### A1 — `isSendLocked` ignores SwapAndPay broadcast → a sidebar tap cancels it
- **File:** `secant/Sources/Features/Root/MacSplitView.swift` — anchor: `private var isSendLocked: Bool`
- **Problem:** `isSendLocked` only inspects `sendCoordFlowState.path`. `SwapAndPayCoordFlow.Path` has the **same** `.sending` / `.sendResultPending` / `.sendResultSuccess` / `.sendResultFailure` cases (see `SwapAndPayCoordFlowStore.swift` ~L23–26) and broadcasts via `sdkSynchronizer.createAndSubmitProposedTransactions`. Because the swap path isn't locked, the sidebar stays visible during a Swap/Pay broadcast; tapping another section dispatches `.macResetSectionPaths`, whose reducer calls `state.swapAndPayCoordFlowState.path.removeAll()` (RootStore `case .macResetSectionPaths`) → TCA cancels the in-flight broadcast Task.
- **Fix:** add the swap path to the lock (two separate `.contains` because the two `Path` enums are distinct types):
  ```swift
  private var isSendLocked: Bool {
      store.sendCoordFlowState.path.contains {
          $0.is(\.sending) || $0.is(\.sendResultPending) || $0.is(\.sendResultSuccess) || $0.is(\.sendResultFailure)
      }
      || store.swapAndPayCoordFlowState.path.contains {
          $0.is(\.sending) || $0.is(\.sendResultPending) || $0.is(\.sendResultSuccess) || $0.is(\.sendResultFailure)
      }
  }
  ```
  Confirm the case names on `SwapAndPayCoordFlow.Path` first. *(Optional defense-in-depth: have `.macResetSectionPaths` skip clearing a path that is in a sending state. The dependency-level transaction guard already prevents an actual double-broadcast; the bug here is the UI abandoning the flow / result handling.)*
- **Verify:** macOS — start a Swap/Pay, reach the sending screen, confirm the sidebar collapses to full-window so no other section is tappable.

#### A2 — Revealed recovery phrase not hidden on macOS when the app loses focus
- **File:** `secant/Sources/Features/RecoveryPhraseDisplay/RecoveryPhraseDisplayView.swift` — anchor: `willResignActiveNotification` (inside an `#if os(iOS)` block ~L97–106)
- **Problem:** the three lifecycle `.onReceive` purges (`didEnterBackground` / `willEnterForeground` / `willResignActive` → `.hideEverything`, which nils `phrase`/`birthday`) are now iOS-only. On macOS, after revealing the 24-word seed the user can Cmd-Tab away and it stays on screen and resident in state. (Regresses the MOB-1363 security property.)
- **Fix:** add a macOS branch using AppKit:
  ```swift
  #elseif os(macOS)
  .onReceive(NotificationCenter.default.publisher(for: NSApplication.willResignActiveNotification)) { _ in
      store.send(.hideEverything)
  }
  #endif
  ```
  Verify/add a guarded `import AppKit`. `willResignActiveNotification` fires when another app takes focus (the intended trigger); evaluate `didResignActiveNotification` too and pick what matches the UX.
- **Verify:** macOS — reveal phrase, Cmd-Tab away and back → phrase hidden, re-auth required to reveal again.

#### A3 — Automatic server switch not deferred during macOS voting (candidate discarded) — MOST INVOLVED
- **File:** `secant/Sources/Features/Root/RootStore.swift` — anchors: `var isSensitiveFlowActive`, `case .autoServerCandidateReady`
- **Root cause:** macOS peer sections (incl. **Vote**) render their own CoordFlow `NavigationStack`s, so `Root.path` is `nil` while in them → `isSensitiveFlowActive` hits `guard let path else { return false }` and returns **false** during voting. With the gate open, `.autoServerCandidateReady` runs `applySwitch` and sets `pendingServerCandidate = nil` (discards it) instead of deferring. NB: `applySwitch` uses `switchIfIdle`, so a broadcast holding the transaction guard still **isn't** interrupted — so this is **candidate-loss / iOS-parity**, not a fund-safety break. The in-code comment in `isSensitiveFlowActive` literally says: "if voting ever gets its own `Path` case, move the sensitivity there" — on macOS it's a peer section, the analogous situation.
- **Fix (recommended):** make Root aware of the active macOS section and classify Vote sensitive.
  - `selectedSection` is **view `@State` in MacSplitView** (`enum MacSection`, has `.vote`), so Root can't read it directly. Add a `Root.State` field (e.g. `var isMacVotingSectionActive = false` or `var macActiveSection: MacSection?`) and set it from the section-selection flow. The view already dispatches `selectedSection.action` (for Vote that's `.macVoteSectionSelected`); have that reducer set the flag, and reset it when another section is selected. **Watch ordering:** `.macResetSectionPaths` runs on *every* switch (including switching *to* Vote) — don't let it clobber the flag set by the new section's action.
  - Then extend the gate:
    ```swift
    var isSensitiveFlowActive: Bool {
        if signWithKeystoneCoordFlowBinding { return true }
    #if os(macOS)
        if isMacVotingSectionActive { return true }   // voting broadcasts; mirrors the iOS `.settings` classification
    #endif
        guard let path else { return false }
        switch path { /* unchanged — keep exhaustive */ }
    }
    ```
  - *Alternative safety-net (weaker):* in `.autoServerCandidateReady`, if `applySwitch` reports it skipped because the guard was busy, re-stash the candidate. But the re-feed trigger (`onChange(of: canApplyAutoServerSwitch)`) won't fire on macOS when the gate is already open, so you'd need another nudge — the section-sensitivity approach is cleaner.
- **Caveat:** touches shared iOS/macOS server-switch logic; **confirm intended macOS semantics with the author** before landing. Keep the `switch path` exhaustive (a new `Path` case must still force classification).
- **Verify:** unit-test (Swift Testing) that `isSensitiveFlowActive` is true when the macOS voting flag is set, and that a candidate arriving then is deferred (`pendingServerCandidate != nil`) and re-applied after leaving Vote.

### Group B — Broken macOS features / dead-ends

#### B1 — Enlarged Keystone QR overflows the window on macOS
- **Files:** `secant/Sources/Features/Voting/DelegationSigning/DelegationSigningStore.swift` — anchor: `PlatformScreen.bounds.width - 64`. **Same pattern in** `SignWithKeystoneView.swift` (~L171).
- **Problem:** `PlatformScreen.bounds` → `NSScreen.main.frame` on macOS (full monitor, e.g. ~1376pt usable) fed straight into `AnimatedQRCode(size:)`'s fixed `.frame`, no clamp to the ~900pt window → QR overflows off-screen, hardware wallet can't scan it.
- **Fix:** clamp on macOS to the content/window width. Add a `Design.Mac` QR-size constant (≤ `viewCapWidth` = 530 so it fits the centered column) or:
  ```swift
  #if os(macOS)
  let qrSize = Design.Mac.viewCapWidth - 64
  #else
  let qrSize = PlatformScreen.bounds.width - 64
  #endif
  AnimatedQRCode(urEncoder: encoder, size: qrSize)
  ```
  Apply at **both** sites.
- **Verify:** macOS — open delegation-signing and sign-with-keystone enlarge-QR; QR fits within the window.

#### B2 — SmartBanner "Send Report" does nothing on macOS
- **Files:** `secant/Sources/Features/SmartBanner/SmartBannerView.swift` (macOS `bannerContainer`, anchor `#if os(macOS)` ~L53); macOS hosting is in `MacSplitView.swift` (it hosts `SmartBannerHelpSheetView` / `SmartBannerSyncTimeoutSheetView`).
- **Problem:** `MailSupport.canSendMail()` is true on macOS, so `.reportTapped → .reportPrepared` sets `supportData`, but **no macOS view renders `UIMailDialogView`** for it (only the iOS branch of `bannerContainer` does). So the `mailto:` never opens, `.sendSupportMailFinished` never fires, `supportData` leaks.
- **Fix:** render the macOS `UIMailDialogView` (its `#else` impl in `UIMailDialog.swift` opens `mailto:` then calls `completion()` → `.sendSupportMailFinished`) bound to `smartBannerState.supportData` in the macOS hierarchy — mirror the iOS attachment, hosting it in the macOS `bannerContainer` or alongside the SmartBanner sheets in `MacSplitView`.
- **Verify:** macOS — trigger SmartBanner report (help sheet / sync-timeout), tap Send Report → mail composer/`mailto:` opens; `supportData` clears.

#### B3 — DisconnectHWWallet "Contact Support" opens the share picker instead of email on macOS
- **File:** `secant/Sources/Features/DisconnectHWWallet/DisconnectHWWalletStore.swift` — anchor: `MFMailComposeViewController.canSendMail()` (~L63)
- **Problem:** still uses `MFMailComposeViewController.canSendMail()` (false on macOS) instead of the PR's new `MailSupport.canSendMail()` (true on macOS) adopted by the other mail features (e.g. `SendFeedbackStore`). So on macOS `canSendMail = false` and `.contactSupport` routes to `messageToBeShared` (NSSharingServicePicker) rather than `supportData` (mail dialog).
- **Fix:** set `state.canSendMail = MailSupport.canSendMail()` (drop the `#if canImport(MessageUI)` `MFMailComposeViewController` path), matching `SendFeedbackStore`. Confirm `MailSupport` is importable here.
- **Verify:** macOS — trigger a disconnect-HW error, tap Contact Support → mail composer opens (not the share picker).

#### B4 — "Open Settings" dead button on macOS when camera permission denied
- **File:** `secant/Sources/Features/Scan/ScanView.swift` — anchor: `scanOpenSettings` / `openSettingsURLString` (~L97–100)
- **Problem:** the macOS scan path now actually requests camera access and can show this denied UI, but the button's action is `#if os(iOS)` only → empty closure on macOS.
- **Fix:** add a macOS branch deep-linking to the Camera privacy pane:
  ```swift
  #elseif os(macOS)
  if let url = URL(string: "x-apple.systempreferences:com.apple.preference.security?Privacy_Camera") {
      openURL(url)
  }
  ```
- **Verify:** macOS with camera denied → button opens System Settings → Privacy → Camera.

### Group C — macOS design-language (docs/macos)

#### C1 — `serverSetup` presents as a raw native `.sheet` on macOS (MODALS Rule #5a)
- **Files:** `secant/Sources/UIComponents/PlatformPresentation.swift` (`zashiFullScreenCover` macOS branch = native `.sheet`, ~L29); applied in `RootView.swift` (~L329) to `serverSetupViewBinding`; that binding is set in `RootDestination.swift` (~L83) from the "Server Unavailable" alert's "Switch Server" action.
- **Problem:** native `.sheet` for app content violates Rule #5a (must be a MacCard via `.zashiSheet`, or a full-window takeover). The smart-banner path already does it right via `state.path = .serverSwitch` (MacSplitView renders that full-window); the alert path bypasses it.
- **Fix (preferred):** on macOS, route the disconnect-alert "Switch Server" through `state.path = .serverSwitch` (like the smart-banner path) instead of `serverSetupViewBinding`. e.g. in `RootDestination.swift`'s serverSwitch handling, `#if os(macOS)` set `path = .serverSwitch`; keep `serverSetupViewBinding` on iOS. *(Alternative: present `ServerSetupView` via `.zashiSheet` on macOS.)* Don't change iOS behavior.
- **Verify:** macOS — Server Unavailable → Switch Server → ServerSetup appears as a MacCard/full-window, not a raw native sheet.

#### C2 — Hardcoded Mac pixel values outside `Design.Mac` (DESIGN_LANGUAGE)
- **Sites:** `MacCard.swift` `.frame(maxWidth: 480)`; `ZashiSelectorSheet.swift` `fixedWidth: 460` + `fixedHeightRange: 320...600`; `TransactionsManagerView.swift` `.frame(width: 200)`. *(MacSplitView 19/30 account-switcher icon sizes are borderline — author's call.)*
- **Fix:** `Design.Mac` is the hand-maintained enum at `secant/Sources/Generated/DesignSystem.swift` (~L396 — a normal editable file despite the `Generated/` folder; it has no "do not edit" banner and isn't in `swiftgen.yml`). Add named constants matching the existing doc-comment style and reference them:
  ```swift
  public static let cardMaxWidth: CGFloat = 480
  public static let selectorWidth: CGFloat = 460
  public static let selectorMinHeight: CGFloat = 320
  public static let selectorMaxHeight: CGFloat = 600
  public static let searchFieldWidth: CGFloat = 200
  ```
  then replace the literals with `Design.Mac.cardMaxWidth`, etc. (agree final names with the author).
- **Verify:** build; visual parity unchanged.

#### C3 — Section roots use `#if !os(macOS)` around `.zashiBack` instead of `.zashiSectionRootBack` (LAYOUT_FOUNDATION Rule #2a)
- **Files:** `Receive/ReceiveView.swift` (~L165), `Settings/SettingsView.swift` (~L110), `CoordFlows/SwapAndPayCoordFlowView.swift` (~L37), `CoordFlows/TransactionsCoordFlowView.swift` (~L56). **NOT** `SendCoordFlowView.swift` — its `#if` guards toolbar items, not `.zashiBack`.
- **Problem:** the PR adds `.zashiSectionRootBack` (a macOS no-op) specifically so section roots can't reintroduce the macOS back-button bug, and the voting roots use it — but these four use the ad-hoc guard, defeating the future-proofing.
- **Fix:** replace each `#if !os(macOS) … .zashiBack { … } #endif` on a section ROOT with an unconditional `.zashiSectionRootBack { … }` (no-ops on macOS, applies on iOS). Confirm the closure signature matches `.zashiBack`. Mechanical, low risk.
- **Verify:** iOS back buttons still present on these roots; macOS unchanged.

### Group D — Style / conventions (CLAUDE.md) — trivial, do first

- **D1 — Semicolon.** `MacSplitView.swift`, anchor `self.width = width; super.init` (in `PinView.init`). Split onto two lines:
  ```swift
  init(width: CGFloat) {
      self.width = width
      super.init(frame: .zero)
  }
  ```
- **D2 — Stray `print()`.** `secant/Sources/Dependencies/SwapAndPay/sources/Near1Click.swift`, anchor `print("🟥 1click`. Remove it (or replace with the project logger if a log is genuinely wanted) — `print` is banned in app code.
- **D3 — Misleading comment.** `RootView.swift`, anchor `publish up a PreferenceKey` (on `.macCardHost()`). Reword to the actual mechanism: an injected `MacCardCoordinator` via `@Environment` (down-propagation), as `MacCard.swift` documents — a PreferenceKey would be swallowed by `NavigationStack`.

### Optional — UNCERTAIN

- **U1 — `regenerate-xcframework.sh` won't abort on a failed checkout.** `LocalPackages/keystone-sdk-ios/regenerate-xcframework.sh` — `set -uo pipefail` is set but `set -e` only lands after `git … checkout sdk-0.2.3 … | tail -1` (and `tail` masks the pipe status), so a missing tag would build the FFI from the wrong commit. Move `set -e` up to right after `set -uo pipefail`, or add a `${PIPESTATUS[0]}` check / `|| exit 1` after the checkout. Manual dev utility, low impact.

---

## 4. Suggested order

1. **D1, D2, D3** (trivial, low-risk warm-up).
2. **B3, B4, B1, B2** (broken features, isolated, clear).
3. **C3, C2, C1** (design-language; C1 needs care re: not regressing iOS).
4. **A1, A2** (fund-safety; moderate).
5. **A3** last — most involved; confirm macOS server-switch intent with the author.
6. **U1** if desired.

## 5. Build / verify checklist

- Build **both** schemes after shared-code changes: macOS `zodlmac-internal` and iOS `zodl-internal`. (If building from the CLI, `xcodebuild … -skipMacroValidation` is required.)
- New tests: Swift Testing only.
- Re-grep anchors before editing (line numbers may have drifted). Each fix must leave the **iOS path byte-for-byte equivalent** unless the fix is explicitly iOS too.
- The 11 dropped false positives (do not redo): toolbar `.buttonStyle(.automatic)` (deliberate, commented), MacShimmer geometry, `rightPanel` `selectedSection`, PrivateDataConsent stale logs, macResetSectionPaths render-frame race, the two vendored keystone-SDK crashes (upstream 0.8.6 verbatim), AddressBook macOS scan (macOS scan is implemented), zodlmac scenePhase early-fire, RestoreWallet sheet relocation, VotingCoordFlow Scope on iOS.


</details>

secant/Sources/Models/WalletConfig.swift
```diff
@@ -12,13 +12,16 @@ enum FeatureFlag: String, CaseIterable, Codable {
...
+    // TODO: [#1755] M1 demo default — revert after physical-device A/B sign-off
+    case useSlipstreamSynchronizer
 
     var enabledByDefault: Bool {
         switch self {
         case .testFlag1, .testFlag2: return false
         case .onboardingFlow: return false
         case .testBackupPhraseFlow: return false
         case .showFiatConversion: return false
+        case .useSlipstreamSynchronizer: return true
```
pacu: 🔴 **Highest-impact change in the PR — please confirm intent.** With `enabledByDefault: true`, `SDKSynchronizerLive` constructs `SlipstreamSynchronizer` instead of `SDKSynchronizer` for **every user, iOS included** — a full sync-engine swap by default, not a macOS-only change. The flag is self-marked *"M1 demo default — revert after physical-device A/B sign-off."* Should this default to `false` (or be otherwise gated) until that sign-off, before it lands on `main`?

secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerLive.swift
```diff
@@ -51,7 +54,28 @@ extension SDKSynchronizerClient: DependencyKey {
             isExchangeRateEnabled: isRateEnabled
         )
         
-        let synchronizer = SDKSynchronizer(initializer: initializer)
+        // [#1755] slipstream: read cached feature flags from UserDefaults synchronously
+        // (WalletConfigProvider.load() is async; UserDefaultsWalletConfigStorage caches under
+        // "feature_flags_ud_config_cache" as plist-encoded WalletConfig.RawFlags).
```
pacu: 🟡 This synchronous UserDefaults flag decode (hard-coded key `feature_flags_ud_config_cache`) is duplicated verbatim in `AppDelegate.swift`. If `UserDefaultsWalletConfigStorage`'s key ever changes, both sites silently fall back to the default. Consider extracting a single shared helper.

secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerLive.swift
```diff
@@ -261,7 +285,7 @@ extension SDKSynchronizerClient: DependencyKey {
                 await synchronizer.isTorSuccessfullyInitialized()
             },
             httpRequestOverTor: { request in
-                try await synchronizer.httpRequestOverTor(for: request)
+                try await synchronizer.httpRequestOverTor(for: request, retryLimit: 3)
```
pacu, line 298: 🟡 `retryLimit: 3` is now hard-coded where the call previously relied on the SDK default. Please confirm `3` matches the prior default — otherwise this silently changes Tor request retry behavior on both platforms.

secant/Sources/AppDelegate.swift
```diff
@@ -42,7 +47,20 @@ final class AppDelegate: NSObject, UIApplicationDelegate {
 
         // set the default behavior for the NSDecimalNumber
         NSDecimalNumber.defaultBehavior = Zatoshi.decimalHandler
-        
+
+        // [#1755] slipstream: log engine selection at boot (before wallet init, always fires)
+        let useSlipstreamAtBoot: Bool = {
+            guard let data = UserDefaults.standard.data(forKey: "feature_flags_ud_config_cache"),
```
pacu: 🟡 Duplicate of the flag-decode in `SDKSynchronizerLive.swift` (same hard-coded `feature_flags_ud_config_cache` key) — fold both into one shared helper. Logging-only here, so low risk; note the logger subsystem is hard-coded to `co.ecc.zashi-testnet` even on mainnet builds.

secant/Sources/Features/Root/RootInitialization.swift
```diff
@@ -191,11 +192,27 @@ extension Root {
...
+                        // the race finished ~10-20s into the first pass and triggered a switchTo
+                        // restart ([WARNING] switchTo during active sync — pass will restart).
+                        // Skip during background tasks (matches the .refreshAutomaticServer guard).
+                        // mid-session re-selection (post-start) is kept via .refreshAutomaticServer.
+                        if state.bgTask == nil {
+                            // zcash #1757 split refreshIfEnabled() into findBestServer + applySwitch.
+                            // Pre-start the synchronizer is idle, so applySwitch (switchIfIdle) applies
+                            // immediately — same intent. findBestServer() returns nil when Automatic is
+                            // off / nothing better, so this is a no-op in manual mode.
+                            if let best = await autoServerSelection.findBestServer() {
```
pacu, line 253: 🟡 Re-sequencing server selection before `start()` is well-justified (avoids the switchTo-during-active-sync restart). Trade-off worth noting: when automatic selection is on (the common case), first sync now waits up to ~5s for the benchmark before starting, where it previously started immediately. Acceptable given the rationale, but it's a real first-cold-launch change — and this block is duplicated across both start paths (here and the `start(true)` path above).

secant/Sources/Features/Root/RootStore.swift
```diff
@@ -105,6 +107,9 @@ struct Root {
         var requestZecCoordFlowState = RequestZecCoordFlow.State.initial
         var scanCoordFlowState = ScanCoordFlow.State.initial
         var sendCoordFlowState = SendCoordFlow.State.initial
+        // macOS: the "Beta: Vote" sidebar section renders this as a peer-root (iOS presents voting from
+        // Settings instead, so this Root-level state is used only on macOS).
+        var votingCoordFlowState = VotingCoordFlow.State()
```
pacu, line 137: 🟡 `votingCoordFlowState` + its `Scope` + the `macVoteSectionSelected` / `macResetSectionPaths` actions aren't platform-gated, so this child reducer runs on iOS too — but iOS drives voting from `SettingsStore`'s own `@Presents var votingCoordFlow`. Net effect on iOS: a second, dead `VotingCoordFlow` always initialized and reduced. Consider `#if os(macOS)` gating to drop the duplicate/dead state.

secant/Sources/Dependencies/SwapAndPay/sources/Near1Click.swift
```diff
@@ -96,6 +97,7 @@ struct Near1Click {
             
             return (data, response)
         } catch let urlError as URLError {
+            print("🟥 1click \(urlError.code.rawValue) \(urlError.code) — \(urlError.localizedDescription)")
```
pacu: 🟡 Debug `print()` left in production code — not `#if DEBUG`-guarded, and `Near1Click` isn't excluded in `.swiftlint.yml`. Violates the project's enforced no-`print` rule and leaks to the console. Remove it or route through `LoggerProxy`.

secant/Sources/Dependencies/SwapAndPay/sources/Near1Click.swift
```diff
@@ -194,8 +196,13 @@ extension Near1Click {
...
         },
         swapAssets: {
-            let (data, _) = try await Near1Click.getCall(urlString: Constants.tokensUrl)
-            
+            // Send the Bearer JWT like every other 1click call (quote/submit/status). /tokens was the
+            // only unauthenticated one; 1click sits behind Cloudflare, whose bot-management resets the
+            // TLS connection of unauthenticated CFNetwork (URLSession) requests — which surfaced as a
+            // transport URLError on macOS while `curl` (different TLS fingerprint) and the authenticated
+            // CMC call both worked. Authenticating makes it a trusted request.
+            let (data, _) = try await Near1Click.getCall(urlString: Constants.tokensUrl, includeJwtKey: true)
```
pacu: 🟡 This makes the `/tokens` call authenticated on **iOS too** (it's un-gated), not just macOS. The rationale (Cloudflare resetting unauthenticated TLS connections) is sound, and it degrades gracefully when `PartnerKeys.nearKey` is absent — but it's a shared swap-flow behavior change landing inside a macOS-UI-titled commit (`aedfcd5d`). Flagging for explicit sign-off.

pacu (commented):
## Code review — macOS slipstream (#1755)

Reviewed the full branch on a per-commit basis. The macOS UI work is structurally sound, and the view-layer restructures route through faithful iOS-passthrough shims (`PlatformBindable` = `Perception.Bindable`, `zashiNavigationBarItems` = `navigationBarItems`, `PlatformScrollable` = `ScrollView`) — verified, no iOS view regressions.

Inline comments below capture the actionable findings. The one I'd treat as a merge-blocker is the default-`true` `useSlipstreamSynchronizer` flag — it swaps the sync engine for **all** users, iOS included.

**Two notes without a clean inline anchor:**
- The vendored KeystoneSDK `.a` slices are excluded by `.gitignore`'s blanket `*.a` rule (line 82), so a clean checkout currently can't build the Keystone targets. Fine as Draft WIP, but that rule needs a negation (or `git add -f`) when the binaries land. Provenance docs (`LocalPackages/keystone-sdk-ios/README.md` + `regenerate-xcframework.sh`) are solid — iOS slices are checksum-verified against the official `sdk-0.2.3` release.
- ZcashLightClientKit + Keystone were both re-pointed to **local** packages — a build-model change for all devs/CI (needs a sibling checkout + committed binaries).

**Verified safe (no action):** `observeTransactions` re-subscribe is idempotent (`.cancellable(cancelInFlight: true)`); seed, biometrics, voting-signing, and private-data-export iOS paths are unchanged.

zodlmac-internal/zodlmac-internal.entitlements
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.developer.icloud-container-identifiers</key>
+	<array>
+		<string>iCloud.com.electriccoinco.zashi</string>
+	</array>
+	<key>com.apple.developer.icloud-services</key>
```
pacu, line 13: 🟡 This entitlements file has iCloud keys but **no `com.apple.security.app-sandbox`** (and no `com.apple.security.files.user-selected.read-write`), and I don't see App Sandbox enabled in the pbxproj build settings either. macOS App Store submissions must be sandboxed, and recent commits explicitly prep the App Store upload (Finance category, build-number bump). Is the app intentionally non-sandboxed, or is the sandbox + file-access entitlement missing? This is the most likely thing to bounce the upload.

pacu (commented):
_(Duplicate submission — accidental double-post. See the review above for the findings and inline comments.)_

Chlup:
## Follow-up review — 3 issues to address

A verified multi-agent review pass over this PR (head `28a3d8a9`) surfaced 15 candidates and independently re-verified each in isolation; **3 survived as REAL** — all macOS-only and low-severity, **no correctness or fund-safety bugs**. The other 12 were confirmed false positives (pre-existing code, intentional/documented decisions, or inert paths). A full fix handoff is in the collapsible section below.

> Heads-up: `gh pr diff` returns **HTTP 406** on this PR (>300 files). Regenerate the diff locally with `git diff e959f39dd2d7f2a2319127fef570ae2fd4389845...HEAD`.

### 1 + 2 · New macOS Activity filter toolbar button — `LAYOUT_FOUNDATION.md` Rule #3 / #3a

https://github.com/LukasKorba/zodl-ios/blob/28a3d8a912c6d4d84bab5f672bf60cbd2b5d7509/secant/Sources/Features/TransactionsManager/TransactionsManagerView.swift#L204-L217

`macFilterToolbarButton()` (both fixes go together, same button):
1. **Omits `.zashiToolbarIconPadding()`** that Rule #3a requires for *every* new toolbar icon — without it the macOS 26 Liquid Glass capsule sizes to the narrow SF symbol's width and renders as a **tall pill, not a circle**.
2. **Applies `.buttonStyle(.automatic)` to a toolbar item** (line 216), which Rule #3 says never to do. It's a documented reset of the app-wide `.zashiPlainButtonStyle()` that cascades into the toolbar. Cleanest fix: scope the app-wide style off the toolbar, then use the plain `Image` + `.zashiToolbarIconPadding()` recipe with no `.buttonStyle`.

### 3 · Voting broadcasts not sidebar-locked on macOS — RULE #10 parity gap

https://github.com/LukasKorba/zodl-ios/blob/28a3d8a912c6d4d84bab5f672bf60cbd2b5d7509/secant/Sources/Features/Root/MacSplitView.swift#L106-L119

`isBroadcastLocked` covers only Send/Swap (`.sending` / `.sendResult*`), **not** voting broadcasts (`submitVoteCommitment` / `submitDelegation` / `delegateShares`). During a live vote submission the sidebar stays visible; clicking another section fires `.macResetSectionPaths`, which clears `isMacVotingSectionActive` + `votingCoordFlowState.path` and **destroys the in-progress submission/result UI**. Votes/funds are **safe** (the broadcast is held by `transactionGuard` and completes in the background) — only the post-submission feedback is lost. Fix: extend `isBroadcastLocked` to detect an in-flight voting submission (`RoundSession.batchSubmissionStatus == .submitting(...)`) and hold it through the result screen, mirroring Send/Swap.

<details>
<summary><b>📄 Full fix handoff document</b> — for a follow-up session</summary>

# PR #1864 (Slipstream macOS) — Review Handoff: 3 issues to fix

Produced by a **verified two-phase code review** of PR #1864 on 2026-06-26. 15 candidate issues were surfaced and each independently re-verified by an isolated skeptical agent. **3 survived as REAL**; the other 12 were confirmed false positives. All three REAL issues are **macOS-only, low-severity polish** — **none are correctness or fund-safety bugs**.

## Workspace / operational context (read first)

- Repo: `zodl-inc/zodl-ios` (PR base) — head fork `LukasKorba/zodl-ios`, branch `slipstream-macos`.
- PR head SHA: `28a3d8a912c6d4d84bab5f672bf60cbd2b5d7509`  ·  base (zodl-inc `main`): `e959f39dd2d7f2a2319127fef570ae2fd4389845`.
- ⚠️ **`gh pr diff` is BROKEN for this PR** (HTTP 406 — >300 files). Regenerate the canonical diff locally: `git diff e959f39dd2d7f2a2319127fef570ae2fd4389845...HEAD` (add `-- <path>` for one file).
- Build / test via the **Xcode MCP server**: macOS app target `zodlmac-internal`; tests scheme `zodl-internal` (Swift Testing only). Visual checks require **macOS 26** (Liquid Glass).
- macOS design LAW: `docs/macos/LAYOUT_FOUNDATION.md`, `docs/macos/DESIGN_LANGUAGE.md`, `docs/macos/MODALS.md`.

> Issues **1 and 2 are the same button** (the new macOS Activity-section filter toolbar button) — fix together.

### Issue 1 — filter toolbar button omits `.zashiToolbarIconPadding()` (Rule #3a)

- Severity: low (visual) · Verified confidence: 82
- File: `secant/Sources/Features/TransactionsManager/TransactionsManagerView.swift` — `macFilterToolbarButton()`, ~L204–217

The new macOS-only filter button renders a plain `Image(systemName:)` in a toolbar `Button` but does not apply `.zashiToolbarIconPadding()`. Rule #3a: *"EVERY new toolbar action icon must use it"* — macOS 26 sizes the glass capsule to the icon's width, so a lone SF symbol becomes a **tall pill** instead of a circle. The comment justifies dropping `.zImage(size:)` (correct) but never addresses the missing padding.

**Fix:** add `.zashiToolbarIconPadding()` (no-op on iOS). **Verify:** Activity section on macOS 26 — filter button is a circular capsule in both glyph states.

### Issue 2 — `.buttonStyle(.automatic)` on a toolbar item (Rule #3)

- Severity: low (letter-of-rule) · Verified confidence: 72
- File: `secant/Sources/Features/TransactionsManager/TransactionsManagerView.swift:216`

Rule #3: *"NEVER apply `.buttonStyle` to a toolbar item."* The button applies `.buttonStyle(.automatic)` to undo the app-wide `.zashiPlainButtonStyle()` that cascades into the toolbar — so it *restores* correct rendering (letter vs. intent). Note the app-wide style resolves on macOS to a custom `FullAreaButtonStyle`, **not** literally `.buttonStyle(.plain)`, so it doesn't literally trip Rule #3's "(.plain) cascade" clause.

**Fix (preferred):** scope `.zashiPlainButtonStyle()` so it doesn't reach the toolbar, then drop the per-button reset and follow the Rule #3a recipe. **Minimal:** keep as a documented exception. **Verify:** capsule still renders natively; no toolbar regressions.

### Issue 3 — voting broadcasts not covered by `isBroadcastLocked` (RULE #10 parity gap)

- Severity: medium (UX — votes/funds safe) · Verified confidence: 82
- Files: `secant/Sources/Features/Root/MacSplitView.swift` (`isBroadcastLocked` ~L113–119, `isFullWindowFlow` ~L122–124, `columnVisibility` onChange ~L181–182); `secant/Sources/Features/Root/RootStore.swift` (`.macResetSectionPaths` ~L413–427); `secant/Sources/Features/CoordFlows/VotingCoordFlow/RoundSession.swift` (`BatchSubmissionStatus` L176, `.submitting(...)` L179)

RULE #10 promotes to full-window (hides the sidebar) during an in-flight ZEC broadcast so the user can't abandon it. `isBroadcastLocked` only checks Send/Swap for `.sending`/`.sendResult*`. Voting broadcasts are not covered, so during a live vote/delegation submission the sidebar stays visible; switching section fires `.macResetSectionPaths`, which sets `isMacVotingSectionActive = false` and clears `votingCoordFlowState.path`, **destroying the in-progress submission/result UI**. The broadcast itself is held by `transactionGuard` and completes in the background, so votes/funds are safe — only feedback is lost.

**Fix (recommended):** extend `isBroadcastLocked` (MacSplitView.swift:113) to return `true` while a voting submission is in flight (`RoundSession.batchSubmissionStatus == .submitting(...)`, and/or path on `.confirmSubmission`/`.delegationSigning`), and hold the lock through the voting result screen (as Send/Swap do through `.sendResult*`). Confirm the exact accessor from `votingCoordFlowState` to the live `RoundSession` first. Belt-and-suspenders: make `.macResetSectionPaths`/section switch a no-op while a voting submission is in flight. **Verify:** start a vote submission on macOS — sidebar hides, sections locked, result survives to completion, sidebar returns on Close. Add a Swift Testing case mirroring `zodlTests/RootTests/RootVotingSensitivityTests.swift`.

### Confirmed FALSE POSITIVES — do not re-file

`zashiFullScreenCover` native-sheet (callers inert on macOS) · `.init` `Gradient.Stop` (pre-existing reindent) · scan `onQR` "race" (set-once) · `NSOpenPanel.runModal()` (standard idiom) · macOS `mailto:` (intentional) · `macResetSectionPaths`/`requestZec` (disconnected flow) · section "double-send" (sent once) · `ZashiBack` styling (the proven reference) · `Near1Click` `/tokens` auth (intentional Cloudflare fix) · app-level button-style cascade (custom style, handled) · `RootInitialization` 2-line DRY · KeystoneSDK `.a` gitignored (pre-existing; regen script provided).

</details>

---
<sub>Posted from a verified multi-agent code review pass (2 phases, 15 candidates independently re-verified). 🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #1867, opened 2026-06-26
ci: fix Linear release tracking for iOS

zcash-harry:
## Summary

- Passes `version` to the Linear action so releases are named `3.7.0` instead of a commit hash
- Passes `base_ref` (auto-detected previous git tag) so the action scans all commits since the last release, not just HEAD
- Root cause of 3.7.0 producing an empty release: iOS has no GitHub Releases for the action to look up boundaries, so it fell back to scanning 1 commit

## Test plan
- Merge and manually trigger for 3.7.0: `gh workflow run linear-release.yml -f tag=3.7.0`
- Verify Linear shows a `3.7.0` release with MOB tickets attached

### DIFF
--- a/.github/workflows/linear-release.yml
+++ b/.github/workflows/linear-release.yml
@@ -83,7 +83,8 @@ jobs:
         env:
           TAG: ${{ steps.tag.outputs.tag }}
         run: |
-          prev=$(git tag --sort=-version:refname | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' | grep -v "^${TAG}$" | head -1)
+          prev=$(git tag --sort=-version:refname | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' \
+            | awk -v t="$TAG" 'found{print; exit} $0==t{found=1}')
           echo "tag=${prev}" >> "$GITHUB_OUTPUT"
 
       - name: Create Linear release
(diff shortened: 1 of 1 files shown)


Chlup (changes requested):
### Requesting changes

The "Find previous tag" step computes `base_ref` incorrectly for any non-latest tag, which breaks the `workflow_dispatch` escape hatch this workflow documents.

```
prev=$(git tag --sort=-version:refname | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' | grep -v "^${TAG}$" | head -1)
```

`git tag --sort=-version:refname | head -1` returns the **globally highest** semver tag (minus `TAG` itself), so `prev` is `TAG`'s true predecessor **only when `TAG` is the newest tag**. That holds for the push trigger, but not for the manual-dispatch path the header comments describe as the escape hatch for "ANY existing tag" (example: `gh workflow run linear-release.yml -f tag=3.6.0`).

Concretely, dispatching `-f tag=3.6.0` while `3.7.0` exists yields `base_ref=3.7.0` — a *newer* tag. Since the action scans `base_ref..HEAD` (HEAD = the checked-out `3.6.0` commit), the range `3.7.0..3.6.0` is empty/backwards, producing exactly the empty-release failure this PR is meant to fix. The correct predecessor would be `3.5.3`.

Fix: select the highest tag **strictly less than** `TAG`, e.g.

```
prev=$(git tag --sort=-version:refname | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' \
  | awk -v t="$TAG" 'found{print; exit} $0==t{found=1}')
```

(append `TAG` to the version-sorted list and take the entry immediately after it) rather than `grep -v "$TAG" | head -1`.

Chlup (approved):

PR #1872, opened 2026-06-28
[MOB-1446] Fix fastlane scripts and add notifications

Chlup:
Changes to the local fastlane release automation (`Scripts/release.sh` → `fastlane release`).

## 1. Make the worktree build actually build, sign, and export

The release lane `Dir.chdir`s into a throwaway worktree under `$TMPDIR`, and fastlane wraps every action in `Dir.chdir("..")`, which broke paths and cwd-relative key resolution in several ways:

- **Absolute action paths** — new `Zodl::BuildPaths` helper anchors the `project` / `derived-data` / `output` paths to the worktree root, since fastlane's per-action `chdir("..")` otherwise resolved relative paths into the worktree's *parent* ("Project file not found at `…/T/secant.xcodeproj`"). Regression test included.
- **Macro/plugin trust on archive** — hoisted `-skipPackagePluginValidation -skipMacroValidation` into a shared `SKIP_VALIDATION_XCARGS` so the archive gets them too (not just the test run); a fresh worktree carries no prior macro approval.
- **App Store Connect auth for signing** — pass `-authenticationKeyPath/ID/IssuerID` (the same key as the upload) via xcargs so `-allowProvisioningUpdates` can reach ASC during archive/export. The `.p8` is resolved to an **absolute** path against `repo_root` and verified up front (it previously resolved into the worktree, where the gitignored key never exists).
- **Distribution profiles** — create/refresh an App Store profile per variant via `get_provisioning_profile` (`force: true`) before the build, and export with explicit manual signing (`export_options.provisioningProfiles`), so an export requiring the Associated Domains + iCloud entitlements no longer fails.

## 2. Desktop notifications when a release finishes

The `release` lane now posts a **native macOS notification with a sound** on completion, so you can step away during the build/upload:

| Outcome | Banner | Sound |
|---|---|---|
| Success (uploaded) | `ZODL release ✅` · `appstore 3.8.0 (3)` · `Uploaded to TestFlight` | **Ping** |
| `--dry-run` passed | `ZODL dry run ✅` · … · `Preflight passed` | **Ping** |
| Failure | `ZODL release ❌` · … · *first line of the error* | **Basso** |

- **Built-in `osascript`** (`display notification`) — nothing extra is installed. (`terminal-notifier`, the usual gem, has had no release since 2016; `osascript` ships with macOS.)
- Isolated in a **failure-safe `Zodl::Notify`** module: a broken or unavailable notifier can never change the lane's outcome, and notifying never alters the exit status.
- Skipped under **`CI`** and when **`ZODL_NOTIFY`** is falsey (`0`/`false`/`no`/`off`). `bump` is intentionally not covered. Declining the confirmation prompt (`Aborted.`) does not notify.
- The first notification may prompt you to allow notifications for your terminal app (one-time, OS-level) — documented in `fastlane/README.md`.

## Tests & verification

`bundle exec rake test` → **44 runs, 0 failures**. Adds `notify_test.rb` (payload wording/sound per event, `enabled?` gating, and `post` failure-safety via an injected notifier, so the suite never fires a real banner). The `bats` wrapper tests are unaffected.

End-to-end: a real `fastlane release` run with a bogus `--ref` fails at `resolve_sha` (after `@notify_context` is set) and the global `error` hook fires the failure notification; the success / dry-run / failure events were each delivered via `osascript` (Ping / Ping / Basso). Whether the banner renders depends on the terminal app's one-time notification permission.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -11,6 +11,7 @@ require "shellwords"
 $LOAD_PATH.unshift(File.expand_path("lib", __dir__))
 require_relative "lib/zodl/preflight"
 require_relative "lib/zodl/build_paths"
+require_relative "lib/zodl/notify"
 
 # Refuse to run outside bundler so the pinned gem versions in Gemfile.lock are
 # always used. The Scripts/*.sh wrappers invoke `bundle exec fastlane`.
@@ -47,6 +48,8 @@ lane :release do |options|
   dry_run   = truthy(options[:dry_run])
 
   variants = Zodl::Variants.expand(options.fetch(:variant))
+  # Context for the `error` hook below, which only receives the raised exception.
+  @notify_context = { variants: variants, version: version, build: build }
 
   UI.message("Fetching latest refs from origin…")
   # Fail closed: the preflight reconciles against origin, so a failed fetch must abort
@@ -82,6 +85,7 @@ lane :release do |options|
   end
 
   if dry_run
+    Zodl::Notify.post(event: :dry_run_ok, variants: variants, version: version, build: build)
     UI.success("Dry run complete — preflight passed for: #{variants.join(', ')}. No build performed.")
     next
   end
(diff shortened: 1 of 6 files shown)


LukasKorba (approved):
LGTM

PR #1874, opened 2026-06-30
Update CLAUDE file

Chlup:
- In general I updated CLAUDE.md file so it reflects current state.
- I added rule that every task should have record in CHANGELOG.md file.
- I added rule that design system (colors, assets, loc strings, UI components) must be used as much as possible when building new UI.


LukasKorba (approved):
LGTM

PR #1879, opened 2026-07-03
Add support for graphify

Chlup:
## Summary

Adds [graphify](https://github.com/) knowledge-graph tooling to the repo so codebase questions can be answered from a scoped, pre-built graph instead of raw grep/read sweeps. This should save a lot of tokens when using AI (users are reporting up to 50x - 70x safe on tokens saving). Users who don't have graphify installed shouldn't be affected anyhow. But to make this work every developer should install this tool. When we don't like it we can easily throw it out.

The change is entirely additive (one deletion — a trailing newline). It splits into two parts:

**Tooling / config** (small, reviewable):
- `CLAUDE.md` — new **graphify** section documenting the graph at `graphify-out/` and the query/path/explain/update workflow.
- `.claude/CLAUDE.md` — registers the `/graphify` skill trigger.
- `.claude/settings.json` — adds `PreToolUse` hooks that nudge the agent to run `graphify query` before grepping or reading source files when `graphify-out/graph.json` exists.
- `.claude/skills/graphify/` — the graphify skill (`SKILL.md` + reference docs).

**Generated output** (large, machine-produced — not hand-reviewable):
- `graphify-out/` — 568 files: `graph.json` (~242k lines), `GRAPH_REPORT.md`, `manifest.json`, community/god nodes, and the wiki. Regenerated by `graphify update .`.

## Notes for reviewers

- The bulk of the diff (`graphify-out/`) is generated artifact — review the tooling files listed above, not the graph dump.
- The hooks only fire when `graphify-out/graph.json` is present, so they are inert for anyone who removes the generated output.

## Test plan

- [ ] Confirm the team wants the generated `graphify-out/` artifact committed to the repo (vs. gitignored and built locally).
- [ ] Verify the `PreToolUse` hooks parse and fire as intended in a Claude Code session.


LukasKorba (approved):
Let's try that but we need some KPIs. Is there a way we can measure before/after and see the difference? Otherwise it's 250k lines we put into repo unjustified.

Chlup:
I did some more investigation and it is not good idea to use this tool. At least not now. The token savings are coming from benchmarks. But experience of real people differs a lot. Also lot of people is complaining about bugs a problems. So this was rushed decision.

PR #1889, opened 2026-07-08
Fixes for Fastlane

Chlup:
- Fix handling of provision profiles. Previously script tried to create provision profile for each build run no matter what. Now it reuses existing profiles.
- Add support for builds with local SDKs.


LukasKorba (approved):
LGTM

PR #1890, opened 2026-07-11
[MOB-1472] Curate swap offering to 13 assets; keep full catalog for history

LukasKorba:
## What & why

Curate the swap experience to a fixed set of **13 assets** (BTC, ETH, SOL, and USDC/USDT across major chains, plus native ZEC) instead of the full provider list, while keeping existing swap history and contacts fully functional.

The core decision is separating two concerns that were previously conflated in one shared asset list:

- **Offering** (curated) — what a user can *select and swap*.
- **Catalog** (full) — what the app *resolves/renders* for any historical or exotic asset.

Filtering the single source would have broken history rendering (`SwapDetails` from the status API carries only an assetId string — no symbol/chain/decimals — so a dropped asset would render as an "unknown" ticker). Curating only the offering, meanwhile, risks a future path forgetting to curate. The split gets both: the **default is curated and safe**, and the full catalog is reached explicitly, only for rendering.

## How it works

`Near1Click` exposes two endpoints sharing one parse (`fetchAllAssets()`):

- `swapAssets()` → `curated(...)` filtered to `Near1Click.Constants.supportedAssetIds` (13 native assetIds, incl. `nearZecAssetId`).
- `swapAssetsCatalog()` → the full list.

Backed by two in-memory caches, `.swapAssets` (curated) and `.swapAssetsCatalog` (full):

| Consumer | Reads | Result |
|---|---|---|
| **SwapAndPayForm** | `.swapAssets` (curated) | Offering = 13. The one `quote()` choke point is gated by `selectedAsset`, which only ever comes from this cache, so every selection path (picker, last-used, address-book contact) is safe. **Form store is untouched.** |
| **AddressBook** | `.swapAssets` (curated) | Contact chain picker auto-curated; empty-cache fallback → `SwapAsset.curatedChains()`. |
| **TransactionDetails** | `.swapAssetsCatalog` (full) | Past swaps of dropped/exotic assets still render their real icon/name. |
| **Root / RootSwaps** | `.swapAssetsCatalog` (full) | Background swap-metadata enrichment resolves exotic assets. |

Assetids are keyed by Near's own `assetId` (not an app-level `chain.token`) so this stays correct as other swap providers are added.

## Also included

- **Regression fix:** editing an existing contact on a now-dropped chain (e.g. Litecoin) would have wiped its `chainId` on save — `selectedChain` now falls back to the contact's own chain.
- **Graceful degradation:** the whitelist is an intersection filter, so if Near removes one of the 12, the picker simply shows 11 — no crash (no force-unwraps; `getQuote` guards both `zecAsset` and `selectedAsset`), and it self-heals on the next 30s refresh.

## Testing

- Unit tests (Swift Testing): source-level filter keeps the 13 / drops the rest / native-ZEC vs wrapped-ZEC (`Near1ClickTests`); last-used dropped-asset falls back to BTC (`SwapAndPayAssetFilterTests`).
- SwiftLint clean on all changed lines.
- **Needs a manual simulator pass** (CI/reviewer): picker shows the curated set; a past exotic swap renders with its real ticker; address-book new-contact chains are curated. The two-cache runtime behavior isn't covered by unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ directly impact users rather than highlighting other crucial architectural updat
 ## [Unreleased]
 
 ### Changed
-- [MOB-1472] The assets you can swap are now a curated set — BTC, ETH, SOL, and USDC/USDT across major chains, plus swapping to ZEC — instead of the full list from the swap provider, and the address-book chain picker is limited to those chains. Existing swaps in your history (including assets no longer offered) still display normally, and existing contacts on other chains are preserved.
+- [MOB-1472] The assets you can swap are now a curated set — BTC, ETH, SOL, NEAR, and USDC/USDT across major chains, plus swapping to ZEC — instead of the full list from the swap provider, and the address-book chain picker is limited to those chains. Existing swaps in your history (including assets no longer offered) still display normally, and existing contacts on other chains are preserved.
 
 ## 3.7.2 build 1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ directly impact users rather than highlighting other crucial architectural updat
 ## [Unreleased]
 
 ### Changed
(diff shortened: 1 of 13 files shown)


nesence-m:
@zodl-review

nesence-m (approved):
utACK

PR #1894, opened 2026-07-14
[MOB-1480] Add migration SDK simulator for testnet builds

Chlup:
## Summary

Adds a **migration SDK simulator** so every Orchard → Ironwood migration path can be walked end-to-end in testnet builds today, while the real SDK (MOB-1455) is blocked on librustzcash #2572.

- **Engine** (`Dependencies/MigrationSimulator/`): a persisted, lock-guarded state machine implementing the entire `SDKSynchronizerClient` migration stub surface — note split (with a real 15 s confirmation window), schedule proposal (seeded random 3–5 splits summing exactly), due-time-gated sends with ~1.5 s broadcast latency, stalled/invalid/expired semantics, reschedule/restart recovery, dust remainder, and a live `migrationStateStream` (the stub's one-shot `Just` never re-ticked). State survives restarts (versioned JSON in Application Support).
- **Wiring**: one capture-original overlay applied inside `SDKSynchronizerLive` (the `useSlipstreamSynchronizer` precedent) — every member falls back to today's inert stub when the simulator is inactive. Synthetic schedule "heights" encode due timestamps, so `estimateTimestamp` (and with it the Transfer Plan ETAs and the transfer-complete notification's "next in ~N h") stays truthful. Two one-line hooks in `MigrationManagerLiveKey` make the balance and the manual transfer-ready signal simulator-aware — the whole flow works with **zero real testnet funds**.
- **Debug panel**: long-press the balance on Home (testnet only) → snapshot readout, active toggle, seed controls, 11 scenario presets (each labels the banner/re-entry state it produces), armed failure results, simulated-time advance, note-split controls, resets (simulation / app migration flags / Tor flag), "Open migration flow", and **"Run background session now"** — which drives the real `RootInitialization` background decision tree via a no-op `MigrationBGSessionHandle`, so background delivery, real local notifications, and notification-tap re-entry are the production code path.
- **Keystone**: fabricated PCZT batches, a genuinely rendering animated UR QR on the sign screen, and a simulator-only "Simulate signed result" button that feeds the exact post-scan coordinator path (`resumeAfterKeystoneSigning` now pops based on what's actually on the stack — provably equivalent for the real scan path).
- **Gating**: one hardcoded flag, `true` only under `SECANT_TESTNET`; that's the single `#if` in the feature. Everything compiles in all targets (so `zodlTests`, built as zodl-internal, covers it) but is unreachable outside testnet — production behavior is byte-identical, including the existing 5-row demo fixture.

Approved spec: in the [MOB-1480 ticket description](https://linear.app/zodl/issue/MOB-1480/migration-sdk-simulator-for-testnet-builds-debug-menu-background) — specs live on the ticket, not in the repo. Inspired by the MOB-1451 prototype simulator (5f4bccef) and its follow-ups; re-derived against the post-MOB-1478 UI rather than ported.

Linear: [MOB-1480](https://linear.app/zodl/issue/MOB-1480/migration-sdk-simulator-for-testnet-builds-debug-menu-background)

## Test plan

- Full `zodlTests` on `zodl-internal`: **1023 tests in 134 suites passed** (1 pre-existing known issue: `CurrencyConversionSetupTests`, MOB-1364), `-skipMacroValidation`. Net **+78 tests** vs the base branch: engine lifecycle/timing/failure semantics, state-store versioning, simulated-client overlay (active + inactive fallback), panel store, reset-flags, and the Keystone bypass coordinator round-trips.
- `zodl-testnet` build green (device + simulator).
- Manual QA (suggested): run `zodl-testnet` in a simulator, long-press the balance, walk the presets against the flows; fire a real background wakeup via LLDB `_simulateLaunchForTaskWithIdentifier` (hint shown in the panel footer).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ directly impact users rather than highlighting other crucial architectural updat
 - [MOB-1466] Ironwood migration screens are wired into a complete guided flow (entry, scheduling, status, recovery) driven by migration state, with a Home banner entry point — dormant until the migration SDK ships.
 - [MOB-1467] Scheduled Ironwood migration transfers can now send from background tasks, with local notifications for progress, required actions, and manual-mode send reminders — dormant until the migration SDK ships.
 - [MOB-1468] Keystone hardware-wallet accounts can complete the Ironwood migration — transfers are signed by QR in a single batched session — dormant until the migration SDK and Keystone batch support ship.
+- [MOB-1480] Testnet builds gain a migration simulator: long-press the balance on the home screen to open a debug menu that walks every Ironwood migration path — scenario presets, failure injection, simulated time, on-demand background-delivery runs with real notifications, and a simulated Keystone signing pass. Off outside testnet builds; production behavior is unchanged.
 
 ### Changed
 - [MOB-1472] The assets you can swap are now a curated set of major coins and stablecoins across the supported chains (plus swapping to ZEC), instead of the full list from the swap provider, and the address-book chain picker is limited to those chains. Existing swaps in your history (including assets no longer offered) still display normally, and existing contacts on other chains are preserved.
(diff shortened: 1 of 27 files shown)


Chlup:
Superseded by #1900 — the head branch was renamed to `michal/MOB-1458/MOB-1480-migration-sdk-simulator` (Linear-ticket branch schema), and GitHub closes open PRs on a head-branch rename. Identical commit (043189ce); no review activity was lost.

PR #1896, opened 2026-07-14
[PRO-325] Surface FAILED and PROCESSING statuses for swaps out of ZEC and CrossPays

zcash-harry:
## Why

The 1Click status parser in `Near1Click.swift` only mapped `INCOMPLETE_DEPOSIT` / `PENDING_DEPOSIT` / `REFUNDED` / `SUCCESS` for the out-of-ZEC direction — `FAILED` and `PROCESSING` fell through to `.pending`. A swap out of ZEC or CrossPay that failed on NEAR's side after the deposit landed therefore showed "Paying…" / "Swapping…" indefinitely, kept polling forever, and never surfaced the failed state or the contact-support footer. (The swap-to-ZEC branch already mapped both correctly.)

Context: [PRO-325](https://linear.app/zodl/issue/PRO-325/suppress-the-amount-for-failed-transactions-in-the-activity-list#comment-28337a0b) — we can't safely message "funds are still in your wallet" while a NEAR-side failure is indistinguishable from a pending swap.

## What

- Extracted the status-string mapping into `Near1Click.swapStatus(from:isSwapToZec:)` (pure, unit-testable — same pattern as `curated(_:)` / `amountMessageResolution`) and added `FAILED` → `.failed`, `PROCESSING` → `.processing` to the out-of-ZEC branch.
- The direction-specific `PENDING_DEPOSIT` mapping is preserved (`.pendingDeposit` for swap-to-ZEC, `.pending` otherwise) — the transaction-detail deposit-info footer keys off `.pendingDeposit`.
- Android already stores `FAILED` / `PROCESSING` in shared user metadata, so cross-platform metadata is unaffected (verified against `SwapStatusSerializer.kt`).
- Drive-by (separate commit): updated `curatedKeepsSupportedAndDropsRest`, which has been failing on `main` since USDC@pol was added to the curated list in 196249a0.

## Testing

- New Swift Testing cases in `Near1ClickTests` cover the full mapping in both directions, including the `KNOWN_DEPOSIT_TX` → `.pending` fallback.
- `xcodebuild test -only-testing:zodlTests/Near1ClickTests` — 14/14 pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ directly impact users rather than highlighting other crucial architectural updat
 
 ## [Unreleased]
 
+### Fixed
+- [PRO-325] Swaps out of ZEC and CrossPays that fail on the swap provider's side now show "Swap Failed" / "Payment Failed" (with the contact-support option) instead of appearing to stay in progress forever. Long-running swaps in this direction also correctly show their processing state.
+
 ## 3.7.3 build 1 (20026-07-12)
 
 ### Changed
--- a/secant/Sources/Dependencies/SwapAndPay/sources/Near1Click.swift
+++ b/secant/Sources/Dependencies/SwapAndPay/sources/Near1Click.swift
@@ -216,6 +216,32 @@ struct Near1Click {
         assets.filter { Constants.supportedAssetIds.contains($0.assetId) }
     }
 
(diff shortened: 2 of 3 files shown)


Chlup (approved):
utACK

PR #1898, opened 2026-07-15
[MOB-1483] Gate the migration offering on Ironwood network activation

Chlup:
## Summary

Gates the Orchard → Ironwood migration offering on **"Ironwood (NU6.3) is actually activated on the current network"**: known chain tip ≥ the network's activation height. Before that — no migration banner, no way into the flow, no reconcile side effects, and background migration sessions complete immediately without notifying or re-arming. Today that means mainnet builds are gate-closed (tip < 3,428,143) while testnet is open — which is the point: the rust migration crate has no activation guard of its own, so this app-layer gate is currently the only thing preventing a pre-activation migration attempt.

- **The check**: `tip > 0 && tip >= ironwoodActivationHeight` — tip from the synchronizer's cached network chain tip; unknown tip (fresh install, offline) fails closed. Heights live on `ZcashSDKEnvironment` (new `ironwoodActivationHeight()`): mainnet **3_428_143**, testnet **4_134_000** — mirrored from `zcash_protocol` 0.10.0 (consensus-frozen; the Rust-sourced getter naturally replaces this at MOB-1455 real-SDK wiring). Custom/regtest networks read `customActivationHeights.nu6_3`, defaulting to never-activated.
- **Derivation tables**: `MigrationDerivations.bannerVariant` / `.reentryRoute` gain a leading `isIronwoodActivated` input — `false` short-circuits to no banner / plain entry before any other branch. `reconcile()` early-returns when gated. The client member defaults to `{ false }` (fail-closed everywhere, including macro-generated test values).
- **Background sessions**: `migrationBackgroundSessionEffect` gets a first branch — not activated → complete the task, **no notification, no re-arm** (a stale-armed task chain self-heals by dying out); the expiration handler likewise skips its re-arm.
- **Reactivity (the subtle part)**: nothing re-evaluated the banner on chain-tip movement, and the cold-launch priority walk runs while the tip is still 0 — so even an activated network would have shown no migration banner for the entire first session. `SmartBannerStore` now merges a flip-latch effect into the existing `synchronizerStateChanged` subscription: when the activation state flips (first tip fetch, activation-day crossing, or a reorg back below), the migration variant re-evaluates through the existing in-place raise/lower path. The old ~80-line sync-status body was extracted verbatim into a helper so the flip check can't be swallowed by that case's early returns (which fire on nearly every tick during a restore — exactly the cold-launch scenario).

Approved spec: in the [MOB-1483 ticket description](https://linear.app/zodl/issue/MOB-1483/dont-offer-ironwood-migration-before-the-network-activates-ironwood).

## Integration notes for PR #1894 (migration simulator — unmerged)

Whichever of the two lands second needs two mechanical follow-ups on the combined tree (both compiler-enforced or one-liners):
1. **Simulator gate bypass**: prepend `MigrationSimulatorFlag.isEnabled && MigrationSimulatorClient.sharedEngine.isActive → true` to `MigrationManagerImpl.isIronwoodActivated()` (same idiom as #1894's two existing hooks) — otherwise a fresh-install/offline testnet run (tip 0) hides the simulator's banner.
2. #1894's preset derivation-table tests call `MigrationDerivations.bannerVariant/reentryRoute` directly and need the new leading `isIronwoodActivated: true` argument.

Also a trivial one-line CHANGELOG conflict, as between any two of the open PRs.

## Test plan

- Full `zodlTests` on `zodl-internal`: **964 tests in 130 suites passed** (1 pre-existing known issue: MOB-1364), `-skipMacroValidation`. Net **+19 tests** vs the base branch: activation-formula boundaries (tip 0 / below / exact / above), gate-beats-everything cases for both derivation tables (+ a new leading row in the all-routes precedence table), reconcile-guard call-count spies, gated background session and gated expiry (no re-arm), and 5 SmartBanner flip-latch tests (first-observation both ways, raise, lower, no-spam).
- `zodl-testnet` build green.
- Every pre-existing test that the fail-closed default would have silently flipped was updated deliberately (a shared `isIronwoodActivated = { true }` baseline in the BG tests; explicit tip overrides in the two reconcile tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ directly impact users rather than highlighting other crucial architectural updat
 - [MOB-1466] Ironwood migration screens are wired into a complete guided flow (entry, scheduling, status, recovery) driven by migration state, with a Home banner entry point — dormant until the migration SDK ships.
 - [MOB-1467] Scheduled Ironwood migration transfers can now send from background tasks, with local notifications for progress, required actions, and manual-mode send reminders — dormant until the migration SDK ships.
 - [MOB-1468] Keystone hardware-wallet accounts can complete the Ironwood migration — transfers are signed by QR in a single batched session — dormant until the migration SDK and Keystone batch support ship.
+- [MOB-1483] The Ironwood migration is only offered once the network has actually activated Ironwood — before the chain tip reaches the activation height there is no migration banner, no way into the flow, and no migration background work or notifications. The offer appears automatically once the activated network is detected, without restarting the app.
 
 ### Changed
 - [MOB-1472] The assets you can swap are now a curated set of major coins and stablecoins across the supported chains (plus swapping to ZEC), instead of the full list from the swap provider, and the address-book chain picker is limited to those chains. Existing swaps in your history (including assets no longer offered) still display normally, and existing contacts on other chains are preserved.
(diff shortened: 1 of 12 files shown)


Chlup:
Superseded by #1901 — the head branch was renamed to `michal/MOB-1458/MOB-1483-migration-ironwood-activation-gate` (Linear-ticket branch schema), and GitHub closes open PRs on a head-branch rename. Identical commit (25818b99); no review activity was lost.

PR #1812, opened 2026-06-08
chore: triage test suite + drop unused secantUITests target

LukasKorba:
## Summary

The \`secantTests\` target's Sources phase had drifted to compile only **6 files** (all recent voting tests), while **49 legacy test files sat orphaned on disk** under \`secantTests/\` — a worse credibility signal for an OSS repo than an honest, lean suite. After the rebrand + \`@Reducer\` macro adoption, most legacy tests couldn't compile against the current code anyway.

This PR:

1. **Revives 23 test files** (296 test methods, **100% passing** under \`xcodebuild test\` on \`zashi-internal\`).
2. **Deletes 32 dead files** — Root/Home/Send/Scan/Settings/Receive/Tabs/TransactionList/BalanceBreakdown/SyncProgress/Deeplink TCA tests broken by macro-era state-shape drift, all SnapshotTests (broken by view-API drift; snapshot recordings stale post-rebrand), \`WalletConfigProviderTests\` (Combine → async/await architecture shift). Recoverable from \`git log\` if anyone needs to resurrect a specific test later.
3. **Drops the \`secantUITests\` target** entirely — it contained a single stub file with no real coverage and the team uses Maestro for E2E. Down to **6 targets** (5 app variants + 1 unit-test target).

## Surviving test set (23 files, 296 tests)

Pure utility/parsing tests are the bulk: \`Zatoshi\`, \`Logger\`, \`DatabaseFiles\`, \`SecItemClient\`, \`UserPreferencesStorage\`, \`WalletStorage\`, \`MemoByteCounting\`, \`QRCodeGenerator\`, \`ZatoshiStringRepresentation\` (×2), \`CurrencyConversion\`, \`DeeplinkURLParsing\`, \`DecommissionedServerMigration\`. Plus 5 voting tests, 1 Home test (\`testSynchronizerErrorBringsUpAlert\`), \`PrivateDataConsent\`, \`ReviewRequest\`, \`SensitiveData\`, \`RecoveryPhraseBackup\`.

## Small API-drift fixes applied to the survivors

- \`SecItemClientTests\` — \`WalletStorage.deleteData(forKey:)\` is now \`throws\`, not \`Bool\`-returning. Test updated to use \`XCTAssertNoThrow\` / \`XCTAssertThrowsError\`.
- \`WalletStorageTests\` — \`WalletStorageError\` switch made exhaustive; added \`try\` on \`storage.resetZashi()\`.
[...]

PR #1907, opened 2026-07-19
[MOB-1496] Rewire the migration feature onto the real SDK Synchronizer API

Chlup:
Closes the app side of the Orchard → Ironwood integration: every migration stub on `SDKSynchronizerClient` is replaced by the real per-account `Synchronizer` migration API (SDK branch `michal/slipstream-support`, consumed via the existing local package reference), and the app's privacy logic is aligned with the current specs (MOB-1458, ZIP-0318, the network-privacy decisions of 2026-07-19). Spec: MOB-1496 (Linear), approved 2026-07-19. Supersedes the prototype-integration approach of MOB-1469 (PR #1888 can be closed).

## What changed (12 commits)

**Seam + models.** The 28 stub members became the real per-`AccountUUID`, throwing surface (USK derived at confirm-time via the send-flow pattern; Keystone lanes never touch USK members). The 8 shadow model types are deleted in favor of the SDK's (4 renames); the app-only `.transferStalled` state became derived logic. The MOB-1480 simulator is fully adapted and keeps faking the surface on testnet builds.

**App-owned persistence + progress.** The SDK retains no proposal list, so the app persists the committed schedule per account (append-only sent records across restarts) and derives all rows/summaries from it plus live SDK flags. The migration state stream is app-owned, with the banner emit-fix and Root-level reconcile triggers (sync-completion edge, gate flips).

**The 10-minute gates, split by owner.** The SDK owns broadcast→sync (the app's duplicated gate is retired; every `start()` defers silently on `migrationSyncBlocked` and resumes off the gate stream — including after the app itself stops sync for a broadcast). The app owns sync→broadcast (`sendGate` re-keyed to observed sync completions + `migrationPrivacySyncBufferDuration`; the S10 footer reads the real value). Sync is stopped before every foreground migration broadcast (all lanes share one helper). `migrationRecordFailedAfterBroadcast` is treated as broadcast-landed everywhere (never re-sent).

**Network privacy.** Per-account atomic `MigrationNetworkSnapshot` (Tor choice + sync provider + broadcast endpoint), taken under the transaction guard at the first broadcast-bearing confirmation and immune to mid-run switches. The broadcast endpoint auto-picks the *other* provider family (`*.zec.rocks` ↔ `*.zec.stardust.rest`, benchmarked); **a user-selected custom server is used for all operations** (no separation, per decision); testnet falls back to same-server. Auto server selection is pinned to the snapshotted sync provider while any run is active (pending candidates re-validated); manual switches onto a run's broadcast provider get a privacy warning (EN + ES/needs_review).

**Background + ZIP alignment.** The BG decision tree runs per account (software + Keystone concurrently): sync-priority sessions, at most one broadcast per session across accounts (overdue first, then earliest due), re-arm from the earliest window across accounts, and background work cancels only when *every* account is done. The S10 "Send now" lane is capped to exactly one overdue transfer (ZIP-0318 MUST — a behavior change from the shipped sequential batch).

**Keystone integrity + dust.** Fixed a latent real-SDK break: the note-split sentinel no longer rides the schedule-PCZT store (the engine rejects non-engine ids) — and, per the whole-branch review's engine trace, the store sequence is now the engine-safe one: **split stored first (it creates the engine run), split broadcast, then the schedule stored on broadcast success** (`BroadcastScheduled` is set after the phase the denom-advance can fire from), with a store-only retry sheet and Continue gated until the schedule is stored. Batch re-pairing is validated (any mismatch abandons with nothing stored). The Keystone dust lane works (batch-of-1 → store → execute). The mid-commit split screen hides back navigation, and the split-screen Continue uses a deferred pop (a TCA inline-pop crash was caught and fixed pre-commit).

**Sweep.** Activation height now sourced from the SDK's `ZcashNetwork.ironwoodActivationHeight` (values verified equal at the pinned rev); stub-era comments retired; review-triage fixes (storage self-heal on corrupt payloads, sentinel drift-guard tests, dead-code removal, alert-state cleanup); CHANGELOG consolidated.

## Verification

- Full `zodlTests` on `zodl-internal`: **1258 Swift Testing tests (exactly the 1 known MOB-1364 issue) + 48 XCTest, 0 unexpected failures**; `zodl-testnet` builds clean. Red-first TDD on every behavior change (order pins, caps, gate deferral, emit rules).
- Process: fresh implementer + adversarial reviewer per task with fix loops, then a whole-branch review that read the **`zcash_pool_migration` engine crate** — which is what caught the Keystone store-ordering defects (C-1/C-1b) the Swift-contract-level reviews could not see. Final verdict on the shipped tree: see the review record on the ticket.
- The SDK repo is untouched; both `SDKSynchronizer` and `SlipstreamSynchronizer` carry the surface, so the `useSlipstreamSynchronizer` flag is unaffected.

## SDK/engine-boundary findings (reported, deliberately NOT worked around app-side)

The whole-branch review's engine trace surfaced three contract gaps at the pinned rust rev (`5115cf26`) that exist independent of this PR and belong to the SDK/engine follow-up (details on MOB-1496):

1. **Engine `.complete` is one-shot** — the first state read consumes it; later reads see `.notStarted`. The app's completion UX (banner, Complete screen, dust/lock resolution) is racy-to-unreachable until the engine makes completion sticky (or the app is told to latch it).
2. **Completion requires `orchard_spendable == 0` strictly**, but scheduled runs deliberately leave the residual — runs with change may never reach `.complete`, orphaning the dust flow in exactly the cases it exists for. Likely engine fix: a below-threshold tolerance.
3. **Software split retry re-signs into a fresh engine run per attempt** (works, but orphans runs; a false-negative network error can strand `splitPendingConfirmation`). An executeNext-style retry for the software prep would fix it.
4. **A failed engine record of the split's broadcast success can park a run the same way** (the prep stays pending while the phase machine can't advance) — byte-identical on the software lane, purely engine-fault-triggered; noted from the final review's phase-machine trace.

## Notes for reviewers

- **Tor doc reconciliation is still owed** (spec §M): the network-privacy requirements doc (R1–R13) never arrived; the network sections were built from the approved decisions + ZIP-0318. R13's disclosure surface is deliberately NOT implemented until the doc lands.
- **QA/testnet items**: real-engine residual-only `proposeMigrationTransfers(includeResidual: true)` semantics post-completion; "scheduled run with residual reaches engine-complete" (finding 2 above); the live gate chain (stop → broadcast → buffer → resume) under real network conditions; Keystone batch signing against the real firmware when it ships.
- **Known limitations on record**: the simulator does not fake the gate trio; `refreshStaleMigrationTransfers` stays an available-but-unused SDK affordance (recovery re-creates plans by design); a narrow redundant `start(true)` can fire when a resume races the recovering start's own seed read (bounded, self-terminating); a stored dust plan whose transfer isn't height-due yet surfaces the retry sheet (Retry-mitigated).
- The feature remains gated behind Ironwood network activation and is not in any released build; all CHANGELOG entries carry that qualifier.

## Round 11 — final-engine adaptation (2026-07-21)

The SDK moved off the v1 `zcash_pool_migration` crate onto the final rust engine (MOB-1495 Round 5; SDK PRs zcash/zcash-swift-wallet-sdk#1813 / #1812, consumed from `michal/slipstream-support` @ `af3c51ca`). Four new commits (`2e73e081..047b6ad9`) adapt the app:

- **The Keystone note-split pair went plural** — the final engine builds N preparation transactions, not one split tx. `SDKSynchronizerClient` reshapes to `proposeNoteSplitPCZTs -> [MigrationUnsignedTransferPczt]` / `storeSignedNoteSplits([MigrationSignedTransferPczt])`; the coordinator's `"note-split"` sentinel id becomes a per-entry `"note-split#<engineId>"` prefix (partition + strip in the same two helpers); the Keystone retry lane persists the pair array; UR encode/scan/re-pair were already count-agnostic.
- **Unconditional prep fold** — immediate-mode batches can carry preps under the final engine (the immediate flag only rewrites transfer heights; the old S1 "immediate is split-free" engine claim is obsolete). `proposeKeystoneBatch` drops its mode gate and folds the (possibly empty) prep subset unconditionally.
- **Abandon reconciliation** — the engine creates the run at PCZT-build time and always resumes a non-terminal run, ignoring newer previews. Abandoning a signing ceremony after PCZTs were proposed now cancels the stray run (`restartCurrentMigrationStep`, fire-and-forget) both in-flow and at the external-teardown sites that release the send-wait hold. NOT covered: process death mid-ceremony (the in-memory ceremony flag is gone on relaunch) — see QA items.
- **Sequential runs** — engine `.complete` is per-run and re-readable now, never "nothing left". On completion the app asks the engine whether anything remains via one fresh propose per completion transition (persisted tri-state flag; deliberately never per-reconcile — a propose overwrites the SDK plan cache, and re-proposing under an in-flight review would `migrationPlanStale` the commit). More remains ⇒ new "Migration batch finished" notification (EN + ES `needs_review`) and the migration banner resurfaces as required after the user acknowledges the Complete screen; genuinely done ⇒ "Migration complete" exactly as before. The BG `cancelAll()` stays in both cases — nothing is broadcastable until the user consents to the next run, whose commit re-arms scheduling.
- **Dust "Migrate anyway"** is now a benign dead end (the engine returns an empty residual schedule, so the composite yields nil → the generic failure sheet). Michal-decided: left as-is until the rust dust-locking primitive lands and the dust UX is redesigned on top of it.

**The four SDK/engine-boundary findings above are RESOLVED by the final engine**: (1) `.complete` is sticky/per-run (this round builds on exactly that); (2) completion is per-run mined-ness — the residual staying in Orchard no longer blocks it; (3+4) `commit_or_resume` resumes a stored non-terminal run instead of re-signing fresh runs, and the v1 phase-machine parking hazards no longer exist under the final engine's state model.

**Gates at `047b6ad9`**: full `zodlTests` 1351 tests / 141 suites, 1 known issue (MOB-1364), TEST SUCCEEDED; `zodl-testnet` BUILD SUCCEEDED.

**New QA/testnet items**: kill-app-mid-Keystone-ceremony leaves a stray engine run (the account can look mid-migration; recovery = the restart lane); software-immediate with a prep-requiring balance shows the retry sheet between prep mining and the transfer send (executeNext correctly reports nothing due yet); sequential-run flow end-to-end (run completes → batch-finished notification → banner → second run commits and re-arms BG).


Chlup:
## Ultra-review remediation (R8) — 10 commits, `8585cad0..2666df32`

This pushes the full remediation of the `/code-review ultra` findings (24 confirmed defects: the ranked 15, the 6 confirmed-but-cut, and the 3 PLAUSIBLE — all verified, plus the 6 sanctioned cleanups). Spec and per-finding dispositions live on the MOB-1496 ticket ("R8 spec" comment + the shipped-summary comment).

**Process**: 7 task commits + 3 fix-wave commits, each through an adversarial per-task review (2 Criticals and 1 Important were caught by those reviews mid-round, fixed, and verified closed by their finders), then a cumulative whole-branch review that read the rust engine crate wherever app assumptions touch engine semantics. Final verdict: **READY**. Full suite: **1337 Swift Testing tests (+79 vs the pre-round baseline; 1 pre-existing known issue MOB-1364) + 48 XCTest**, zodl-internal and zodl-testnet builds green.

**Highlights** (full list in the commit messages):
- Immediate migration is structurally split-free (the engine's immediate path never expects a split — the old lane could sign a double-spending sweep).
- Cold launches survive the persisted privacy sync-gate, and the BG delivery chain can no longer be silently consumed (stash-and-replay + guarded expiration + conservative-active unreadable accounts).
- Completion acknowledgment is per-account and guarded on the engine's `.complete`; the immediate-mode close no longer wipes a live run's records.
- **Send-now redesigned per the approved silence-window decision**: the CTA enables when a transfer is due; tapping stops sync and broadcasts after a countdown inside a sync-silence window (Cancel resumes sync; external flow teardown releases the hold).
- Migration notifications carry their account and switch to it on tap.
- Automatic server selection (including Settings' automatic Save) can no longer land sync on an active run's broadcast provider.

**Recorded remainders** (deliberate, non-blocking): the engine escalations from the PR body remain deferred to the rust update — with one addendum: the rust fix must make `.complete` re-readable (it is one-shot at every layer today, so the new ack guard only fully engages after that; cleanup meanwhile self-heals via reconcile + the abandon-clear). Open product flags: whether the dust/manual lanes should also wait out the silence window; a quick manual QA pass on swipe-dismiss of the waiting screen; 4 new strings need the design/copy pass (`migrationPlan.proposeFailedTitle/Body`, `migrationSending.waitingTitle/Body`).

The MOB-1497 branch (PR #1909) is being rebased onto this head next.

PR #1921, opened 2026-07-22
[#1920] Handle Keystone hardware wallet connection failures

### LINKED ISSUE #1920: App gets stuck when adding a Keystone device fails.
When the users wants to connect their existing keystone to ZODL and an error occurs, the app gets stuck in the "keep open while syncing" screen and user is not shown an error or a way to return to the previous screen. the OK button does nothing.

pacu:
## Summary
Fixes #1920. Two related fixes to the add-Keystone flow:

**1. Import failures are surfaced (originally silent).** A failed Keystone account import used to do nothing (`// TODO: error handling` in `AddHWWalletStore`), leaving the user stuck on the "keep ZODL open" screen with a dead OK button. Failures now present a "Connection Failed" sheet (same UI pattern as `DisconnectHWWallet`) at the coordinator level, so it works from both entry points that trigger the import — "Connect new device" (`KeystoneDeviceReadyView`) and the restore/birthday flow. The sheet offers **Contact Support** (mail, or share sheet when Apple Mail isn't available) and **Cancel**, which exits the whole flow so the user is never stranded. (An earlier revision had a **Try Again** action; it was removed — the user can't tell whether a retry will help, and the firmware guidance already covers the fix.)

**2. The failure sheet no longer pops over the success screen.** With the error flow in place, a *successful* connect could still show the failure sheet on top of the success screen. Root cause: the buttons that start the import stayed enabled with no feedback while `importAccount` ran (a slow, network-bound call), so re-taps started **duplicate imports**; the duplicate throws "account already exists" and that error surfaced as a bogus Connection Failed sheet. Fixed in three layers:
- `unlockTapped` now guards against re-entry via an `isImportingAccount` flag (cleared on success/failure so genuine retries work).
- The connect/OK buttons show a `ProgressView` and disable while the import runs (established `accessoryView: ProgressView()` pattern; the restore-info OK button gets it via a new `RestoreInfo.State.isProcessing` flag managed by the coordinator).
- Belt-and-braces: the coordinator drops an `accountImportFailed` that arrives after `keystoneConnected` is already on the stack.

Also fixes a previously **unhandled `try`** when reloading wallet accounts after a successful import.

## Privacy
The string carried into the support message is `error.localizedDescription` **only**. For `ZcashError` that resolves to a static `"<code>: <message>"` (e.g. `ZRUST0067: Error from rust layer when calling ZcashRustBackend.importAccountUfvk`). The raw Rust error associated value — which could in principle contain key material — is **not** interpolated into `localizedDescription`, and we never use `String(describing:)`/`"\(error)"`. So **no UFVK/seed/fingerprint can leak** into logs or support emails. This matches the existing safe pattern in `DisconnectHWWallet`.

## Changes
- `AddHWWalletStore.swift` — `accountImportFailed` carries the (safe) error string; both throw sites route to it; `isImportingAccount` in-flight guard.
- `AddKeystoneHWWalletCoordFlowStore/Coordinator/View.swift` — failure + support state, the failure sheet, restore-info `isProcessing` wiring, and the drop-late-failures guard.
- `KeystoneDeviceReadyView.swift` / `RestoreInfoView.swift` / `RestoreInfoStore.swift` — progress + disabled states on the import buttons.
- `Localizable.xcstrings` — new keys for the failure sheet (en + es).
- `CHANGELOG.md` — Fixed entries.

## Testing
- ✅ New Swift Testing coverage: element store (hex parsing, unlock guard, re-tap suppression, failure propagation, flag resets) and coordinator (failure sheet shown/hidden, late-failure ignored, cancel/contact-support routing, and a success-path regression test asserting exactly one `importAccount` call with no failure sheet).
- ✅ Fixed the support-flow tests that hit the live `walletStorage` dependency when actually run.
- ✅ `AddKeystoneHWWalletTests` + `AddKeystoneHWWalletCoordFlowTests`: 23/23 passing.
- ✅ Verified manually in the UI: failure path shows the sheet; successful connect shows only the success screen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ directly impact users rather than highlighting other crucial architectural updat
 ## [Unreleased]
 
 ### Fixed
+- [#1920] Connecting a Keystone hardware wallet that fails now shows a clear "Connection Failed" message (with Try Again and Contact Support options) instead of silently doing nothing. The support message includes a safe error identifier and never exposes any wallet keys.
 - [PRO-325] Swaps out of ZEC and CrossPays that fail on the swap provider's side now show "Swap Failed" / "Payment Failed" (with the contact-support option) instead of appearing to stay in progress forever. Long-running swaps in this direction also correctly show their processing state.
 
 ## 3.7.3 build 1 (20026-07-12)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ directly impact users rather than highlighting other crucial architectural updat
 ## [Unreleased]
 
 ### Fixed
-- [#1920] Connecting a Keystone hardware wallet that fails now shows a clear "Connection Failed" message (with Try Again, Contact Support, and Cancel options) instead of silently doing nothing. Cancel leaves the flow so the user is never stuck on the connection screen. The support message includes a safe error identifier and never exposes any wallet keys.
+- [#1920] Connecting a Keystone hardware wallet that fails now shows a clear "Connection Failed" message (with Contact Support and Cancel options) instead of silently doing nothing. Cancel leaves the flow so the user is never stuck on the connection screen. The support message includes a safe error identifier and never exposes any wallet keys.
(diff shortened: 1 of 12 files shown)


Chlup (approved):
utACK

Chlup (approved):
utACK

PR #1939, opened 2026-07-24
[MOB-1513] Sanitize note-split sentinel ids around the Keystone batch apply

Chlup:
## Summary

Companion to zcash/zcash-swift-wallet-sdk#1834, fixing MOB-1513 R8 finding 1 from the app side: the SDK's batch-apply FFI numeric-parses every PCZT id, but the planCommit Keystone ceremony hands it the proposed batch verbatim — including preparation entries under the app-side `note-split#` sentinel prefix (stripped only after apply, in `splitKeystoneBatch`). The non-numeric ids made `applyKeystoneBatchSignatures` throw right after every successful device scan, abandoning any scheduled Keystone ceremony that carried a preparation transaction — the same failure class as the immediate-lane loop fixed in #1935 (now at least logged, but still failing).

## Changes

- `MigrationCoordFlow.sanitizedForBatchApply` strips the sentinel down to the bare (numeric) engine id right before the apply call; `MigrationCoordFlow.reattachOriginalIds` restores the ceremony's original ids positionally onto the result (apply's contract: output order/count == input's; a contract-breach count mismatch returns the SDK result untouched rather than zip-truncating). Nothing downstream changes — `splitKeystoneBatch` and both stores see exactly the ids they always have.
- Kept as defense-in-depth alongside the SDK contract fix (opaque pass-through ids, SDK PR #1834): correct against both SDK generations, and the app's sentinel bookkeeping no longer needs to cross the FFI at all.
- CHANGELOG entry.

## Tests

Red-first: `foundKeystoneBatchSignaturesSanitizesSentinelIdsForApplyAndRestoresThemDownstream` captures the ids actually crossing the SDK boundary (must be `["7", "12"]`, not `["note-split#7", "12"]`) and re-asserts the downstream prep/schedule stores still receive the sentinel round-tripped; three pure-helper tests cover strip-only-sentinels, positional re-attachment, and the count-mismatch fallback. Full `zodlTests`: **1689 tests / 155 suites passed**; `zodl-testnet` build succeeded (`-skipMacroValidation`).


Chlup:
Note on the red CI: the build/unit_tests failures are the repo-wide pre-existing condition, not this change — the workflows have no checkout step for the local-path SDK dependency (`../zcash-swift-wallet-sdk` doesn't exist on the runner; it fails in ~20s at package resolution, identically on the already-merged #1938). Real gates for this PR ran locally against the SDK checkout: full zodlTests 1689/1689 + zodl-testnet build, -skipMacroValidation. Adapting CI to the local-dependency era (checking out the SDK repo at a pinned branch next to the app) is a separate decision/change.

PR #1959, opened 2026-07-28
[MOB-1580] Show only the fee for a transaction sent to yourself

Chlup:
A transaction the user sends to themselves — a manual send to one's own address, or an Orchard → Ironwood migration crossing — now displays the network fee, on every device and at every point in the transaction's life.

## The change is a deletion

`zecAmount` is the account balance delta. Every output of a self-transfer returns to the account, so the delta is exactly `-fee` and the default path already displays the fee. No self-transfer concept is needed.

The only thing preventing it was the note-count fallback added for the Ironwood migration display, which substituted `totalReceived` whenever the SDK reported no counted sent or received notes.

## Why that condition could not work

- `sent_note_count` counts rows in the SDK's `sent_notes` table, which only exists on the device that **created** the transaction — and its CTE further excludes sent notes whose output was recorded as change.
- `received_note_count` excludes change.
- Scanning marks a self-send's payment output as change (upstream's comment names "notes sent from one account to itself" as an intended case), and the DB upsert ratchets the flag upward.

So both counts are device-dependent *and* time-dependent. The same self-transfer reads 1/1 before its block is scanned and 0/0 afterwards — which is why the displayed value grew once the transaction confirmed, and why a second device disagreed with the sending one.

## Scope

| Shape | Device / timing | Before | After |
|---|---|---|---|
| Manual self-send | sending device, pre-scan | fee | fee |
| Manual self-send | after scan, or 2nd device | payment + change | **fee** |
| Ironwood migration | any | crossing amount | **fee** |
| Ordinary send | sending device, pre-scan | amount + fee | amount + fee |
| Ordinary send | after scan, or 2nd device | change | amount + fee |
| Receive, shielding, swaps, pending send, swap deposit | any | — | unchanged |

The ordinary-send rows are corrections of the same defect, not new behaviour — one condition produced both wrong numbers and nothing within it distinguished the shapes.

`sentNoteCount` / `receivedNoteCount` are removed from the model with the condition, since nothing else read them. The CHANGELOG entry promising that an Ironwood transfer shows the amount that moved is removed along with the behaviour it described; no new entry is added, as the fallback never reached a finalized release.

## Verification

Full `zodlTests` target: 679 tests in 111 suites pass. Written test-first — `ironwoodMigrationDisplaysTheFee` and `selfSendDisplaysTheSameFeeBeforeAndAfterScanning` both failed against the old code before the deletion.

The previous implementation of this branch is preserved at `michal/MOB-1580-self-transfer-display-value-superseded`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ directly impact users rather than highlighting other crucial architectural updat
 
 ### Added
 - [Ironwood] ZODL now recognizes funds held in the Ironwood shielded pool. Balances on the home screen, in the balances breakdown and in the shielding banner include Ironwood alongside Sapling and Orchard, so those funds are visible and counted as soon as the network upgrade activates.
-- [Ironwood] A transfer that moves funds into the Ironwood pool now shows the amount that actually moved, in both the transaction list and the transaction detail screen. Previously such a transfer displayed only its fee.
 - [Ironwood] Once ZODL sees that the Ironwood network upgrade is live on the network, it shows a short one-time screen introducing the change, with a link to a support article. It appears once per device — after you continue past it, it never comes back.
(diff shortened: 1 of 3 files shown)


LukasKorba (commented):
LGTM

LukasKorba (approved):

PR #1960, opened 2026-07-28
[MOB-1581] Refresh the transaction list the moment a send completes

Chlup:
## Problem

After sending, the "Sending" row could take over a minute to appear in the Activity list (MOB-1581, reported on 3.8.0 (9)).

The transaction is provably in the local database the moment creation succeeds — the SDK's create path even reads it back through the same connection the Activity list queries. The data was never late; the refresh was:

- The Activity list refetches only on sync events, on `.upToDate` transitions, and on the send flows' result-screen **Close** buttons.
- Leaving the success screen via **View Transaction** and then closing the transaction detail refreshed **nothing** (those arms only cleared navigation state).
- An idle, fully-synced wallet emits no sync event until the next block (~75 s average) — so with the wrong exit path, the row stayed invisible until the next block's sync cycle. Activation-day sync/mempool congestion stretches that further.

(The wrong amount shown while pending is a separate issue, tracked and fixed under MOB-1580. The Jul 25 reports of the same symptom on internal builds had a different, already-fixed cause — the fetch-starvation regression addressed on `release/3.8.0`.)

## Fix

`Root` now dispatches `.fetchTransactionsForTheSelectedAccount`:

1. **The moment a send reaches a terminal outcome that stored a transaction** — `sendDone`, `sendPartial`, and `sendFailed` with the txid present in the DB — across all four send-capable flows (Send, Scan, Sign with Keystone, Swap & Pay), including the Keystone `confirmWithKeystone` path elements and the Swap flow's own flow-level send actions. `sendFailed` with nothing stored stays silent.
2. **On those flows' `transactionDetails(.closeDetailTapped)` arms** (the View Transaction exit), which previously returned `.none`.

Concurrent fetches are safe by design: the fetch effect deliberately has no `cancelInFlight` (see the starvation comment in `RootTransactions.swift`), and the `.fetchedTransactions` provenance guard drops wrong-account payloads.

## Tests

New `RootSendCompletionRefreshTests` (13 tests, Swift Testing, serialized, plain-`Store` + `LockIsolated` spy pattern per `RootTransactionsAccountSwitchTests`). TDD: all 12 positive tests failed before the source change; all 13 pass after. Full `zodl-internal` suite: **692 tests in 112 suites passed** (1 pre-existing known issue, unrelated).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ directly impact users rather than highlighting other crucial architectural updat
 - [MOB-1512] The "Wallet data replaced" notice now stays on screen after the healed wallet opens, instead of disappearing during the transition to the home screen.
 - [Keystone] Switching between the ZODL and Keystone accounts now always shows the selected account's transaction history — a slow in-flight refresh for the previous account can no longer overwrite it — and connecting a Keystone hardware wallet refreshes the transaction list and balance immediately.
 - [Keystone] The transaction list no longer gets stuck showing its loading placeholder when switching to an account whose transaction list turns out identical to the previous one — in practice, switching between two accounts that both have no transactions, such as right after connecting a Keystone.
+- [MOB-1581] The Activity list now shows a sent transaction immediately after sending, regardless of how the send flow is closed, instead of waiting for the next sync cycle.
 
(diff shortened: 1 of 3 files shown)


LukasKorba (approved):
utACK

PR #1963, opened 2026-07-28
[MOB-1581] Fix transactions stuck at "Sending…"

pacu:
## Summary

A sent transaction could keep showing "Sending…" indefinitely, even after many confirmations. ZODL displays "Sending…" for exactly one reason: the transaction is a send and the SDK's locally stored `mined_height` is absent (`TransactionState.swift`). Confirmations are never consulted, so the row resolves only when ZODL re-reads the local transaction database — and every trigger for that re-read could be lost.

Restarting the app fixed the display, which is what pointed at the refresh pipeline rather than at scanning or the database.

## What was wrong

1. **Transaction events were filtered after throttling.** The raw synchronizer event stream was throttled with `latest: true` and only then filtered for `foundTransactions`/`minedTransaction`. An unrelated event (a connection-state change, stored UTXOs) arriving later in the same 200 ms window replaced the transaction event as the window's latest value, and the filter then discarded it — so the refresh never happened.
2. **Foregrounding did not restore transaction observation.** Backgrounding cancels the transaction subscriptions, but the foreground restart path re-registered only the general synchronizer state observation; `.observeTransactions` was dispatched exactly once, at app start. After a single background/foreground cycle, the list depended on incidental refreshes for the rest of the process lifetime.
3. **Nothing pull-based backstopped a lost push.** Once a signal was missed, nothing ever re-read the database again.
4. **The fetch swallowed errors and could race.** `try?` discarded every failure, and two overlapping fetches could complete out of order, letting a stale pre-mining snapshot overwrite a newer one.

## Changes

- Filter transaction events **before** throttling, in `RootTransactions` and `HomeStore`. Post-filter, every element that survives the throttle triggers the same fetch, so coalescing a burst is harmless.
- Re-dispatch `.observeTransactions` from the foreground `.retryStart` path, and cancel the event subscription and poller on backgrounding so teardown and restore are symmetric. Re-dispatch is safe because the inner effects are `.cancellable(cancelInFlight: true)` with stable ids, and the trailing fetch doubles as the catch-up for anything mined while backgrounded.
- Add a reconciliation poller: while any transaction is pending, re-read the local database every 30 seconds; stop when nothing is pending or the app backgrounds. This is a local SQLite read and an in-memory diff — no network request. It is managed before the state-equality early return, so an unchanged list still keeps the poller alive.
- Make the fetch effect self-replacing (`cancelInFlight`) and log failures through `LoggerProxy` instead of discarding them; the poller provides the retry.

A side benefit of the poller: the existing "chain tip passed expiry → show Failed" fallback in `TransactionState` finally gets a chance to fire for genuinely expired transactions, instead of the row sitting at "Sending…" because no fetch ever ran again.

## Tests

New `RootPendingTransactionRefreshTests` (Swift Testing, `DispatchQueue.test` scheduler, `PassthroughSubject`-backed event stream):

- a `foundTransactions` event sandwiched between unrelated events in one throttle window still triggers a fetch — this fails against the previous stream shape, where no number of retries could produce one;
- foregrounding after backgrounding restores transaction observation (events trigger fetches again, and the backgrounded window triggers none);
- the poller ticks while a transaction is pending, survives an unchanged still-pending result, and stops once nothing is pending.

`xcodebuild test -scheme zodl-internal` over `RootPendingTransactionRefreshTests`, `RootTransactionsAccountSwitchTests`, `RootSendCompletionRefreshTests` and `HomeTests`: 29 tests in 4 suites, all passing.

## Notes

This change is local-only and independent of the SDK: it does not add or alter any network request. Related SDK and librustzcash work is in flight separately; this can land on its own.

https://claude.ai/code/session_0128K6ptRPzgDWDzRUFAbjnC


schell:
As far as I can tell there are no UI changes so I can only assume that the e2e smoke tests are being flakey.

secant/Sources/Features/Root/RootTransactions.swift
```diff
@@ -132,12 +142,34 @@ extension Root {
...
+                // every 30 seconds until nothing is pending — a cheap SQLite read, no network.
+                // Managed on every completed fetch, including ones whose payload equals the current
+                // state, so an unchanged list keeps the poller alive.
+                let pendingTransactionsPoller: Effect<Root.Action>
+                if identifiedArray.contains(where: \.isPending) {
+                    pendingTransactionsPoller = .run { send in
+                        while !Task.isCancelled {
+                            try await mainQueue.sleep(for: .seconds(30))
+                            await send(.fetchTransactionsForTheSelectedAccount)
+                        }
```
schell, line 167: I think there's a case where this may poll forever.

Claude says: "the arm condition is identifiedArray.contains(where: \.isPending). TransactionState.isPending (TransactionState.swift:237) returns true for non-zcash types when swapStatus == .pending || .incomplete. Swap status is updated from userMetadataProvider.allSwaps() (external metadata), not the local SDK DB that the poller re-reads. So a swap stuck in .pending/.incomplete will poll every 30 s indefinitely — the local DB re-read can never resolve it."

schell: Oh, also this will _only_ trigger ever y30 seconds, not when the task is cancelled. That means the sleep could linger around for up to a little under 30s after the task is cancelled.

schell (approved):
Just had a couple notes. This seems fine after the tests pass.

pacu:
Note on the red E2E smoke tests: they are not caused by this PR. `build` and `unit_tests` both pass here; the `e2e-ios` phases are failing on every branch in the repo right now, including `michal/improve-fastlane` (07-29), `michal/MOB-1593`, `michal/MOB-1585`, `michal/MOB-1581` and `michal/MOB-1580`, with the same signature — the flows never reach a usable home screen, so assertions like `"Zodl" is visible`, `"Receive" is visible` and `Element not found: home.payButton` fail.

This change is also local-only: it alters when ZODL re-reads its own transaction database and does not add or modify any network request, so it cannot plausibly affect wallet restore or sync in those flows.

zcash-harry:
Let's please ignore the E2E tests for now, A lot of these tests are UI E2E, I would tackle them after all the design and new features settle

zcash-harry (approved):
utACK on E2E being red, I manually do the flows for now. They are all simple fixes but don't want to fix them everyday as the design changes

secant/Sources/Features/Root/RootTransactions.swift
pacu: You're right, and thanks for catching it — this would have run for the rest of the session with nothing it could settle.

Confirmed the mechanism: `isPending` reports the swap status for every non-`.zcash` type, and that status is owned by the swap provider's metadata — refreshed by `.autoUpdateCandidatesSwapDetails` in `RootSwaps` — not by the SDK database this poller re-reads. So an abandoned or stalled swap parked in `.pending`/`.incomplete` kept a 30-second timer alive against state the re-read could never change.

Fixed in 15f37964: the arm condition is now

```swift
if identifiedArray.contains(where: { $0.type == .zcash && $0.isPending })
```

which is exactly the set whose pending state (`minedHeight == nil`) a local re-read can resolve. The reasoning is in a comment at the call site so it doesn't get widened back later.

Covered by a new test, `swapPendingAloneDoesNotArmThePoller`: a swap-only pending list produces no fetch across several poll intervals.

Worth noting this doesn't weaken the original fix — a `.swapFromZec`/`.crossPay` row's displayed pending state comes from its swap status too, not from `minedHeight`, so nothing that the poller used to rescue stops being rescued.

pacu: I tested this one rather than reasoning about it, since it hinges on whether the scheduler's `sleep` observes cancellation.

New test `cancellingThePollerPreventsAnyLaterTick`: it arms the poller, lets it tick once so the effect is provably running, advances 10s into the next interval, then tears the poller down (via `didEnterBackground`) and advances well past the point where the in-flight sleep would have completed. No further fetch is dispatched, so the sleep does observe cancellation and no tick escapes it.

To be precise about what that does and doesn't show: it proves no *action* is dispatched after cancellation, which is the part that could have caused a spurious fetch against a torn-down synchronizer. If a work item lingers inside the underlying `DispatchQueue` until its deadline, that's invisible from the outside and harmless — but it's also not something this test would catch, so I don't want to claim more than it proves.

Test added in 15f37964 alongside the swap fix.

PR #1964, opened 2026-07-29
Submit appstore builds to App Review from the release tooling (--submit-review)

Chlup:
# Description

`Scripts/release.sh` gains `--submit-review` (appstore variant only): submit a build to App Review with no manual App Store Connect work. `--ref` becomes the "build it" signal — with it, the existing pipeline (preflight → tests → build → upload) runs and then submits; without it, a build already on App Store Connect is submitted as-is.

What submitting does:

- Creates the App Store version record, or **adopts one created manually** in any editable state (prepare-for-submission, developer-rejected, rejected, metadata-rejected, invalid-binary), renaming it if the version string differs (warned in the summary).
- **Copies promotional text from the live version** into any localization where it is empty — App Store Connect does not carry this field over to new versions, unlike the rest of the metadata. Manually entered promotional text is always kept, never overwritten.
- Writes **What's New** for every enabled App Store localization from the current checkout's `secant/Resources/WhatsNew/whatsNew*.json` (entry matching `--version`; locales are enumerated live from App Store Connect and mapped by language prefix, so en-US/es-ES/es-MX resolve to `whatsNew.json` + `whatsNew_es.json`).
- Attaches the requested build — keeping a correct one already attached, replacing a wrong one.
- Submits with **manual release** (you still press Release after approval), no IDFA, precheck without IAP.
- Errors out (fail-closed) when a version is already in review or approved-awaiting-release, when the requested version is already live, when the build is missing or still processing, or when any enabled locale lacks a What's New entry.
- After submitting, reads back promotional text / What's New per locale and warns loudly if promotional text is missing where a copy was planned (fixable in ASC without a new review).

`--dry-run` covers all of the submission checks read-only; `--yes` also skips the submit confirmation. Failure/success macOS notifications cover the submit path.

Implementation: pure decision logic in `fastlane/lib/zodl/whats_new.rb` and `fastlane/lib/zodl/submit_preflight.rb` (minitest-covered, mirroring the existing preflight pattern — 82 runs / 198 assertions green), wrapper arg surface bats-covered (10/10), Spaceship fact-gathering + a single `deliver` call in the `Fastfile`. The What's New renderer is parity-tested against the `/update-whatsnew` helper's output. Verified end-to-end with a read-only `--dry-run` against the real App Store Connect app (locale enumeration, build/version state detection, and correct refusal all confirmed live). The actual submission write path can only be exercised at the next real release.

Docs: `fastlane/README.md` and `docs/release-automation.md` document both usages, the convergence behavior, and new troubleshooting rows. No CHANGELOG entry — release tooling, not a user-facing app change.

### DIFF
--- a/Scripts/release.sh
+++ b/Scripts/release.sh
@@ -8,7 +8,9 @@ usage() {
 Usage: Scripts/release.sh --variant <v> --ref <ref> --version <X.Y.Z> --build <n> [options]
 
   --variant     internal | testnet | appstore | internal-testnet
-  --ref         branch, tag, or commit SHA to build
+  --submit-review  appstore only: submit the build to App Review after upload;
+                   without --ref, submits a build already on App Store Connect
+  --ref         branch, tag, or commit SHA to build (optional with --submit-review)
   --version     marketing version you intend to ship (X.Y.Z)
   --build       build number (integer)
   --dry-run     run all preflight checks, then stop before building
@@ -19,7 +21,7 @@ EOF
 }
 
 VARIANT="" ; REF="" ; VERSION="" ; BUILD=""
-DRY_RUN=false ; YES=false ; SKIP_TESTS=false
+DRY_RUN=false ; YES=false ; SKIP_TESTS=false ; SUBMIT_REVIEW=false
 
 while [[ $# -gt 0 ]]; do
   case "$1" in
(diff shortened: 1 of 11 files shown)


LukasKorba (approved):
utACK

PR #1968, opened 2026-08-04
ZIP 318 transaction labels: Home Activity + Transaction Detail

Chlup:
## Summary

Implements the approved "Transaction Statuses/Labels — Final Designs" (Figma, section `5864:82943` in [PR App Designs Q3 '26](https://www.figma.com/design/1aeq8gleYh9Yr1l33TwELR/PR-App-Designs-Q3--26?node-id=5864-82943)): ZIP 318 migration transactions are labeled by what they do, on the Home Activity list and the Transaction Detail screen, consuming the SDK surface introduced in zcash-swift-wallet-sdk PR #1927 (`zip318Kind`, `poolCrossingValue`).

| kind | in-progress (unmined) | completed (mined) | failed (expired) |
|---|---|---|---|
| `.transfer` | "Migrating..." / detail "Migrating" | "Migrated" | "Migration Failed" |
| `.preparation` | "Splitting Balance..." / detail "Splitting Balance" | "Balance Split" | "Split Failed" / detail "Balance Split Failed" |

- **Model** (`TransactionState`): `isMigrationTransaction`, title matrix (7 new `Localizable.xcstrings` keys, en+es), coins-swap glyph in all states (existing pending/failed icon colors carry the state), amount color stays primary even when failed, `displayedAmount` presents the moved value — `poolCrossingValue` (transfer) or `totalReceived` (preparation) — instead of the negated-fee balance delta, unsigned. The SDK's `poolCrossingValue` was previously dropped on the floor; it is now stored and consumed.
- **Views**: row and detail header switched to `displayedAmount` ("−" prefix suppressed for migration rows); migration details open expanded (the design's mock state), hide the message section and the collapse toggle, hide "Sent to" when there is no resolvable recipient (with shared `isSentToRowShown` keeping the card's corner rounding consistent), and no longer offer a Save-address form that could never be submitted.
- **Behavior change**: the canonical Activity build now **presents** stored-but-unmined migration rows as labeled in-flight history instead of filtering them out (supersedes the earlier hide-the-rows approach from this unreleased cycle — that behavior never shipped). The balance-breakdown "Pending" correction is unchanged: the same canonical pass still publishes the unmined-migration sum.

## Notes for reviewers

- **Failed rows persist by design**: an expired-never-broadcast transfer that the migration engine rebuilds leaves a permanent "Migration Failed" row next to the successful rebuilt one. The design explicitly commissions failed states (two of the eight Figma frames), so they are shown; if product prefers to suppress expired-unmined rows, it is a one-predicate change in `RootTransactions`.
- **App-wide subtitle change**: any pending transaction with no block time yet (regular sends included) now shows "Today" instead of a blank subtitle; failed unmined rows keep the blank. This matches the design and the See-all screen's existing "Today" grouping for such rows.
- Known follow-up (not in this PR): Activity search matches on the raw balance delta, not the displayed moved amount, for migration rows.

## Verification

- Full `zodlTests` gate: **1047 tests / 157 suites, exit 0** (baseline 1037 + 10 new model tests; one reworked Root test asserts rows are retained and the pending sum still publishes).
- TDD throughout: title/amount/color/glyph/daysAgo matrix in `TransactionStateTests`; red-first rework of the Root filter test (`RootTransactionsMigrationRowsTests`).
- SDK unchanged — everything needed is already public on `feature/ironwood-slipstream`.


LukasKorba (approved):
utACK

PR #1973, opened 2026-08-06
[MOB-1661] Fix migration pool balance accounting

zcash-harry:
## Summary

- stop treating proved-but-unbroadcast migration transactions as already stored in wallet pool balances
- display the complete Orchard balance during migration, including proposal-scoped advisory locks
- apply pool corrections symmetrically and clamp both sides by the same amount so pool totals cannot exceed the wallet total
- update regression coverage for proved, broadcast, mined, and stale/clamped balance states

## Root cause

The app-side pool correction assumed the SDK stored migration transactions at prove time. With the current SDK, proving only advisory-locks Orchard inputs; the wallet stores the transaction at the broadcast seam.

Consequently, correcting a proved transaction added its value back to Orchard even though that value was already present. The previous one-sided clamp could simultaneously reduce Ironwood to zero, producing either missing balance or surplus balance across Home and Migration Progress. Migration Progress also used the migration-unlocked Orchard view for display, hiding advisory-locked notes from every pool before broadcast.

## Impact

Home pool breakdown and Migration Progress now agree with the wallet total throughout the prove → broadcast → mined lifecycle. Proved transfers remain in Orchard until broadcast, and transient/stale reads cannot create value.

Linear: https://linear.app/zodl/issue/MOB-1661/the-balance-in-home-pool-breakdown-and-migration-progress-screen-pool

## Validation

- 17 focused tests passed in `MigrationPoolTruthTests` and `WalletBalancesPoolCorrectionTests`
- `zodl-testnet` built successfully from app revision `9f696fa2` with SDK revision `0b5e678c`
- rebuilt the exact SDK arm64 FFI slice and launched on a physical iPhone
- baseline physical-device pool breakdown summed exactly to the detailed wallet total (Home's shorter value differed only by display precision)
- `git diff --check` passed


zcash-harry:
Verified locally, self tACK

image
image

secant/Sources/Dependencies/MigrationManager/MigrationManagerLiveKey.swift
```diff
@@ -796,31 +796,35 @@ final class MigrationManagerImpl: @unchecked Sendable {
...
 
         // R13 refinement 1 — the loader's first canonical query: pools render as if every
-        // migration transaction not wallet-mined never happened. The SDK's figures count
-        // stored-unmined migration outputs (store-at-prove, DB-autopsy-pinned); the bubbles must
-        // not, or the header claims "migrated" over checkmarks that say otherwise. Both bubbles
-        // move by the same figure, so the flip to green and the value move land in the SAME
-        // derivation — the contradiction is structurally impossible, not carefully avoided.
+        // migration transaction not wallet-mined never happened. The current SDK enters a
+        // transaction into the wallet at the broadcast seam; proving only advisory-locks its
+        // inputs. Consequently only broadcast/mined-but-wallet-unsynced rows need correction.
```
nuttycom: I'm curious as to why "correction" is perceived to be needed in the first place. The wallet balance from `get_wallet_summary` should be correct at every point in the migration process, so the idea that a correction needs to be applied in the first place makes me suspicious.

zcash-harry:
Encountered another issue where "Sent" state was again causing discrepancy, but the Done state would show it correctly. fixed it here [b0f46da](https://github.com/zodl-inc/zodl-ios/pull/1973/commits/b0f46da65ba3542e6f8cd47c65b4ba513985aef2)

.github/workflows/e2e-smoke.yml
```diff
@@ -55,6 +55,19 @@ jobs:
...
         uses: actions/checkout@v4
 
+      # The Xcode project intentionally uses the SDK as a sibling local
+      # package (../zcash-swift-wallet-sdk). Check out the exact TestFlight
+      # SDK revision inside the workspace, then expose it at that path.
+      - name: Checkout SDK
+        uses: actions/checkout@v4
+        with:
+          repository: zcash/zcash-swift-wallet-sdk
+          ref: 0b5e678cedd8ded308ac131934f001448e69717c
```
nuttycom: These refs are fragile, and it will be very easy to use a different version of the sdk locally than is pinned by this revision, or to forget to update the pin. How can we make this more reliable?

.github/workflows/pull-request.yml
```diff
@@ -36,6 +36,19 @@ jobs:
...
           fetch-depth: 0
 
+      # The Xcode project intentionally uses the SDK as a sibling local
+      # package (../zcash-swift-wallet-sdk). Check out the exact TestFlight
+      # SDK revision inside the workspace, then expose it at that path.
+      - name: Checkout SDK
+        uses: actions/checkout@v4
+        with:
+          repository: zcash/zcash-swift-wallet-sdk
+          ref: 0b5e678cedd8ded308ac131934f001448e69717c
```
nuttycom: Same question here - this seems very fragile for github workflow yml.

nuttycom (approved):
utACK, great to see all these deletions! My only question is how can we better handle the revision pins in the github workflows.

zcash-harry:
> utACK, great to see all these deletions! My only question is how can we better handle the revision pins in the github workflows.

As mentioned over slack, will handle those as a follow up and keep this PR to the limited scope of fixing the balance.

secant/Sources/Dependencies/MigrationManager/MigrationManagerLiveKey.swift
Cosmos-Harry: Agreed. After tracing the pinned SDK/Rust transaction lifecycle, get_wallet_summary is the source of truth. The application-side in-flight correction was incorrect and has been removed. Both screens now display orchardBalance.total() and ironwoodBalance.total() directly from the SDK.

PR #1975, opened 2026-08-07
[MOB-1496] Behavior-based migration gate + advance-driven SDK adoption

nuttycom:
**Draft — work in motion.** Adopts the SDK's advance-driven migration rework (zcash/zcash-swift-wallet-sdk#1957) on the app side, in two stages:

1. **Landed: the app-side send buffer dies** (`2be16784`). `MigrationSendGate.waitUntil` was the mirror image of the SDK's deleted post-broadcast sync buffer — a fixed 600s/180s delay measured from `migrationLastSyncCompletedAt`, i.e. the same correlation-signature fingerprint the SDK's ruling removed, pointing the other direction. Deleted rather than pinned app-side; `MigrationFastLane`, the tick fast-path, and the buffer clamps collapsed with it. What remains is behavioral: `broadcastOneTransfer` still stops a running sync before submitting, and the SDK still refuses a start while a submission is in flight.
2. **In flight: the compile-mandatory shim adaptation.** The `SDKSynchronizer` dependency client still declares the four members SDK Task 4.4 removed (`MigrationTransferAttempt`, `executeNextPendingMigrationTransfer`, `pendingMigrationTransferProposal`, `finalizeReadyMigrationTransfers`) — the current build failure is entirely that pre-existing gap, not the gate change (which type-checks clean). The adaptation deletes those, adds the instruction-executor members (`proveMigrationTransactions` / `performMigrationBroadcast`), reworks the submit sites and `runProveSweep` onto step-payload dispatch, and deletes the AUD-1 estimated-tip tiebreaker (compile-forced — its query is gone; the FIND-5 deadlock class closes at the root).

**Deliberately left for team decisions** (see `.plans/SDK-ADVANCE-DRIVEN-MIGRATION-HANDOFF.md` in the working tree): session policy (broadcast-during-sync dispatch), user-intent sync-unlock, and the now-vestigial gate-attribution probe.


secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerLive.swift
```diff
@@ -125,11 +125,11 @@ extension SDKSynchronizerClient: DependencyKey {
             signAndStoreMigrationSchedule: { accountUUID, schedule, usk in
                 try await synchronizer.signAndStoreMigrationSchedule(accountUUID: accountUUID, schedule, usk: usk)
             },
-            executeNextPendingMigrationTransfer: { accountUUID, options, useEstimatedTip in
-                try await synchronizer.executeNextPendingMigrationTransfer(
+            performMigrationBroadcast: { accountUUID, instruction, options in
```
nullcopy, line 128: > This closure doesn't acquire the transaction guard, but the interface doc (updated in this PR) says "Broadcast-bearing: guarded by the transaction guard in the LiveKey", and CLAUDE.md requires every broadcast-sensitive closure to take transactionGuard.withSubmission in its LiveKey — as createAndSubmitProposedTransactions, createAndSubmitTransactionFromPCZT, and the new submitMigrationPreparation all do.
>
> Without it, while a migration transfer broadcast is on the wire, the automatic server switch's switchIfIdle sees the guard idle and runs switchTo(endpoint:), tearing down the synchronizer under the in-flight submission — the exact overlap the guard exists to prevent. As it stands, preparations are protected but the more sensitive pool-crossing transfers are not. Can we wrap this in withSubmission like its siblings?

secant/Sources/Dependencies/MigrationManager/MigrationStepDriver.swift
```diff
@@ -487,14 +449,145 @@ extension MigrationManagerImpl {
...
+        from result: SDKSynchronizerClient.CreateProposedTransactionsResult
+    ) -> MigrationTransferResult {
+        switch result {
+        case let SDKSynchronizerClient.CreateProposedTransactionsResult.success(txIds):
+            return MigrationTransferResult.success(txId: txIds.first ?? "")
+        case let SDKSynchronizerClient.CreateProposedTransactionsResult.partial(txIds, _):
+            // Unreachable for the single transaction this lane submits (`.partial` needs both an
+            // acceptance and a failure), but an acceptance is an acceptance.
+            return MigrationTransferResult.success(txId: txIds.first ?? "")
+        case SDKSynchronizerClient.CreateProposedTransactionsResult.failure,
```
nullcopy: > This maps every non-acceptance — including a permanent validation rejection carried in .failure(code:description:) — to .networkError(retryable: true) (line 579), and nothing gets recorded into the engine. The deleted D2 delivery path classified outcomes through MigrationBroadcastFailureClass and recorded the real MigrationTransferResult (.invalidNote, .expired, …).
>
> Concretely: if the server permanently rejects a note-split preparation (consensus/validation failure, not transport), the engine is never told, so every prove pass and 30-second tick re-takes and re-submits the same doomed transaction until ZIP 203 expiry. The user sits in the split phase for hours with no failure sheet or attention state. Permanent rejections need to be classified and recorded like the old lane did.

secant/Sources/Dependencies/MigrationManager/MigrationStepDriver.swift
```diff
@@ -487,14 +449,145 @@ extension MigrationManagerImpl {
...
+    ///
+    /// NO FAILURE-CLASS ROUTING. `broadcastOneTransfer` runs a non-success through
+    /// `MigrationBroadcastFailureClass` for its PERSISTENT effects — the Tor-hold indicator, the
+    /// pending-prompt latch — because that lane submits under the run's pinned privacy options and
+    /// a Tor failure there is a fact about the user's chosen transport. This lane submits over the
+    /// app's ordinary connection with no privacy options at all, so it has no Tor verdict to
+    /// record and no held prompt to raise; what it sees is transport success or transport failure,
+    /// and it reports exactly that. Routing a plain server rejection into the Tor UX would put a
+    /// privacy claim behind an outcome that made none.
+    private func submitProvedPreparations(accountUUID: AccountUUID, txids: [Data]) async {
```
nullcopy, line 500: > This lane submits proved preparations (called from the prove discharge at line 634) without setting any of the in-flight markers the replaced D2 same-pass delivery set via runBroadcastSession — setMigrationWorkInFlight, broadcastsInFlight, activeBroadcastTxIds, and the preparationBroadcastsInFlight keep-open banner map — so the app reads idle while a preparation broadcast is on the wire.
>
> During the submit window (afterSync edge or 30s tick), isPreparationBroadcastInFlight and isBroadcastInFlight both read false — no keep-open/"splitting balance" banner — and the re-entry route's isMigrationWorkInFlight short-circuit doesn't engage. A user who backgrounds or kills the app mid-broadcast stalls the split until a later pass re-proves it. These markers should be set around the submit, same as the old session did.

secant/Sources/Features/Migration/MigrationSending/MigrationSendingStore.swift
```diff
@@ -388,76 +414,34 @@ struct MigrationSending {
...
-                }
+                // MOB-1513: the immediate lane's success came from
+                // `createAndSubmitProposedTransactions` (the ordinary-send submission path), which
+                // never touches the ENGINE's own migration-sync privacy gate at all — unlike the
+                // engine lanes, nothing else will ever prompt `RootInitialization`'s
+                // resume-once-clear machinery to re-check, so this stop needs an explicit nudge
+                // just as a non-success outcome does. (The engine-lane branch that skipped the
+                // nudge here went with the scheduled-run delivery arm; this lane always nudges.)
+                await migrationManager.refreshMigrationSyncGate()
             } catch ZcashError.migrationRecordFailedAfterBroadcast(_) {
```
nullcopy: > This arm still skips the refreshMigrationSyncGate() nudge ("no nudge either"), but its justification left with the deleted scheduled-run delivery arm — the success arm's own comment now says so explicitly ("the engine-lane branch that skipped the nudge here went with the scheduled-run delivery arm; this lane always nudges") and always nudges at line 444, and the general catch nudges too.
>
> If this arm is hit (broadcast landed, recording failed), stopSyncBeforeMigrationBroadcast() has already run and no path restarts sync: the synchronizer stays stopped for the rest of the foreground session and balances/transactions silently stop updating until relaunch. Either this arm needs the same nudge as the success path, or — if it's now unreachable from commitImmediateSoftware — it's dead code documenting a lane that no longer exists and should go.

secant/Sources/Dependencies/MigrationManager/MigrationStepDriver.swift
```diff
@@ -487,14 +449,145 @@ extension MigrationManagerImpl {
     /// One action, executed. The switch is exhaustive over `MigrationStepAction` (I1) — every case
     /// the planner can produce has a body here, and adding a case to either breaks the build.
     // swiftlint:disable:next cyclomatic_complexity
```
nullcopy: > proofBudget(for:) (line 459) was inserted between this directive (plus execute's doc comment) and the execute(...) it targeted. The disable now suppresses nothing on the trivial helper, while execute — the ~10-case switch the directive was added for — lints unsuppressed, so every build emits a cyclomatic_complexity violation from the lint build phase (and fails outright anywhere violations are errors). Quick Help also now shows execute's doc on proofBudget. Moving the helper above the doc comment + directive fixes both.

nullcopy (commented):
AI assisted review, as of 9cdcead6305b2e9d6f8c7370ff95e0ec50292709

secant/Sources/Dependencies/SDKSynchronizer/SDKSynchronizerLive.swift
nuttycom: Fixed in ddb62c62 — the shim now takes `transactionGuard.withSubmission` exactly like its named siblings. Nesting check: the only production caller is `MigrationManagerImpl.broadcastOneTransfer`, which holds no guard, so wrapping introduces no re-entrancy hazard.

*Filed with Claude Code assistance.*

secant/Sources/Dependencies/MigrationManager/MigrationStepDriver.swift
nuttycom: Fixed in 52bca933. `.failure` is indeed a server verdict (`mapSubmissionOutcomes` produces it only from a `.rejected` outcome), so it is now classified by the same rules the SDK's own delivery ceremony applies (`MigrationBroadcaster.map`, constants mirrored and attributed since that type is SDK-internal): duplicate re-submission → success, expiry fragments → `.expired`, any other rejection → `.invalidNote`. Permanent rejections are recorded on the engine via `recordMigrationPreparationBroadcast` — the rust record path handles the rejection tags with `report_broadcast_failure` on the still-`Proved` row, so the next crank re-adjudicates and can raise attention instead of re-offering the doomed row until ZIP 203 expiry. `.grpcFailure` (transport — no verdict exists) keeps the retryable network error with no record, per the engine's contract.

*Filed with Claude Code assistance.*

secant/Sources/Dependencies/MigrationManager/MigrationStepDriver.swift
nuttycom: Fixed in 76d1fb46 — the submit-to-record span now runs inside a new `MigrationManagerImpl.withPreparationBroadcastMarkers(accountUUID:id:_:)`, which brackets all four markers (`broadcastsInFlight`, `activeBroadcastTxIds`, `preparationBroadcastsInFlight`, `setMigrationWorkInFlight`) in `runBroadcastSession`'s own order, including its `broadcastsInFlight` re-entrancy guard — an account already mid-broadcast skips the row and the engine re-offers it on the next crank — with a `defer` clearing every exit path. A new test pins `isMigrationWorkInFlight` true inside the mocked submit and false after the pass.

*Filed with Claude Code assistance.*

secant/Sources/Features/Migration/MigrationSending/MigrationSendingStore.swift
nuttycom: Deleted in c9e8acc2, with the trace in the header doc: the SDK's sole throw site of `migrationRecordFailedAfterBroadcast` is `OrchardMigration.broadcastAndRecord`, reachable only through `performMigrationBroadcast`/`submitNoteSplit`, and the immediate pipeline calls neither — `commitImmediateSoftware` throws only from `createAndSubmitProposedTransactions` or its own `immediateSubmitNotSuccessful`, and `recordImmediateMigrationBestEffort` catches internally. Were the error ever thrown again, the generic catch handles it soundly: `classify(error:)` maps it to no route, and the nudge fires because sync was stopped.

*Filed with Claude Code assistance.*

secant/Sources/Dependencies/MigrationManager/MigrationStepDriver.swift
nuttycom: Fixed in b9aeb925. By the time of the fix, three helpers had accumulated between the directive and `execute` (`proofBudget`, then `submitProvedPreparations`/`transferResult` from 8e2deb38), so the doc comment + `swiftlint:disable:next` were moved down to sit immediately above `execute` rather than moving one helper up — same pure reordering, same effect. The cyclomatic violation and the superfluous-disable warning are both gone; the remaining warning on `advance` (:149) is pre-existing at baseline.

*Filed with Claude Code assistance.*

PR #1981, opened 2026-08-10
Run E2E against the integrated Slipstream stack

zcash-harry:
## What changed

- Check out the exact Slipstream Swift SDK revision used by the integrated app branch.
- Build and cache matching Rust FFI artifacts before resolving the app packages.
- Retry Xcode package resolution after its observed internal package-registration crash.
- Include SDK source inputs in the Swift package and app cache keys.
- Start the Phase 2 wallet database cache at `v3`, give reruns unique keys, and never save an empty cache.

## Why

This runs the iOS E2E suite against the complete Slipstream app and SDK stack. The first Phase 2 run is intentionally a cold restore so CI can verify the expected restore-time improvement without an old database masking the result.

## Validation

- Workflow and composite-action YAML parses successfully.
- `git diff --check` passes.
- Full build and E2E validation will run on this PR.


zcash-harry:
Superseded by #2044, which carries the identical change set (same 8 files, 184/51) onto `maint/3.10.x` so the e2e fixes reach the existing 3.10.x builds directly.

The branch `agent/slipstream-e2e` is left intact if we need to reopen and retarget this at `main`.

PR #1983, opened 2026-08-11
[MOB-1630] Don't offer migration below the 0.01 ZEC/TAZ floor

Chlup:
## Summary

The migration banner's fresh-offer decision (`MigrationDerivations.bannerVariant`, `.notStarted` arm) compared the unlocked Orchard balance against **zero**, so any dust balance raised "Migration Required". The engine can never serve that offer: below ZIP 318's `MAX_RESIDUAL_VALUE` (0.01 ZEC) no migratable note can be formed and `proposeMigrationTransfers` answers an empty schedule — its documented "nothing to migrate". The user-visible result was a permanent "Migration Required" banner whose tap ran the plan screen's ~60 s empty-schedule retry and ended on the propose-failure sheet, every time.

The offer now requires at least **0.01 ZEC/TAZ**, introduced as the named constant `MigrationDerivations.minimumOfferableOrchardBalance` (`Zatoshi(1_000_000)`, ZIP 318 `MAX_RESIDUAL_VALUE`).

Deliberately minimal:
- Only the fresh `.notStarted` offer is gated. Running-state arms keep rendering through balance changes, and the post-completion re-offer stays gated on the engine's own answer (`isMigrationRemainderPending`), as before.
- Known remaining sliver, out of scope here: a balance in roughly [0.01, 0.01 + ZIP-317 fee buffer) still offers and still cannot be planned (fees make it unplannable). The exact floor is only known to the engine; gating on an engine peek is a follow-up option.

## Test plan

- New `MigrationAdvanceStepBannerTests` cases (written first, watched fail):
  - `noRunWithOnlyDustShowsNothing` — no run + `Zatoshi(999_999)` → no banner (failed before the fix with `.required`, passes after)
  - `noRunAtTheOfferFloorOffersTheMigration` — no run + `Zatoshi(1_000_000)` → `.required` (pins the boundary from the other side)
- Full `zodlTests` target: 1158 tests in 175 suites passed (`xcodebuild test`, scheme `zodl-internal`, iPhone 17 Pro simulator). The single "known issue" is the pre-existing, explicitly-marked `CurrencyConversionSetupTests` expectation (MOB-1364 gap), unrelated to this change.


LukasKorba (approved):
utACK

PR #1984, opened 2026-08-11
Add print_notifs debug command listing pending migration notifications

Chlup:
## Summary

The hidden database debug screen (Settings → What's New → long-press the logo) now accepts a special `print_notifs` query. Typing it (surrounding whitespace ignored, case-insensitive) and tapping Execute prints the pending Orchard → Ironwood migration local notifications into the output field instead of executing SQL:

- notification authorization status first (a denied/undetermined status silently no-ops every arm, so it's the first thing to rule out when a poke never arrives),
- then one block per pending request, soonest first: full identifier, absolute + relative fire time, account hex, title and body,
- `No pending migration notifications.` when nothing is armed.

Every other query behaves exactly as before — the SQL path still receives the original untrimmed query, and the pre-existing Face ID gate on Execute covers the new command unchanged.

## Implementation

- New `PendingMigrationNotification` value model + pure, table-tested `debugReport` formatter (`Dependencies/UserNotifications/`), same pure-layer split as `MigrationNotification`.
- New read-back member `pendingMigrationNotifications` on `UserNotificationsClient`; the LiveKey maps `UNUserNotificationCenter` pending requests filtered by the `migration.` prefix (read-only, no sorting — ordering is the tested formatter's job). Adding the member required threading the new parameter through 7 explicit client constructions in `MigrationTickDriverTests`.
- `WhatsNew` reducer branches on the command inside `.executeQuery` and delivers the report via a new `notifsReportReady` action.

## Test plan

- New `PendingMigrationNotificationTests` table suite (7 tests): exact report rendering incl. sorting, nil fire date/account placeholders, overdue/relative-time boundaries, authorization variants.
- `WhatsNewTests` additions: `print_notifs` routes to the report and never calls `debugDatabaseSql` (spy-verified); whitespace/case-variant input; existing SQL-path regression test untouched and green.
- Full `zodlTests` bundle green on iPhone 17 simulator at branch head (only pre-existing known-issue expectation in `CurrencyConversionSetupTests` reported, unrelated to this change).

Note for the merger: commit `f789d373` fixes test-target compilation introduced by its parent, so a squash merge closes the one-commit test-bundle bisect gap; a regular merge is fine too.


LukasKorba (approved):
utACK

PR #1985, opened 2026-08-11
Add Max amount button to Send, Pay and Swap (internal and testnet only)

Chlup:
## Summary

Adds a **Max** chip to the amount fields on the Send, Swap and Pay screens that fills in the largest amount the user can send right now. The amount is computed by the SDK's send-max proposal (`Synchronizer.proposeSendMax`), so the ZIP-317 fee is exact for the actual recipient rather than estimated.

The button ships in **internal and testnet builds only** — the production / App Store build is unchanged.

## What's included

- **`SDKSynchronizerClient.sendMaxAmount`** wraps `proposeSendMax` and returns `totalSpendValue - totalFeeRequired`. It's read-only, so it does not take the transaction guard.
- **`ZashiMaxChip`**, a shared UI component whose `Style` pairs each design variant's corner radius with its label casing so the two designs cannot be mixed.
- **Send:** the button enables once the typed address is valid, and the max is proposed against that address so the fee matches the recipient. The fill routes through `zecAmountUpdated` to keep the fiat field in sync.
- **Swap and Pay:** the provider deposit address doesn't exist until after a quote, so the max is proposed against the account's own transparent receiver, whose fee class matches the transparent deposit address that will be used. Swap fills ZEC (or USD when the input is toggled); Pay converts to the target token and mirrors the USD and canonical fields.
- **Rounding is always down, never to nearest** — rounding up would put the fill above the true maximum and make the user's own Max unsendable.

## Build gating

The chip is gated on `ZODL_INTERNAL || SECANT_TESTNET`, which follows the *target* rather than the build configuration (a new `ZODL_INTERNAL` condition is set on all three `zodl-internal` configurations). Gating on `!SECANT_DISTRIB` was wrong: that flag exists only on `zodl-production`'s Release-AppStore configuration, so Debug and Release-Testflight builds of the production target showed the button.

| Target | Max button |
|---|---|
| `zodl-internal` | visible (every configuration) |
| `zodl-testnet` | visible (every configuration) |
| `zodl-production` | hidden (every configuration) |

The reducers stay compiled in all targets so tests cover them; only the chips are conditionally compiled out.

## Testing

- New Swift Testing suites: `SendFormMaxButtonTests` and `SwapAndPayMaxButtonTests` (528 lines) covering the fill paths, rounding, and enablement.
- Verified by running each target in the simulator: the chip is present on Send, Pay and Swap in the internal and testnet builds, and absent from all three in the production build.


LukasKorba (approved):
utACK

PR #1987, opened 2026-08-12
[MOB-1697] Warn on the send review screen when a send must spend Orchard funds

Chlup:
## Summary

Re-lands the Orchard-spend warning from the closed PR #1967 on top of main, replacing the migration-gated variant that arrived with the MOB-1466 migration work. Behavior now matches #1967 exactly:

- When a send review screen appears and the built proposal spends legacy Orchard funds (`Proposal.spendsLegacyOrchardFunds`), a warning bottom sheet recommends migrating first. It shows once per confirmation-screen instance, on both `SendConfirmationView` and `RequestPaymentConfirmationView`, driven by a dedicated `confirmationScreenAppeared` action so screens that share the reducer without attaching the sheet (the SwapAndPay Keystone path) can never trip or burn the one-shot latch.
- **Continue anyway** (outline destructive) dismisses the sheet; the user stays on the review screen. Swipe-down behaves the same.
- **Cancel** (primary) dismisses the sheet and then navigates back to the send form — the pop is deferred until the sheet's real `onDismiss`, and it reuses `.cancelTapped`, the exact case each coordinator's existing back-button handling already covers (in the scan flow that pops past a stale `.scan` element back to the form).
- The warning is no longer gated on a live migration run — it fires for any Orchard-spending send. The tap-time sheet (`SendOrchardWarningSheet`), `MigrationManagerClient.shouldWarnBeforeManualSend`, `MigrationManualSendRisk`, and their tests are removed as superseded, and `.sendTapped` returns to authenticating directly.

Also fixes a bug on main: `RequestPaymentConfirmationView` ran the tap-time risk check through the shared reducer but never attached the warning sheet, so when the warning condition fired on the ZIP-321 request-payment review, the Send button silently did nothing. With presentation moved to screen-appear and the sheet attached on both screens, that dead end is gone.

## Testing

- Ported `SendConfirmationOrchardWarningTests` (7 tests): presentation for orchard/non-orchard/nil proposals, the one-shot latch, an `.onAppear`-alone regression, and continue vs. cancel vs. bare-dismissal semantics.
- Ported the `ScanCoordFlowZip321Tests` regression: cancelling from the request-payment confirmation pops past a stale scan screen back to the send form (documents the pre-existing `coordinatorReduce()`-before-`forEach` TCA warning via `withKnownIssue`).
- Full `zodlTests` suite: 1177 tests, 0 failures.


LukasKorba (approved):
utACK

PR #1996, opened 2026-08-18
[MOB-1678] Use resilient voting config pin

czarcas7ic:
Before, the bundled production voting trust anchor fetched directly from raw GitHub, making configuration loading depend on GitHub availability.

After, it uses the immutable voting.valargroup.dev production pin. The gateway reads GitHub normally and serves the automatically published Cloudflare copy during a GitHub outage, while ZODL still enforces the embedded SHA-256. ZODL has no separate bundled stage anchor, so this PR intentionally changes only production.


LukasKorba (approved):
utACK

PR #2001, opened 2026-08-20
[MOB-1755] Prevent zero-balance shielding prompts

zcash-harry:
## Summary

The shielding offer (smart-banner priority 7) and the Balances Shield button run on one source of truth and can no longer contradict the wallet's actual transparent balance — and the banner arbiter around the offer is hardened against the races that used to lose or duplicate banners.

**Offer correctness**

- The banner's offered amount is always the SDK-reported transparent balance, and the offer/retract decision is made synchronously against that same value — the offer cannot display one figure while deciding on another. The ladder's balance fetch is account-scoped and cancellable, so a result in flight across an account switch cannot poison the newly selected account's offer.
- A nil shielding proposal is a first-class "Nothing to Shield" outcome: a plain titled alert (en + es) instead of a shielding-error code with a support-report prompt. The spending key is derived only after a proposal exists.
- Shielding eligibility (balance >= threshold, per the SDK contract) lives in one shared predicate used by the banner, the Shield button, and the zero-balance states of both balance screens; the 2-days / 2-weeks / month reminder ladder is likewise one shared rule for the shielding and wallet-backup lanes.

**Offer lifecycle**

- "Remind me later" is restored on the shielding help sheet — the button was lost in an earlier restructuring, leaving its handler with no sender, so dismissing never actually stored a reminder. It now stores the phased reminder, every path that can raise the offer honors it, and while a reminder is pending the ladder hands the slot to the next lane instead of ending the pass.
- A transparent deposit crossing the threshold mid-sync is latched and offered once sync completes. The latch means "an offer is owed": declines for reasons that expire (shield still pending, reminder not yet due) keep it armed for later ticks, and a higher-priority banner (e.g. syncing) displacing the seated offer re-arms it — the offer returns once the wallet is back up to date.
- Terminal shielding outcomes (success, nothing-to-shield, Keystone proposal hand-off) retract a seated offer immediately and clear a merely-latched request without tearing down the banner actually holding the slot. The banner keeps observing the shielding processor while Home is covered by the Send/Pay flow, so a shield finished off-screen retracts the stale offer too; the processor still resets its replayed state after every terminal outcome, so re-subscribers (Root re-init) cannot replay a stale success, alert, or Keystone signing launch, while an in-flight `requested` still replays for a sheet opened mid-shield.

**Arbiter robustness**

- Every seat bumps a generation counter and clean closes deferred through the retraction path are generation-guarded: a retraction aimed at an old banner can no longer wipe a banner (sync-error included) seated on the same tick, a lane seating over an already-retracted "zombie" seat no longer re-arms the offer latch, and the ladder's own retraction closes synchronously so its walk-down deterministically seats the next lane.
- A stale scheduled open can no longer expand an empty banner shell; an incumbent collapsed to make room for a request that dies mid-swap re-opens instead of stranding collapsed; and seat-time re-validation of a surviving shielding request also rejects offers whose funds are already being shielded.

## Testing

- Full `zodlTests` bundle: 1342 tests / 195 suites passing (`xcodebuild test-without-building`, iPhone 17 / iOS 26.5 simulator).
- New/extended coverage: `SmartBannerShieldingOfferLifecycleTests` (16 tests: same-tick retraction vs. error seat with the banner closed and open, zombie-seat displacement, reseat-collapse recovery incl. the producer flag, displacement re-arm, latch survival through declines, cross-account fetch drop, off-Home terminal retraction, reminder round-trip, empty-shell guard), `SmartBannerShieldingZeroBalanceTests` (crossing-during-sync, walk-down, reminder phases, account switch, clean-close survival rules), `ShieldingProcessorLiveKeyTests` (terminal replay reset, propose-before-derive), `ShieldingEligibilityTests` + `WalletBalancesEligibilityTests` (inclusive threshold boundary on both balance screens), and nothing-to-shield paths in `BalancesTests` / `RootShieldingAlertTests`. The shielding suites pin their in-memory shared storage so parallel suites cannot leak state into each other.

Linear: https://linear.app/zodl/issue/MOB-1755/prompt-to-shield-zero-balance-after-deposit


zcash-harry:
https://github.com/user-attachments/assets/324e13ef-e5d3-474d-9b7e-14ed83a358de

Chlup (approved):

Issue #16, opened 2021-08-26
Request ZEC Scaffold


pacu:

geffenz:
Assigning to myself to add screen examples.

geffenz:
![image.png](https://images.zenhubusercontent.com/5ddc49a1bc69ec000196b9a0/23fecb91-34eb-4f6d-a871-bf382fd4cb6d)
This includes an address book screen that might not be part of MVP.

PR #2002, opened 2026-08-20
[MOB-1751] Parse and interpret cross-chain payment URIs

zcash-harry:
## Summary

Scan a payment request QR on the Pay screen, not only a plain address.

- Parse Bitcoin, Litecoin, EIP-681 and Solana payment URIs through one Swift SDK parser.
- Fill in the recipient, the asset and the requested amount.
- Apply a request as a whole, or not at all. Never leave the three disagreeing.
- Report every rejection. None are silent.

## What it accepts and refuses

| Input | Result |
|-------|--------|
| `bitcoin:`, `litecoin:` mainnet | Accepted |
| `ethereum:` native and ERC-20 | Accepted |
| `solana:` transfer, native and SPL | Accepted |
| Testnet or regtest address | Refused |
| Solana interactive transaction link | Refused |
| ENS name or wrong-length hex recipient | Refused |
| Asset the app cannot pay | Refused, form unchanged, toast |
| Amount below 8 decimal places | Refused, toast |

Chain-less EIP-681 requests resolve to Ethereum mainnet. The app has no network setting, so the previous fallback made the same QR mean different assets depending on an earlier, unrelated swap.

Amounts are floored at 8 decimal places, never rounded up. The prefilled string is the amount that gets paid, so rounding up could set the user up to overpay.

## Dependencies

| Dependency | State |
|------------|-------|
| [librustzcash#2989](https://github.com/zcash/librustzcash/pull/2989) | Merged 22 Aug |
| [zodl-swift-wallet-sdk#16](https://github.com/zodl-inc/zodl-swift-wallet-sdk/pull/16) | Open |

Upstream bugs found in the shared parser are tracked in [librustzcash#3026](https://github.com/zcash/librustzcash/issues/3026). The Android SDK shipped with them. This SDK behaves the same.

## Revert before merge

`22e35afe` is temporary. It points the project at a local SDK package so the FFI is built from source. Without it the branch does not link, because no published libzcashlc release contains `zcashlc_payment_uri_parse`.

To merge:

1. Merge SDK #16.
2. Cut a libzcashlc release that contains `payment_uri`.
3. Revert `22e35afe`. Restore the remote package reference and empty `.sdk-pin`.

`setup-sdk-companion.sh` already warns that this pin lives only on a deletable PR branch.

`6c2e3df9` also does not belong here. It fixes an unrelated break on `main` (see the review thread). It is isolated so it cherry-picks cleanly.

## Testing

- Full unit suite green: **1644 passed, 0 failed, 1 skipped, 2 expected failures of 1647**.
- 29 CrossPay tests, run against the real Rust FFI.
- Money-critical resolution logic also verified standalone, by execution.

Caveat: CI pins Xcode 26.3 for the Swift 6.0 concurrency diagnostics. The local run used 26.5, which is more permissive.

---

AI disclosure: implemented with OpenAI Codex assistance. Review feedback addressed with Claude.


zcash-harry:
**Follow-up (not blocking):** `evmChain`/`nativeTokens` in `CrossPayRequest.swift` are hand-duplicated in the Android app's `CrossPayRequest.kt` (same chain-id→ticker map, same native-token aliases). A chain added to one side and forgotten on the other fails silently rather than at build time.

Suggest collapsing this into one shared, generated or config-driven source: options include a synced JSON resource with a pinned-hash guard in each repo, generated native constants from one canonical file, or eventually surfacing it from the live NEAR/curated-asset feed instead of hardcoding it at all. Flagging for a deliberate choice among those trade-offs in a follow-up, not scoping it here.

secant/Sources/Features/SwapAndPayForm/SwapAndPayStore.swift
```diff
@@ -1554,6 +1554,29 @@ extension SwapAndPay.State {
...
+        guard !isSwapExperienceEnabled, !isSwapToZecExperienceEnabled else { return }
+
+        selectedContact = nil
+        selectedAsset = request.resolveAsset(in: swapAssets, current: selectedAsset)
+        amountAssetText = ""
+        amountUsdText = ""
+        amountText = ""
+
+        guard let amount = request.resolvedAmount(for: selectedAsset),
+              let formatted = conversionCrossPayFormatter.string(from: NSDecimalNumber(decimal: amount)) else { return }
```
pacu: hmmm I've burned with this before 😅

This doesn't account for locales. I understand that the payment request uses en-us locale for numbers. but it's not the case for the user's textfields. So this will could have localization bugs hidden

pacu (changes requested):
Please check for localization bugs that may arise from Payment URI being en-us and users having different locales on their OS.

Chlup (changes requested):
Thorough review of the cross-chain payment-URI parsing. The overall shape is good — delegating protocol parsing to the Rust crate and keeping a thin typed adapter in the app is the right call, and the `.solanaTransaction` / `.unrecognised` rejections are correctly placed. But the adapter and the form-application step have a number of real defects, several of which change or misdirect money, so I'd like these addressed before merge.

Two notes before the list:

- **The branch does not currently build.** `xcodebuild build-for-testing -scheme zodl-internal` fails with `cannot find 'zcashlc_payment_uri_parse' in scope`. The symbol is absent from both slices of the vendored `libzcashlc.xcframework` and from its headers, which matches the still-open librustzcash dependency. So this is blocked on merge order, and my review below is static analysis plus standalone Swift/Rust replication of the logic rather than a compiled run.
- **Needs a rebase.** `SwapAsset.swift` and `SwapAndPayStore.swift` have both moved on `main` since the merge base (~192 lines in the store), including a refactor right around `conversionCrossPayFormatter`. `main` also gained a `Decimal.roundedDown(scale:)` helper that is directly relevant to item 6.

---

## Blocking

**1. `SwapAndPayStore.swift:1563` — a scanned QR overwrites the user's Swap-to-ZEC refund address.**
`address = request.address` runs *before* the `guard !isSwapExperienceEnabled, !isSwapToZecExperienceEnabled` on the next line. In Swap-to-ZEC that field is the user's own refund address: `addressView()` binds `$store.address`, is labelled `swapToZecRefundAddress`, and hosts the scan button; `getQuote` sets `destination = state.address` and `Near1Click.swift:333` maps `refundTo: isSwapToZec ? destination : refundTo`. So scanning a `bitcoin:` payment request while setting up a BTC→ZEC swap silently replaces the refund address with the payment request's recipient, and a refunded swap (`SwapConstants.refunded`) sends the deposit to that third party. This PR escalates the severity rather than causing it outright: previously the raw URI string landed there and the provider would almost certainly reject it; now a clean, valid, acceptable address does. The guard should come first, or the swap modes should be routed separately by the coordinator.

**2. `CrossPayRequest.swift:41` — a chain-less EIP-681 native request resolves to the wrong asset entirely.**
With no `@chainId`, the fallback is the currently-selected asset's chain, and `nativeTokens`' permissive `default: [chain.lowercased()]` then matches that chain's own native token. A user whose last payout was AVAX (restored from `lastUsedAssetHistory`) scans the ordinary `ethereum:0x…?value=1e18` and gets **1 AVAX** pre-filled instead of 1 ETH — roughly a 100× per-unit value difference. Same for BNB@bsc. Worse, `current = USDT@bsc` resolves to **BNB@bsc**: `current` is not even among the candidates, so it falls through to the `count == 1` rule and manufactures a native token out of a non-native selection. `evmNativeRequestWithoutChainIdStillFallsBackToACurrentEvmAsset` uses ETH@arb, where the symbol coincidentally matches, so it cannot catch this.

I know ERC-681 says "if no `chain_id` is present, the client's current network setting remains effective", and I initially thought that defended the design — but this app has no EVM network setting. `selectedAsset` is a swap-payout preference left over from an unrelated earlier swap, so treating it as the current network invents one. Returning `nil` (make the user pick) or pinning chain-less native requests to `eth` both seem defensible; silently substituting the asset does not.

**3. `CrossPayRequest.swift:137` — testnet/regtest Bitcoin and Litecoin URIs are accepted as mainnet.**
`UTXOPaymentURIRequest.network` is decoded by the SDK precisely so callers can reject non-mainnet requests, and the parser drops it. `bitcoin:tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx?amount=0.1` resolves to the curated mainnet BTC@btc asset and pre-fills the amount. SwapAndPay does no local per-chain address validation (unlike SendForm's `derivationTool.isZcashAddress`), so nothing downstream catches it. Same for `bcrt1…`, `tltc1…`, `rltc1…` and base58 testnet version bytes.

**4. `CrossPayRequest.swift:74` — the amount survives a failed asset resolution and is re-denominated onto a different asset.**
`.display` returns the amount even when `asset` is nil, while `.atomic` correctly guards. Scan `solana:…?amount=0.01&spl-token=<any uncurated mint>`: `selectedAsset` becomes nil but `amountAssetText`/`amountText` become `"0.01"`. The user opens the picker and taps SOL, `.assetTapped` deliberately preserves `amountAssetText` via the `helper` round-trip, `isValidForm` passes — and they pay 0.01 SOL for a request that asked for 0.01 of an unrelated SPL token. There is also a path with no user interaction at all: `applyScannedRequest` sets `selectedContact = nil`, so the next `.swapAssetsLoaded` (fired from `.onAppear`) hits the `selectedAsset == nil && selectedContact == nil` branch at `SwapAndPayStore.swift:813-820` and silently re-defaults to the last-used asset or BTC@btc, leaving the carried number re-denominated. Bitcoin, Litecoin and Solana all map to `.display`, so all three are affected, and nothing covers `resolvedAmount(for: nil)`.

**5. `CrossPayRequest.swift:129` — `powerOfTen` crashes on a negative `decimals`.**
`(0..<exponent)` is built from `SwapAsset.decimals`, read straight off the provider JSON with no range check (`Near1Click.swift:261`). I ran the exact function: `powerOfTen(-1)` traps with *"Range requires lowerBound <= upperBound"*, and `powerOfTen(130)` returns `Decimal.nan`, which the formatter renders as the literal string `"NaN"` straight into the amount field. `getQuote`'s existing `pow(10.0, Double(toAsset.decimals))` tolerates both. Foundation's `pow(Decimal(10), exponent)` fixes the crash and deletes the hand-rolled loop.

**6. `SwapAndPayStore.swift:1573` — the requested amount is silently rounded, and dust becomes zero.**
`conversionCrossPayFormatter` caps at 8 fraction digits with half-even rounding, and the resulting string *is* the payment amount (`getQuote` reads `amountText` back and multiplies by `10^decimals`). Running the exact formatter config: `1.234567890123456789` ETH → `"1.23456789"`, and `0.0000000001` ETH → `"0"`. Because rounding is half-even it can round **up** — `1.999999999999999999` → `2` — so a user can be set up to overpay an invoice. `main` already treats this as a hazard: its Max path uses `.roundedDown(scale: 8)` with the comment *"Floored, never rounded, at the 8 digits the field accepts… the token amount must not creep above the max even by a half-ulp."* Worth reusing that helper here, and considering whether the user should be told the pre-filled amount differs from what the QR asked for.

**7. `SwapAndPayStore.swift:1568` — scanning a URI with no amount wipes what the user typed.**
The three amount fields are cleared unconditionally for anything that parses, then the `guard let amount = …` on 1572 returns before writing anything back. Select USDC@base, type `25`, scan the common static shop QR `bitcoin:1FsSia9rv4NeEwvJ2GvXrX7LyxYspbN2mo` (no `amount=`) and the 25 is gone. This is a new regression — the pre-PR handler only ever assigned `address`. Same for `ethereum:0x…` with no `value=` and `solana:<recipient>` with no `amount=`.

**8. `SwapAndPayStore.swift:1567` — failed resolution nils the user's asset, and every rejection is silent.**
`selectedAsset` is overwritten unconditionally, so a valid-but-unsupported request (native POL on chain 137 — `evmChainsByID` maps it, but the curated list has no native POL asset; or any unlisted ERC-20; or a chain id absent from the table such as zkSync 324) clears the user's selection and drops the amount with no explanation. And when the URI doesn't parse at all, the `guard … else { address = value; return }` dumps the **raw string** into the address field — a Solana interactive transaction link leaves `address == "solana:https://example.com/pay"`, which `getQuote` passes verbatim as `destination`. The feature already owns `@Presents var alert`, an `AlertState` extension, and toast writes; none are used here.

---

## Should fix

**9. `CrossPayRequest.swift:169` — ENS names and over-long hex reach the provider as the destination.**
Checked against the pinned `eip681` crate: `ethereum:vitalik.eth?value=1e18` parses with `recipient_address == "vitalik.eth"`, and `ethereum:0x` + 64 hex chars *also* parses as a native request carrying the full 66-character string — `AddressOrEnsName::parse` uses `HexDigits::parse_min(40)` and `to_erc55_validated_string` only special-cases `Erc55Validation`, so `IncorrectEthAddressLen` falls through and returns `Ok`. The app has no ENS resolver and no validation on this path. The missing validation is pre-existing; what changes here is that a URI is now unwrapped into a plausible-looking but unresolvable destination instead of an obviously-broken raw string. Might be worth reporting the length bug upstream.

**10. `SettingsCoordinator.swift:225` — the other `.scan(.foundString)` handler was not updated.**
The URI-stripping (`CrossPayRequestParser.parse(address)?.address ?? address`) was added to the SwapAndPay coordinator's address-book branch but not here. Settings' scanner uses `[.zcashAddressScanChecker, .swapStringScanChecker]`, so `.foundString` fires exactly when the payload is *not* a Zcash address — the cross-chain-URI case. Scanning `bitcoin:1FsSia…?amount=0.015&label=Shop` from Settings → Address Book → Scan saves a contact whose address is the entire URI including the query string. Contacts match by exact string equality, so it can never match a real address. Same QR, two different outcomes depending on entry point.

**11. `CrossPayRequest.swift:61` — Solana mints compared case-insensitively.**
The `.contract` branch is shared by EVM and Solana, but base58 is case-sensitive while EIP-55 hex is not. A QR whose `spl-token` is a case-mangled variant of the real USDC mint (not itself a valid mint) resolves to the wallet's genuine USDC asset instead of failing closed. Suggest `==` on the Solana path, keeping `caseInsensitiveCompare` for hex.

**12. `Near1Click.swift:276` — the dedupe key excludes the new `contractAddress`.**
`removingDuplicates()` dedupes on `SwapAsset.id` = `provider.chain.token`, so when `/v0/tokens` returns more than one row for a (blockchain, symbol) pair, whichever appears first in the JSON silently decides the canonical contract address — and it runs before `curated(…)`. Since `resolveAsset` now matches on exact `contractAddress`, an ERC-20 request naming the discarded variant resolves to nothing. Harmless before this PR because nothing in `SwapAsset` carried contract identity; now load-bearing on arbitrary API ordering, with no test for duplicate rows.

**13. `CrossPayRequestTests.swift:72` — this test passes without the logic it claims to guard.**
`contractRequestWithoutChainIdResolvesToCurrentAssetChain` asserts `resolveAsset(in: [baseUsdc, arbUsdc], current: arbUsdc) == arbUsdc`, which is satisfied entirely by the `if let current, candidates.contains(current)` shortcut. Delete the whole `resolvedChain` computation and match on `contractAddress` alone — precisely the CREATE2 collision the five-line comment says this test prevents — and it still passes. It needs a `current` that is not itself a candidate (e.g. `current: ethOnArbitrum`, expecting `arbUsdc`).

**14. `CrossPayRequest.swift:78` — EIP-681 native `value` is wei, but is divided by `asset.decimals`.**
The native `value` is protocol-fixed at 18 decimals on every chain in the table; only the ERC-20 `uint256` is in the token's own decimals. The two are conflated in the shared `.atomic` branch. Display and payment are wrong by the same factor so it stays internally consistent, but the pre-filled amount no longer matches the request. All five curated native EVM assets are 18-decimal today so this is probably latent — but two of them (BNB@bsc, AVAX@avax) are `nep245` HOT-omni bridge tokens whose decimals are bridge-defined, and neither `evmNative` test asserts `resolvedAmount` at all, so nothing pins it.

**15. `CrossPayRequest.swift:121` — `nativeTokens(for: "op")` lists OP as a native token.**
OP is an ERC-20 governance token; Optimism's native asset is ETH only. The other aliases all check out (`matic`/`pol` is a genuine rename; `bsc`→`bnb`, `xlayer`→`okb`, `arb`/`base`→`eth`). Latent today since no op-chain asset is curated — but the app already ships `op.imageset` and maps `"op"` in both `chainName` and `tokenName`, so an OP@op listing is clearly anticipated. Requiring `contractAddress == nil` on the native-candidate filters would defuse this whole class of collision, including item 2.

---

## Smaller items

- **`CHANGELOG.md`** — no entry. `CLAUDE.md` requires one under `## [Unreleased]` prefixed `- [MOB-1751] …` for every user-facing change, and this is one.
- **`CrossPayRequest.swift:195`** — `hexDigitValue ?? 0` coerces bad input to a wrong number rather than rejecting: `"0xZZ"` → 0, `"0x1g"` → 16, bare `"0x"` → 0, each of which passes the `guard let amount` and writes an explicit `"0"`. Currently unreachable given SDK validation, but it fails open on an invariant nothing here asserts. (A full uint256 doesn't trap — it silently rounds at the 38th significant digit.)
- **`CrossPayRequest.swift:57`** — `guard chainID == nil || resolvedChain != nil` is fully subsumed by the `guard let resolvedChain` on the next line; both branches behave identically.
- **`SwapAndPayCoordFlowCoordinator.swift:200`** — `CrossPayRequestParser.parse` runs here and again inside `applyScannedRequest`, so the common paths cross the Rust FFI twice and `parsedAddress` is computed even on the branches that discard it. Passing the parsed request in would also collapse the two independent "unparseable → use raw string" fallbacks into one.
- **`CrossPayRequestTests.swift:5`** — missing `@Suite` (165 of the test files declare it; a plain `@Suite` is enough, nothing here touches process-global state).
- **`CrossPayRequest.swift:86`** — `evmChainsByID` / `nativeTokens` are a third and fourth chain vocabulary alongside `SwapAsset.chainName` and `curatedChains()`, and have already drifted (they admit `op`/`xlayer`, which nothing else offers, while `main` has since added `bch`/`dash` this branch predates). Since this PR adds `contractAddress`, deriving native-vs-token from the catalogue would remove the hand-maintained half — worth considering given the file's own note about the Android duplicate.
- **`applyScannedRequest` has no test coverage and is hard to add.** It's a `mutating func` on State called by direct mutation from the coordinator, so it returns no effects and is unreachable from `TestStore` — even though the same coordinator solves the identical root-vs-pushed-form dispatch by *sending actions* 150 lines above. Its last line reads `payUsdLabel`, which derives from `assetAmount`, which short-circuits to `0.0` under `_XCTIsTesting` — so an assertion on `amountUsdText` would pass vacuously. `main`'s Max path documents this trap and works around it by computing values directly.

---

## On the existing localization review

I specifically checked @pacu's concern about `en-US` payment URIs versus other device locales, and the round-trip itself is sound: the parse side uses `Decimal(string:locale: en_US_POSIX)` for the SDK's canonical decimal text, and the write and read formatters both use `Locale.current`, so a `cs_CZ` device round-trips `0.015` → `"0,015"` → `0.015` correctly. The real amount-fidelity problem is not the locale but the 8-digit cap in item 6.

---

Happy to talk through any of these, particularly items 1 and 2 — those are the two I'd want resolved regardless of how the rest lands.

zcash-harry:
Thanks — this was a useful review. I checked every item against the code before I changed anything. All 15 hold up.

Rebased onto `main` (225 commits). Pushed four commits.

I ran the money-critical logic standalone, with stubbed SDK types, because the branch could not link at the time. So each fix is verified by execution, not only by reading.

## Blocking

| # | Item | Status |
|---|------|--------|
| 1 | Scanned QR overwrites the Swap-to-ZEC refund address | Fixed |
| 2 | Chain-less EIP-681 resolves to the wrong asset | Fixed, see note |
| 3 | Testnet and regtest UTXO URIs accepted as mainnet | Fixed |
| 4 | Amount survives a failed asset resolution | Fixed |
| 5 | `powerOfTen` traps on negative `decimals` | Fixed, see note |
| 6 | Requested amount is silently rounded | Fixed |
| 7 | A URI with no amount wipes what the user typed | Fixed |
| 8 | Failed resolution nils the asset, silently | Fixed |

### Items 1, 7 and 8 are one change

These three are the same defect. So I did not patch them separately.

A payment request now applies as a unit, or not at all. Every rejection is reported.

- The swap-mode guard runs first. Swap and Swap-to-ZEC keep the exact raw-string behaviour they had before this PR. I did **not** unwrap the URI into the refund-address field. That is nicer UX, but it is the money path you flagged, so it belongs in a change about swap modes.
- A request naming an asset we cannot pay now changes nothing at all. This is stricter than "do not nil `selectedAsset`". Keeping the user's asset **and** writing the address leaves a recipient on one chain paired with an asset on another. Native POL keeps the address; USDC@base stays selected. Same class of loss.
- The amount fields must always agree with the selected asset. A request with no amount keeps what the user typed **if the asset did not change**. It clears the field if the asset did change.

One note on your USDC@base + `25` + BTC-shop-QR example. It falls on the **clearing** side. After that scan the `25` would mean 25 BTC.

### Item 2: pinned to Ethereum mainnet

I deleted the current-asset fallback. I did not guard it.

You are right that ERC-681's "current network setting" does not exist in this app. I chose `eth` over returning `nil` for three reasons:

1. Resolution becomes a pure function of the URI. The same QR always gives the same result.
2. It matches the default network of every major wallet.
3. It does not degrade the most common EVM URI in the wild.

The user still sees the asset and the amount before they confirm.

Happy to switch to `nil` if you prefer. It is a one-line change and I do not feel strongly.

This also removed `evmChainIfKnown` and the redundant guard from your smaller-items list.

### Item 5: one correction

`pow(Decimal(10), exponent)` alone does not fix this.

It handles the negative trap. It does not handle overflow. I ran both:

```
pow(Decimal(10), -1)    ->  0.1     (was a trap)
pow(Decimal(10), 130)   ->  NaN     (still wrong)
```

So the fix is `pow` **plus** an explicit `(0...127)` guard. 127 is where `Decimal` runs out.

### Item 3

Tests use real bech32 vectors for `tb1`, `bcrt1` and `tltc1`. I generated them and checked my encoder against the BIP-173 mainnet vector first. So they fail for the right reason.

### Item 4

`resolvedAmount(for:)` now requires a non-nil asset in every branch, not only `.atomic`.

The no-interaction path is closed too. Resolution failure is atomic, so `selectedAsset` is never left nil with an address written. The `.swapAssetsLoaded` re-default cannot fire.

### Item 6

Uses `roundedDown(scale: 8)`, the helper you pointed at. I confirmed all three values you cited:

```
1.234567890123456789  ->  1.23456789
0.0000000001          ->  0
1.999999999999999999  ->  2          (half-even, rounds UP)
```

Dust that floors to zero is now refused, not written as `"0"`. A toast tells the user when the prefilled amount differs from the request.

I also stopped reading `payUsdLabel` here. That is the `_XCTIsTesting` trap the Max path documents. The USD value is computed directly.

## Should fix

| # | Item | Status |
|---|------|--------|
| 9 | ENS names and over-long hex reach the provider | Fixed |
| 10 | Settings scan handler not updated | Fixed |
| 11 | Solana mints compared case-insensitively | Fixed |
| 12 | Dedupe key excludes `contractAddress` | Not fixed, see below |
| 13 | Test passes without the logic it guards | Fixed |
| 14 | EIP-681 native `value` is wei | Fixed |
| 15 | `nativeTokens(for: "op")` lists OP | Fixed |

**9.** Rejected for both the recipient and the ERC-20 contract. Reporting the `IncorrectEthAddressLen` fall-through upstream is still on my list. Not done yet.

**13.** You were right that it passed vacuously. Rewritten so `current` is a candidate on its **own** chain. Anything that still consults `current` now returns `arbUsdc` instead of the Ethereum-mainnet asset the request names.

**14.** Split `Amount` into `.wei` (protocol-fixed 18) and `.tokenAtomic` (asset decimals). The test uses BNB@bsc with non-18 decimals, so the bridge-token case is pinned.

**15.** OP dropped from Optimism's native tokens.

I did **not** add the `contractAddress == nil` filter. I cannot confirm whether the NEAR feed reports a contract address for native assets. If it does, that guard silently breaks all native resolution. Item 2's fix already covers the collision class it was meant to defuse. Worth revisiting with real `/v0/tokens` data.

### Item 12: deliberately not fixed

I would like your read on this one.

The hazard is real. Every fix I found trades one problem for another:

- Keying on `contractAddress` puts two "USDC on Base" rows in the picker.
- Any tiebreak between duplicate rows is as arbitrary as JSON order.

What did change is the consequence. Resolution now fails closed. A request naming the discarded variant produces "can't pay this asset" instead of a wrong resolution. Degraded, but safe.

I commented the dedupe site to say contract identity is decided there. Picking a canonical variant looks like a product call about what the picker should show.

## Smaller items

- `CHANGELOG.md` entry added.
- `@Suite` added.
- Hex parsing fails closed instead of `?? 0`. Confirmed by running it: `"0x1g"` gave 16 and bare `"0x"` gave 0.
- The double FFI parse is gone. `applyScannedRequest` takes the parsed request. That also made it reachable from a test.

On testability: there is now a `CrossPayScanApplicationTests` suite. It drives `SwapAndPay.State` directly instead of through `TestStore`. It covers the refund-address guard, atomic rejection, the typed-amount rules, and the floor and dust behaviour. Assertions go through the raw `*Text` fields, for the `_XCTIsTesting` reason you noted.

## On the localization review

@pacu — your concern checks out as safe. The round trip is sound:

- The parse side uses `en_US_POSIX`.
- Both formatters use `Locale.current`.
- A `cs_CZ` device handles `0.015` → `"0,015"` → `0.015` correctly.

The amount-fidelity bug was real, but not locale-related. It was the 8-digit half-even cap. Now floored, with a toast. See item 6.

## Build

You were right that it did not build. It does now. See the next comment.

zcash-harry:
@Chlup — it builds now. The tests you could not run have run.

**Full suite: 1644 passed, 0 failed, 1 skipped, 2 expected failures, of 1647.** That includes all 29 CrossPay tests.

The count matters less than what it proves. The parser tests go through the real Rust FFI. So the crate behaviours I had reasoned about statically are now verified:

- an ENS name parses successfully
- testnet addresses are detected
- `value=1e18` scientific notation is accepted

## What it took

**SDK [#16](https://github.com/zodl-inc/zodl-swift-wallet-sdk/pull/16) was 84 commits behind. It is now rebased onto SDK main.**

It predated the breaking `Synchronizer.evaluateServerSwitch` addition that this app already calls. So it could not have been pinned as it stood.

The rebase was clean. `cargo check` compiles the result as `libzcashlc v4.1.0`. The pinned librustzcash rev is an ancestor of librustzcash main, not a deletable PR sha.

**This branch now uses the local SDK pin** (`22e35afe`), so `build_ffi` compiles the FFI from source.

The built framework exports `T _zcashlc_payment_uri_parse` and declares it in `zcashlc.h`. That is the exact symbol you found missing from both slices.

That commit is marked TEMPORARY and says what to revert.

## Two problems on main, not in this PR

`origin/main` does not compile. I hit both while getting this to build.

### 1. SDK main needs FFI symbols that its own pinned release lacks

`ZcashRustBackend.swift` calls `zcashlc_take_last_error_report` and `zcashlc_free_error_report`. MOB-1201 added them on 24 Aug. `Package.swift` still points at the 4.1.0 release, which predates them.

This is what failed CI on #2049. That PR was merged anyway, with `build` and `unit_tests` both red.

Building the FFI from source produces both symbols. That is why this branch gets past it.

### 2. An app-level break that CI never reached

The same MOB-1201 change turned `ZcashError.rustCreateToAddress`'s payload from a `String` into a `RedactedRustError` struct.

`SendConfirmationStore.swift:389` still called `.localizedCaseInsensitiveContains` on it directly. That is the MOB-385 anchor check from June. The app has no references to `RedactedRustError` anywhere, so it was never adapted.

CI died at the link error first. This only surfaced once the FFI was built locally.

I fixed it in `6c2e3df9` by reading `.message`. That preserves behaviour exactly.

I did **not** switch to the new typed `RustErrorKind.commitmentTreeFailed`, even though that line's own comment invites it. Matching on it would change which failures show the anchor message. That is a MOB-385 call, not a build fix.

**That commit belongs on main, not here.** It is isolated so it cherry-picks cleanly. Main is red for everyone right now.

## Merge order

The SDK owes a libzcashlc release regardless of MOB-1751. If that release is cut from a tree that already contains #16, the payment-URI symbol ships with it at no extra cost.

1. Merge SDK #16.
2. Cut the release. This also fixes main.
3. Revert `22e35afe` here.
4. Merge this.

## One caveat

CI pins Xcode 26.3 for the Swift 6.0 concurrency diagnostics. I ran on 26.5, which is more permissive.

So this is strong evidence, not a guarantee.

secant/Sources/Features/SwapAndPayForm/SwapAndPayStore.swift
zcash-harry: @pacu — good instinct, and I checked it by running the formatters rather than reading them. The path is safe, but not by luck.

## Why this line is safe

The app already has `convertUSToLocale` for exactly the trap you describe. It is used where a **US-formatted string** must reach a locale text field:

- `ScanCoordFlowCoordinator.swift:298`
- `SendFormStore.swift:531`

Cross-pay never holds a US string. It holds a `Decimal`.

| Step | Locale used |
|------|-------------|
| SDK canonical text → `Decimal` | `en_US_POSIX` |
| `Decimal` → `amountText` | `Locale.current` |
| `amountText` → `Decimal` | `Locale.current` |
| `Decimal` → provider API | none |

- Write side is `conversionCrossPayFormatter`, which sets `locale = Locale.current`.
- Read side is `NumberFormatter.zcashNumberFormatter`, which sets **no** locale. That means `Locale.current` too. Both `amount` and `assetAmount` use it.
- The provider never sees a locale-formatted number. `getQuote` sends `BigDecimal.asString(.plain)`, which is locale-independent.

So a `cs_CZ` device is shown `"0,015"` and gets `0.015` back out.

## Verified, across locales

I ran both real formatter configurations over 9 locales and 5 amounts:

```
en_US  cs_CZ  fr_FR  de_DE  pt_BR  ar_SA  fa_IR  hi_IN  ru_RU
0.015   1.23456789   1234.5678   0.00000001   123456789.1234
->  round-trip OK in every case
```

`123456789.1234` is in there on purpose. It has 9 integer digits, and the read formatter sets `maximumIntegerDigits = 8`. Parsing is not affected by that.

`ar_SA` and `fa_IR` cover Arabic-Indic digits, not just a comma decimal separator.

## Added a regression test

An argument is not enough for something you have been burned by before. `prefilledAmountRoundTripsInEveryLocale` drives the **real** formatter objects, not copies of their settings.

It fails if anyone pins either formatter to a fixed locale.

Pushed in `6e6e3c6b`. The CrossPay suite is 30/30 green.

## One related thing I did not change

The other tests in that suite assert literals such as `"0.015"`. That assumes an `en_US` test machine. CI is fixed-locale, so it holds today. It is the same trap one level down, if that ever changes.

## Separately

The real amount-fidelity bug on this line was not the locale. It was the 8-digit half-even cap: `1.999999999999999999` became `2`, which could set a user up to overpay. That is now floored, never rounded up, with a toast when the prefilled amount differs from the request.

secant.xcodeproj/project.pbxproj
```diff
@@ -1359,15 +1359,14 @@
...
-			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "https://github.com/zodl-inc/zodl-swift-wallet-sdk.git";
-			requirement = {
-				kind = revision;
-				revision = 7e44c48aed237202043010775ec7383c45c519fe;
-			};
+/* Begin XCLocalSwiftPackageReference section */
+		3464A8FA303C3A54005A09AD /* XCLocalSwiftPackageReference ".ci/zodl-swift-wallet-sdk" */ = {
+			isa = XCLocalSwiftPackageReference;
+			relativePath = ".ci/zodl-swift-wallet-sdk";
```
Chlup: If using local SDK use `../zodl-swift-wallet-sdk` path.

Chlup (changes requested):
If the URI parsing feature shouldn't be in the Appstore build then it must be enabled only for testnet and internal in code before merging. That is only safe way how to do it. In similar way as send max button for example.

PR #2003, opened 2026-08-21
[MOB-1749] Remaining Orchard funds banner and decision screen

Chlup:
## Summary

Implements [MOB-1749](https://linear.app/zodl/issue/MOB-1749): users who restored a migrated wallet or self-migrated can be left with a small spendable Orchard balance (> 0.0001 and < 0.01 ZEC) that the migration lane never surfaces — below the 0.01 ZEC offer floor it answered "nothing". This adds:

- a new lowest-priority banner, `… ZEC left in Orchard / Tap to decide what happens to it / More`, shown when the selected account has a spendable Orchard balance above 0.0001 ZEC, an Orchard balance below the 0.01 ZEC offer floor on the offer's own basis, a positive Ironwood balance, and the wallet is caught up (the banner and the re-entry route never name a figure the wallet has not verified). It ranks directly below the migration slot itself: only connectivity and sync-error alerts and a real migration outrank it, while the informational reminders (wallet backup, shielding, Tor, currency conversion) rank below it — a knowing trade, since the banner is actionable and retires for good after two taps. A running migration or an unacknowledged Migration Complete always wins. The priority walk also no longer stops at a snoozed shielding reminder, so the rungs below it (Tor, currency conversion) stay reachable.
- a new re-entry route and screen, **Remaining Orchard Funds** ("You've moved to Ironwood"), with an In Ironwood / Locked in Orchard / Pending in Orchard / Left in Orchard summary card (the locked and pending rows appear only when nonzero, so the rows always sum to what the Balances breakdown shows), the lock recommendation, **Lock balance** (→ Locking → Got it) and **Migrate anyway** (immediate review → Sending → Success lane, Keystone included);
- one lock-decision state machine shared by both decision screens: `MigrationLockDecision` (lock/locking/locked, explainer sheet, single-flight Migrate anyway) is embedded by Migration Complete and the residual screen, with the badge / callout / explainer views extracted into `MigrationLockResolutionViews.swift`. The lock and sweep legs are mutually exclusive in both directions.

**Migrate anyway** follows one rule on both screens: it clears output locks exactly when the tapping screen's balance is locked. From an unlocked card it sweeps only the spendable balance the card names and leaves any earlier deliberate lock standing; on a locked card it is the way back — it releases the lock (a blanket release: an earlier visit's lock is released and swept too) and hands over to the review screen, which names the full amount before anything moves. After a release the card re-offers with the released figure folded in.

**Migration Complete** hydrates from the run's own summary first: fresh dust from a completed run gets the decision even when an earlier amount is still locked, and the locked figure takes over (pinning the locked state) only when nothing unlocked remains — so a locked balance can never render as a clean celebratory completion.

Locking is the freshness writer edge: it republishes the snapshot the banner reads before returning (bounded wait), so the banner retires on its own after a lock or sweep on every exit path — including the interactive back-swipe, which closes the flow immediately.

No SDK / librustzcash changes: all figures come from the existing `orchardBalance` buckets (`spendableValue`, `lockedValue`, the pending buckets) already fed by the FFI.

Figma: https://www.figma.com/design/1aeq8gleYh9Yr1l33TwELR/PR-App-Designs-Q3--26?node-id=6855-24737

## Known limitations (deliberate)

- A never-migrated wallet that locks its residual and then leaves the flow has no re-entry — the lock retires the banner by design. The in-session escape ("Migrate anyway" on the locked card) and Migration Complete's permanent one cover the reachable states.
- A banner tap that races a transient sync tick (wallet momentarily not caught up) degrades to the migration Entry fork rather than the residual screen — the safe fallback, never a decision screen over an unverified figure.
- The locked card's "Migrate anyway" button is a deliberate deviation from the Figma locked frame (which draws no button); it currently reuses the warning-styled button on the neutral surface — styling open to a design pass.

## Notes for reviewers

- Spanish copy for the 8 new keys is machine-authored and marked `needs_review` in `Localizable.xcstrings`.
- The summary card reuses `MigrationDetailRow` (radius 12, as on Migration Complete); the Figma frame draws 16.
- Residuals just above 0.0001 ZEC cannot fund a sweep (ZIP 317 fee); "Migrate anyway" then lands on the review screen's existing propose-failure sheet.
- The lock-failure alert reuses the note-split failure copy, as Migration Complete already does.
- Locking waits for one snapshot rebuild before confirming (~5 s typical, capped at 10 s), shown as the in-flight "Locking balance" state.

## Test plan

- [x] Swift Testing suites: banner content, banner + route derivation tables (bounds, Ironwood gate, caught-up gate, precedence), lock-decision reducer (guards, single-flight, release path), `MigrationResidual` store, coordinator re-entry + delegates (unlock-iff-locked on both screens, re-offer fold, immediate back-swipe), Complete hydration (run's dust vs prior lock), snapshot freshness (lock republishes before returning), SmartBanner residual slot + snoozed-reminder walk.
- [x] Full `zodlTests` run (1380 tests / 196 suites) on iPhone 17 Pro (iOS 26.5) simulator.
- [ ] Manual: restored wallet with ~0.005 ZEC spendable in Orchard and funds in Ironwood → banner → More → Lock balance → Got it → banner gone; re-enter → locked row shown → Migrate anyway on the locked card → review names the full amount → send; Migrate anyway from an unlocked card with an earlier lock in place → review names only the spendable amount.


LukasKorba (approved):
utACK

PR #2005, opened 2026-08-24
Switch to new path for local SDK

Chlup:
Just switch path to local SDK build based on new SDK repo (zcash -> zodl).


LukasKorba (approved):
utACK

PR #2008, opened 2026-08-25
Updates for send max feature

Chlup:
## Summary

Follow-up to #1985, which merged with only the two feature commits. This PR carries the review fixes and polish that were meant to ship with it, plus Spanish translations for the new strings. The Max button still ships in **internal and testnet builds only** — the production / App Store build is unchanged (and now provably so, see Verification).

## Behavior fixes

- **The chip enables only when a max can actually be proposed.** `spendability != .nothing` alone was not enough — `.everything` is also the empty-wallet value — so the gates now additionally require a nonzero spendable shielded balance and a selected account. On Swap, USD-input mode also requires a usable ZEC price, matching the Pay gate.
- **Failures are no longer silent.** Every failure path — SDK error, missing transparent receiver, missing/zero USD prices, formatter failure — now shows a toast ("Unable to get the max amount.") instead of just stopping the spinner. The only quiet path left is the deliberate no-op when the swap direction flips mid-request.
- **A stale max is discarded.** On Send, the address stays editable while the request runs; a result that comes back for a different address than the one now in the field is dropped instead of filling the amount (the fee class can differ, e.g. after switching to a TEX address).
- **The max is computed with the memo the proposal will actually carry.** `sendMaxAmount` now takes a `Memo?` and Send derives it with the same shared helper `getProposal` uses (`SendForm.proposalMemo`), so receiver selection and the ZIP-317 fee match what Review will build. Swap passes `nil` (transparent deposit addresses).
- **Leaving the form tears the request down.** Both screens cancel the in-flight request and reset the spinner state on disappear; Send's `.onDisapear` action is now actually dispatched from the view (it previously had no sender, so its cancellation was dead code).

## UI

- New **`ZashiTextFieldTitle`** design-system component: the single source of the field-title style, with a trailing-accessory slot. `ZashiTextField` renders its own title through it, and the Send/Pay amount rows use it to place the chip — removing the three hand-copied title rows.
- The title→field gap is back to the design's **6pt**: the lifted-out label rows were silently adding the VStack's default spacing on top (~14pt total), in production layouts too.
- The chip keeps a **constant width** while its label is swapped for the spinner (the styled title stays in the layout as an invisible sizing ghost).
- The **build gate lives once**, inside `ZashiMaxChip` (`#if ZODL_INTERNAL || SECANT_TESTNET`, else `EmptyView`); the three call-site `#if` copies are gone, so a future call site cannot forget the gate.

## Strings

- `general.max` and `general.maxFailed` now have Spanish translations ("Máx.", "No se pudo obtener el monto máximo."). They were the only display strings in the catalog without `es`; the misleading "needs no translation" comments are gone.

## Tests

- Five new tests: memo pass-through (shielded + transparent), stale-address drop, and teardown cancellation on both screens (hanging-effect pattern — `TestStore.finish()` only passes if the cancellation really happens).
- Existing Max-suite tests updated for the stricter gates and the failure toasts; both suites now pin **and restore** the process-global shared `selectedWalletAccount` they mutate, so they no longer leak state into concurrently running suites.

## Changelog

The two `[Unreleased]` Max entries are `[Internal]`-prefixed and updated to describe the stricter availability, the stale-result handling and the failure toast.

## Verification

- Full `zodlTests` on this branch merged with current `main`: **1333 tests in 192 suites passed** (2 pre-existing known issues, unrelated).
- `zodl-AppStore` (Debug) compile check passes — confirms the production path builds with the chip compiled out and the ungated call sites.
- `grep -rn ZODL_INTERNAL secant/Sources/` → only `ZashiMaxChip.swift`.


LukasKorba (approved):
utACK

PR #2010, opened 2026-08-26
[MOB-1800] Acknowledge the vote Confirm tap instantly and make submission entry idempotent

LukasKorba:
Fixes [MOB-1800](https://linear.app/zodl/issue/MOB-1800/ios-button-transition-from-confirm-needs-to-be-fixed) (Urgent): on iOS, tapping Confirm during voting showed no response until the local-auth round-trip resolved (longer under prover CPU load), and confused re-taps could silently cancel and restart the in-flight batch effect — the likely source of the "already participated" reports.

## What changed

- New `BatchSubmissionStatus.requested`, set synchronously on the Confirm tap: the CTA disables behind a spinner instantly (design-system `ZashiButton` `accessoryView: ProgressView()` pattern), re-taps are no-ops, back navigation stays blocked while auth is pending.
- New `batchAuthenticationDeclined` action: declining Face ID / passcode rolls `.requested` back to `.idle` — a choice, not an error.
- `reduceAuthenticationSucceeded` gains an idempotent entry guard: fresh `.requested` taps and retryable failure states may start the pipeline; an in-flight status is re-enterable only by a resume holding the `pendingBatchSubmission` ticket (which the guard now consumes). Stray duplicates fall through instead of restarting the `cancelInFlight: true` batch effect mid-broadcast.
- The three resume senders (`delegationPrecomputeCompleted`/`Failed`, `delegationProofCompleted`) no longer pre-clear the ticket nor reset the status to `.idle` — the one-action-cycle `.idle` flash (which briefly re-opened back nav) is gone.

No new strings, no timing/sequencing change to when actual submission work starts: the pipeline still begins only after successful auth and finished prep.

## Root cause context

The underlying slowness (sync Rust FFI blocking the Swift cooperative pool + prover CPU saturation) is split out to [MOB-1807](https://linear.app/zodl/issue/MOB-1807/ios-run-voting-rust-ffi-off-the-swift-cooperative-pool) with the full analysis — this PR is the ship-now UX/state-machine layer for the 3.10.x window.

## Testing

- New `zodlTests/VotingTests/ConfirmSubmissionRequestedTests.swift` (7 Swift Testing tests): synchronous `.requested` flip, decline rollback, re-tap no-op (auth-call recorder), duplicate `.authenticationSucceeded` full no-op, and no-`.idle`-flash coverage for all three resume chains.
- Full suite: `Test run with 1452 tests in 206 suites passed` — 0 failures, no existing test modified.


Chlup (approved):
utACK

PR #2011, opened 2026-08-26
[MOB-1749] Residual banner survives its own "Migrate anyway" sweep

Chlup:
## Symptom (field, 2026-08-26)

On a wallet in the Remaining Orchard Funds state (spendable Orchard between 0.0001 and 0.01, Ironwood funds present, synced), using **Migrate anyway** creates and broadcasts the sweep transaction — but back on Home the "… ZEC left in Orchard" banner still shows its pre-sweep state, and tapping **More** now opens the migration entry screen instead of the residual screen. Two surfaces disagreeing about the same wallet.

## Root cause

The banner and the tap answer from different sources. `bannerVariant` serves the **published migration snapshot**; `reentryRoute()` **derives fresh**. After the sweep broadcasts, the fresh derivation immediately answers "no residual" (the recorded immediate run reads as in-progress, which is the designed quiet unmined window → route `.entry`), but nothing on the sweep's success path guaranteed the published snapshot was rebuilt before the user left the flow — so the flow-close re-evaluation re-read the stale `.residual` and re-seated the banner.

Two concrete holes, both pinned by new tests:

1. **The race** — `reconcile()`'s snapshot republish is scheduled on a detached task, never awaited, and a snapshot rebuild is several FFI reads (measured seconds), while Close-on-success → flow-finished re-read is instant. The re-read always beat the rebuild.
2. **The skip** — `reconcile()`'s per-account body guards on the migration-state read *before* scheduling the republish, so one transient advance-step read failure in the post-broadcast window skips the republish entirely; with the wallet idle at the tip, nothing else rebuilds it for the rest of the foreground session.

The lock leg had exactly this bug class and was already fixed by giving `lockMigrationDust` an awaited, bounded republish at the writer edge. The sweep's writer edge never got the same treatment.

## Fix

`MigrationManagerImpl.recordTransferBroadcast(accountUUID:result:)` now ends with the same awaited, bounded snapshot republish the lock uses (`awaitBounded` + `republishSnapshotNow`, 10 s cap). By the time the success chain returns — before the Success screen's Close can fire the flow-finished re-read — the published snapshot tells the post-sweep truth, so the banner retires and the More-tap disagreement disappears with it. The split-phase early return keeps its shape (preparation broadcasts already republish via the broadcast session's own pokes), the function signature is unchanged, and the shared timeout constant's doc now names both writer edges.

Deliberate consequence, same as the lock's precedent: the "Sending…" phase now lasts one snapshot rebuild longer (seconds; 10 s pathological bound) before flipping to Success, in exchange for Home never lying afterwards.

## Tests

- New `MigrationSweepBannerFreshnessTests` (Swift Testing, real `MigrationManagerImpl`, mocked SDK seam), 3 tests:
  - happy-chain convergence (was green before the fix — pins no regression),
  - `theSweepSuccessChainOutrunsItsOwnRepublish` — the race, red before the fix,
  - `aFailedStateReadDuringReconcileStillRetiresTheResidualBanner` — the skip, red before the fix.
- Full `zodlTests`: 1448 tests in 206 suites, TEST SUCCEEDED.


LukasKorba (approved):
utACK

PR #2012, opened 2026-08-26
[MOB-1801] Static voting config v2 + config mirror failover

Chlup:
## Problem

During the live NU7 CHP round, some users could not load Coinholder Polling at all (see MOB-1801 / the internal CHP thread): both legs of voting-service discovery were single-origin, single-attempt. A DNS block, filtered network, or outage on `voting.valargroup.dev` sent the wallet straight to the config-error screen — after hanging up to 120 s — with no fallback. Valar's suggested fix: adopt **static config v2**, whose `dynamic_config_urls` carries an ordered mirror list.

## What changed

- **`StaticVotingConfig` accepts v1 and v2.** v1's singular `dynamic_config_url` and v2's `dynamic_config_urls` array both decode into an ordered `dynamicConfigURLs: [URL]`; unknown versions are rejected with a precise error, and every mirror is enforced HTTPS. v1 stays fully supported, so custom sources saved during the incident keep working.
- **The bundled trust anchor is now the published v2 pin, fetched via a mirror walk.** Two origins for the byte-identical pinned file (the `voting.valargroup.dev` gateway, then `raw.githubusercontent.com`), same embedded SHA-256 for both. The walk falls through on transport failure, non-200, or hash mismatch — availability only; every mirror is independently hash-gated, and a decode/validation failure after a hash match is authoritative and stops the walk.
- **The dynamic config walks its mirror list per the publisher's client semantics**: fall through only on transport failure or 5xx; 4xx and decode/validation failures surface immediately; the first (canonical origin's) error is reported when all fail. GitHub-raw branch mirrors get a per-fetch cache-bust token (their CDN caches ~300 s regardless of request headers), while the clean origin URL is what gets logged.
- **Config requests now time out after 15 s per attempt** instead of inheriting the 120 s session default, so a blackholed route fails over quickly. The Tor routing path (`swapAPIAccess == .protected`) is honored on every mirror attempt.
- **Settings** treats any bundled mirror as the Default source for duplicate-detection and default-matching; user-added custom sources remain single-URL by design.
- Docs (`docs/voting-service-discovery.md`) and `CHANGELOG.md` updated to match.

Trust is unchanged: mirrors widen availability, not trust. The static config is only accepted against the embedded hash pin, and rounds are still authenticated against the pinned `trusted_keys` — a mirror can serve stale rounds but cannot forge one.

## Verification

- New Swift Testing suites `StaticConfigMirrorWalkTests` (9 tests) and `DynamicConfigMirrorWalkTests` (7 tests) cover mirror order, each fall-through class, authoritative stops (incl. hash-matched-but-invalid bytes), first-error retention (distinguishable 503/502 payloads), empty-list guards, timeout/header stamping, host-gated cache-busting with clean-origin reporting, and the bundled-pin invariants (both mirrors pin `28fc9b63…`, checksum never sent as a query param).
- Existing `VotingServiceConfigTests` extended for v1/v2 decode/encode round-trips and validation edges.
- Full `zodlTests` bundle: **1469 tests, 0 failures** (`xcodebuild test`, scheme `zodl-internal`).


LukasKorba (approved):
utACK

PR #2013, opened 2026-08-26
[MOB-1803] Rotate the receive UA ahead of the tap so the screen opens instantly

LukasKorba:
Fixes [MOB-1803](https://linear.app/zodl/issue/MOB-1803/receive-screen-is-slow-during-wallet-sync): opening Receive (and requesting a swap quote) stalled for seconds during catch-up sync, and the tap in turn held up the sync.

## Root cause (full trace in the Linear comment)

`getCustomUnifiedAddress` is a wallet-**DB write** (persists the next diversifier index; `@DBActor`). The slipstream engine writes to the same database near-continuously during catch-up, SQLite allows one writer at a time, and host connections carry a 15s `busy_timeout` — so the awaited-before-navigation write could wait seconds (silently `nil` after 15). Bidirectional: while waiting it holds `DBActor`, and once it gets the SQLite lock the engine waits on it. Matches the Slack thread's conclusion; the structural fix (starvation-free DB delivery, DBActor read/write rework) is deliberately deferred to [MOB-1807](https://linear.app/zodl/issue/MOB-1807/ios-run-voting-rust-ffi-off-the-swift-cooperative-pool).

## The fix: rotate-ahead by one

- `WalletAccount.nextPrivateUA` — a pre-generated, **never-displayed** rotation stash. Startup's `walletAccounts()` enrichment fills the stash instead of the displayed slot.
- Receive tap / Get Quote **promote the stash synchronously and proceed immediately**; a cancellable background effect generates the next one.
- Rotation semantics unchanged: every visit shows a fresh, never-shown-before address — only *when* it was derived moves.
- Empty stash (first run, or re-visit while the refill still fights the engine): Receive opens instantly with a native redacted placeholder + spinner, address controls disabled, and live-fills via `@Shared` when generation lands. The swap flow keeps today's await there (`.getQuote` hard-requires the refund address) and self-heals the stash afterwards.
- Hardening beyond the old code: `updatePrivateUA`/`updateNextPrivateUA` now carry the `AccountUUID` the address was derived for and drop the result if the selection changed mid-generation — closing a pre-existing race where one account's address could be stored under another.

## Review notes

- `RequestZecCoordFlowCoordinator:29` reads `privateUnifiedAddress ?? ""` and would see nil more often — left untouched because the flow is currently unreachable (nothing sets that Root path case); flagged here rather than silently adjusted.
- No new strings, no new components (native `.redacted(reason: .placeholder)`), no pbxproj changes (test dirs are file-system-synchronized).

## Testing

- New `HomeReceiveRotationTests` + `SwapAndPayQuoteRotationTests` (Swift Testing, serialized — they mutate shared state). Key pins: navigation arrives while the generation mock is gated shut (a regression to await-then-navigate times out), the previous visit's address never survives a tap, consecutive visits never expose the same address, swap's both paths keep their contracts.
- Full suite: `Test run with 1450 tests in 207 suites passed` — 0 failures.


Chlup (approved):
utACK

PR #2014, opened 2026-08-26
[MOB-1800] Voting: run proving preparation at interactive priority and stop the 90s cached-delegation stall

Chlup:
## Why

Coinholder-poll submission on iOS is far slower than on Android. The dominant cost is on-device Halo2 proving, and two app-side issues make iOS worse:

1. The proving-key warm-up ran in a `.background`-priority detached task. The Rust keygen threads inherit that QoS (efficiency cores only), and the first real proof of a submission blocks on the same once-initialized key cache — so users sat at "Authorizing…" waiting for background-speed keygen.
2. On a retry after an interrupted attempt, the delegation pipeline probed each bundle's cached tx hash with the full 90-second confirmation budget before falling through to resubmission.

## What

- Warm-up now runs at `.userInitiated` (`VotingCryptoClientLiveKey`).
- The recovery probe in `runDelegationPipeline` is a single fetch (`confirmationTimeout: 0`), matching the flow-entry probe and the Android client; a freshly submitted delegation keeps the full 90-second wait. Covered by two new Swift Testing cases pinning probe-once and fresh-wait-still-retries.

Companion SDK PR (zodl-swift-wallet-sdk → candidate/4.1.0) adds the scoped QoS boost of the shared proving pool; this PR is independent of it and safe to merge alone.


LukasKorba (approved):
utACK

PR #2015, opened 2026-08-26
[MOB-1802] Resume interrupted Keystone voting sessions in place

Chlup:
## The bug (verified end-to-end in code)

A Keystone wallet that lost or interrupted a voting session could end up permanently wedged: restart failed with `Voting backend error: build_and_prove_delegation failed: invalid input: no alpha … (Invalid column type Null at index: 0, name: alpha)`, while re-registering hit "This wallet has already been registered for this voting round". Chain of causes in the round-resume pipeline:

1. The per-bundle registration probe could only recover through a **locally cached delegation tx hash**, made a **single** confirmation attempt (`timeout: 0`), and swallowed its errors with `try?` — so "no hash", "network error", and "not confirmed yet" all collapsed into "not registered".
2. All probes empty → `prepareFreshRound` → **`clearRound`** — destroying alpha/rk/sighash/tx-hash, the only copy of the material a possibly on-chain registration depends on (alpha is random per PCZT build; it cannot be regenerated).
3. Pre-wipe Keystone signatures were then re-stored into the fresh alpha-NULL rows; full coverage fired `.keystoneAllBundlesSigned` straight into `build_and_prove_delegation` → the NULL-alpha crash. Re-delegation attempts were rejected by the chain as duplicates → the "already registered" wall.

## The fix

- **A — typed probe** (`DelegationRegistrationProbe`): `registered` / conclusively `notRegistered` (chain answered: tx failed) / `unknown` (no local hash, network error, unconfirmed, unusable response). Inconclusive is no longer evidence.
- **B — resume in place**: `prepareFreshRound`/`clearRound` is reachable **only** when no bundle holds recoverable material (no signature, no local tx hash) — this invariant is enforced down to the decision function and covered by table tests. Otherwise the round resumes on its intact rows: the SDK's new guarded `resetSessionState` clears only *unsigned* stale setup (signed/registered bundles untouched), and interrupted signing continues from the first unsigned bundle. The unsound clearRound-then-restore-signatures dance is deleted. A failing signature read now aborts the pipeline retryably instead of silently reading as "nothing to preserve".
- **C — trust but verify**: stored signatures are restored only after their sighash is re-checked against the bundle's stored delegation sighash (the SDK's keyless `getStoredPcztSighash` readback). The readback is a plain database read needing no key material, so verification runs **unconditionally** — there is no fail-open path, and a stale or orphaned signature can never be trusted merely because verification could not run. Dropped signatures' bundles simply re-enter the signing queue ("re-sign bundle N" instead of a SQL crash).
- **Defense in depth**: residual incomplete-setup errors (`no alpha for round…` / `Invalid column type Null`) now map to an actionable localized message.

**Scope note:** this makes interrupted sessions resumable and makes the destruction impossible going forward. A wallet the *old* code already wedged (registration on-chain, local record destroyed) cannot be resurrected client-side — that would need a server-side registration lookup by nullifier, tracked separately.

## Depends on

- SDK PR: https://github.com/zodl-inc/zodl-swift-wallet-sdk/pull/21 (`resetSessionState(roundId:)` + the keyless `getStoredPcztSighash(roundId:bundleIndex:)`), head `d544601152a10500d0d66a57d0386511e4a560d8` — rebased onto current `candidate/4.1.0` with every commit building on its own. The app builds against the local SDK path; CI needs that PR merged into `candidate/4.1.0` (or the SDK ref updated) for these symbols.
- Independent of the dark-mode QR fix PR #2009 (same ticket, separate bug).

## History

Rebased from `main` onto `maint/3.10.x` (conflict-resolution notes in the comments below), then reworked for SDK #21's keyless sighash readback with the adaptation folded into the commits that own each piece — the fail-open "verification unavailable ⇒ trust the stored signature" concession was deleted rather than translated (see the handover comment). Follow-ups (in-session retry wedge, `clearKeystoneSignature` adoption on the discard path) are tracked there too.

## Tests

- 16 new Swift Testing tests: probe semantics (5), resume-decision table (4), signature validation (4), error mapping (3).
- Full `zodlTests` suite at the current head, against SDK #21 head `d5446011` (iPhone 17 Pro simulator): **Swift Testing 1559 tests in 213 suites passed** (2 pre-marked known issues) plus **XCTest 48/48** — `** TEST SUCCEEDED **`, and the app target builds with zero errors.


secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -1023,32 +1092,46 @@ extension VotingCoordFlow {
...
+                        // A stored signature is only usable if the bundle row still holds the
+                        // delegation data (alpha/pczt_sighash) it was created against — the
+                        // signature covers that exact sighash. Trusting a stale one routes
+                        // straight into `build_and_prove_delegation`, which dies on the missing
+                        // data. Validate against the stored sighash before the shape check below
+                        // ever sees these signatures, so a mismatch or an incomplete setup drops
+                        // the signature instead of being trusted.
+                        var verifiedSignatures: [KeystoneBundleSignatureInfo] = []
+                        if !storedSignatures.isEmpty {
+                            verifiedSignatures = await Self.validatedStoredSignatures(storedSignatures) { bundleIndex in
```
nuttycom, line 1125: When `getStoredDelegationSighash` throws for an incomplete bundle, this filters the signature only from reducer memory; the persisted signature row remains. That matters because `resetSessionState` ran earlier and deliberately preserves bundles with a stored Keystone signature, so the incomplete alpha/PCZT setup is never cleared and the bundle can hit the same failure when it re-enters signing. Please expose/call the SDK’s `clearKeystoneSignature(roundId:bundleIndex:)` for each rejected signature, then run `resetSessionState` again so the bundle can actually be rebuilt. An integration test for persisted signature + incomplete setup would cover this contract.

nuttycom: Confirmed and fixed in 6c953571, with one deliberate deviation from the literal mechanism. `VotingCryptoClient.clearKeystoneSignature(roundId:bundleIndex:)` is now wired to the SDK's per-bundle delete, and stored signatures are reconciled (`reconcileStoredSignatures`: validate via the stored-sighash check, delete the rows that fail) **right after loading, before the resume decision** — rather than clearing in the late verification block and running `resetSessionState` a second time. Two reasons:

- By the late block, `completeDeterministicRoundSetup` has already stored tree state and rebuilt witnesses; a second reset would drop that work within the same pipeline run. Clearing before the decision lets the resume paths' *existing* reset free the rejected bundles in the same round entry.
- The decision still counts the *loaded* signatures, not the survivors: a stale signature proves an interrupted signing session on this device, and resume-in-place's reset-and-rebuild is the audited recovery for that state. Re-routing an all-stale round to `.freshRound` would send it through `prepareFreshRound`'s `setupBundles` re-run over surviving bundle rows, which is documented as reached only for bundle-less rounds and was never audited for stale delegation fields.

A failing delete throws into `catch:` → `.pipelineFailed` (fail closed, retryable), never a resume on a half-reconciled set. On the test request: no existing test drives the full `.startActiveRoundPipeline` effect (the branch's coverage is all pure-function tables), so the contract is pinned at the same seam the rest of the branch uses — 4 new tests on `reconcileStoredSignatures` covering incomplete setup (throwing readback ⇒ row cleared, signature excluded), mismatch-only clearing, all-valid clearing nothing, and clear-failure propagation. Full suite: 1563 Swift Testing + 48 XCTest, all passing.

LukasKorba (approved):
utACK

PR #2016, opened 2026-08-26
Fix voting submission recovery

czarcas7ic:
Before, a retry after a lost response failed on a spent-nullifier rejection without checking whether that exact submitted transaction had landed. Restart recovery could also clear persisted delegation setup when no transaction hash had been stored or when a database read failure was mistaken for an empty round.

After, spent-nullifier responses query `result.tx_hash` and proceed only when that exact hash returns code 0. Because signing may produce different transaction bytes on retry, a mismatch still errors; critically, the original bundle setup remains intact instead of being wiped. Database read failures also fail closed. This is an app-only patch with no dependency changes.


Chlup (approved):
utACK

PR #217, opened 2022-03-28
Wrappers for the wallet's environment

LukasKorba:
Issue #197 

FYI, the code is slightly bigger that the issue associated, tasks done:
- TCA lightweight wallet wrapper for the keychain storage `WalletStorageInterator`
- SDK integrated, wallet's environment `WalletEnvironment` prepared as well as TCA's `WalletEnvironmentInterator`

Knowing the iInitialised wallet is a combination of keys present in the keychain + database files SDK is handling. We will use WalletEnvironmentInterator as the environment that can handle such functions. To create, validate and check presence of the wallet.

This code review checklist is intended to serve as a starting point for the author and reviewer, although it may not be appropriate for all types of changes (e.g. fixing a spelling typo in documentation).  For more in-depth discussion of how we think about code review, please see [Code Review Guidelines](../blob/main/CODE_REVIEW_GUIDELINES.md).

# Author

- [x] Self-review: Did you review your own code in GitHub's web interface? Code often looks different when reviewing the diff in a browser, making it easier to spot potential bugs.
- [x] Does the code abide by the [Coding Guidelines](../blob/main/docs/CODING_GUIDELINES.md)?
- [ ] Automated tests: Did you add appropriate automated tests for any code changes?
- [x] Code coverage: Did you check the code coverage report for the automated tests?  While we are not looking for perfect coverage, the tool can point out potential cases that have been missed.
- [ ] Documentation: Did you update Docs as appropiate? (E.g [README.md](../blob/main/README.md), etc.)
- [x] Run the app: Did you run the app and try the changes? 
- [ ] Did you provide Screenshots of what the App looks like before and after your changes as part of the description of this PR? (only applicable to UI Changes)
- [x] Rebase and squash: Did you pull in the latest changes from the main branch and squash your commits before assigning a reviewer? Having your code up to date and squashed will make it easier for others to review. Use best judgement when squashing commits, as some changes (such as refactoring) might be easier to review as a separate commit.

# Reviewer

- [ ] Checklist review: Did you go through the code with the [Code Review Guidelines](../blob/main/CODE_REVIEW_GUIDELINES.md) checklist?
- [ ] Ad hoc review: Did you perform an ad hoc review?  _In addition to a first pass using the code review guidelines, do a second pass using your best judgement and experience which may identify additional questions or comments. Research shows that code review is most effective when done in multiple passes, where reviewers look for different things through each pass._
- [ ] Automated tests: Did you review the automated tests?
[...]

PR #216, opened 2022-03-25
CHANGELOG.md for build 0.0.1-17

pacu:
Closes #210 

This is the build corresponding to the sprint ending on March 24th 2022

Issue #47, opened 2021-10-14
Verify Your backup


pacu:
Given an unverified seed phrase, the user must perform an action so that we can attest that the words have been backed up or noted somewhere.

The user is presented a grid with 4 words in a blue button-like element that are missing from the seed phrase below.

On the bottom part of the screen the seed phrase initially show on the previous screen now lacks 4 random chips, one from each visual group

- [x] the user must drag and drop all 4 of them in their corresponding places
- [x] there is no feedback until all words have been placed 
- [x] On success the flow navigates to the success screen 

On failure:
- [ ] Screen shakes and device vibrates half second.
- [ ] Animations revert
- [x] The screen is refreshed. the user must be presented 4 different words from each group to avoid guessing. (@geffen please validate this)

geffenz:
https://xd.adobe.com/view/c058e090-8f05-4777-8370-d5ca819f931a-950b/

![image.png](https://images.zenhubusercontent.com/5ddc49a1bc69ec000196b9a0/be48d22e-d64e-46ed-9d1c-6d7b8392f0ec)

geffenz:
This is PR # 134.

Issue #40, opened 2021-09-28
4/10th Screen Action Sheet


pacu:
Screen name: Walkthrough - done

Create a screen overlay that overlaps the given proportion from the bottom of the screen which can accommodate a view and show the upper part of the screen as a 'background'

out of scope:
- presentation animations

geffenz:
![image.png](https://images.zenhubusercontent.com/5ddc49a1bc69ec000196b9a0/9092ae86-7cc8-4643-b3f1-805a400dc147)

PR #2018, opened 2026-08-26
[MOB-1798] Stop wiping confirmed delegations, and recover ones already lost

dannywillems:
## What happened

`clear_round` ran unconditionally before every `initRound`, to dodge the
`rounds.round_id` UNIQUE constraint on retry. It cascades `bundles` away, and
`bundles` holds `van_comm_rand` — sampled from `OsRng`, never derived from the
seed, and writable through no exported FFI call. Reopening a poll therefore
destroyed the only opening of a VAN the chain already held.

The delivery mechanism was the error message itself. A NULL-column read
surfaced as *"Voting data for this round is incomplete on this device. Please
leave the poll and enter it again to rebuild it"* — and leaving and re-entering
is what called `initRound`. Voters were told to perform the wipe.

`clear_round` is already gone from `VotingCryptoClient`. This PR is about the
devices it already hit.

## What this PR does

**Carves the lost delegation back out of the database.** `van_comm_rand`
survives in freed pages, freeblocks, the unallocated gap, and superseded
write-ahead-log frames, because the bundled SQLite is not built with
`SQLITE_SECURE_DELETE`. `DelegationWalRecovery` parses the SQLite file format
directly to find it.

**Runs at launch, invisibly.** No screen, nothing to tap. Recovery is
read-only and escrows what it finds to `voting-delegation-escrow.json`.

**Recovers the capability tuple, not just the secret.** `van_comm_rand`,
`gov_comm`, `total_note_value` and `delegation_tx_hash` — the per-bundle fields
`import_delegation_capability` builds a bundle from.

**Replaces the error message that caused this** with five diagnosed states,
each carrying a `POLL-NN` code a voter can quote to support.

## Reviewing this

The carver is the part that warrants real scrutiny; it parses a binary format
by hand. `DelegationWalRecovery.swift` opens with diagrams of the four format
levels and cites the sqlite.org spec section for every constant.

**Please confirm for yourself that recovery works**, rather than trusting the
suite: build `zodl-internal`, run
`Scripts/e2e/run-delegation-recovery-e2e.sh`, and read the escrow it prints.

### The two safety properties

Recovery must never run against a healthy database, and must never destroy the
only copy of what it is recovering. Both are asserted, not assumed:

- `theHealthyDatabaseNeedsNoRecovery` and the round-probe suite pin that a
  never-cleared round produces an empty plan.
- `openingTheAppDeletesNothingItRecoveredFrom` compares the preserved
  directory listing before and after, and re-checks every entry still exists
  and is non-empty. The recovery path uses only `FileHandle(forReadingFrom:)`,
  `contentsOfDirectory` and `attributesOfItem` — it opens no SQLite connection,
  precisely so it cannot checkpoint the log away.

### Worth a reviewer's attention

- **Two generations of one row.** `delegation_tx_hash` is written by a *later*
  statement than the secrets, so it lives in a newer generation, while `plan`
  deliberately selects the *oldest* (the broadcast delegation, not the
  rebuild's). The merge is sound only because the field is write-once at the
  source (`AND (delegation_tx_hash IS NULL OR delegation_tx_hash = :tx_hash)`),
  and it is restricted to generations sharing the original's secret.
- **A pre-existing WAL-scan bug this surfaced.** The scan keyed `seen` on
  (round, bundle, `van_comm_rand`) and kept the *first* frame, silently
  discarding the newer of any two generations sharing a secret.
- **Admission is weak on its own.** `isCanonicalPallasElement` only tests
  `value < p`, and p ≈ 2^254, so roughly 1 in 4 random 32-byte blobs passes.
  The escrow stores the full opening (`vanCommRand`, `van`, `totalNoteValue`)
  so the VAN can verify it — **that check is not yet wired**, and is the single
  highest-value follow-up.

## Test coverage

69 tests. 1506 pass on the full target with 0 failures; the device suite is
10 passed / 1 skipped under its flag.

| Suite | Tests | Covers |
|---|---|---|
| `DelegationWalRecoveryTests` | 14 | the carver over both files |
| `VotingRecoveryEndToEndTests` | 10 | build → clear → rebuild → recover |
| `DelegationRoundDataProbeTests` | 9 | never acts on a healthy database |
| `DelegationDiagnosisTests` | 8 | one per error state, plus all 16 signal combinations |
| `DelegationRecoveryDeviceE2ETests` | 6 | recovery driven by opening the app |
| `DelegationWalRecoveryHostileInputTests` | 5 | truncated, corrupt, adversarial bytes |
| `PipelineFailureMessageTests` | 5 | what the voter actually reads |
| `VoteAgain` (nested) | 5 | the incident end to end; 1 disabled |
| `DelegationRecordSizeTests` | 4 | the note-count boundary |
| `RootDelegationRecoveryOnLaunchTests` | 3 | launch triggers it |

The device E2E suites run only under `Scripts/e2e/run-delegation-recovery-e2e.sh`
(`VOTING_DEVICE_E2E=1`) and skip elsewhere. They need exclusive ownership of the
app's Documents directory, which a parallel unit sweep cannot give: Root's
wallet-reset path deletes `voting.sqlite3`, calls `VotingDatabaseSnapshot.reset()`
and removes the escrow, and any suite exercising a reset runs concurrently. The
script does not trust the gate — it requires every test by name and fails on an
unexpected skip.

Fixture filler bytes vary per run from a logged seed
(`[fixture] filler seed N`), replayable with `VOTING_FIXTURE_SEED=N`. Columns
around `van_comm_rand` hold random bytes, so a test asserting on the secret
cannot be satisfied by a neighbour.

## What this PR does NOT do

**It does not let an affected user vote again.** It rescues and escrows the
secrets; nothing consumes the escrow to restore them. That needs an entry point
in `zcash_voting` (written, on a branch, not released) plus an FFI binding.
`restoringTheCarvedDelegationLetsTheRoundBeVotedOn` is written and `.disabled`,
with the blocker spelled out; `theEscrowHoldsEverythingARestoreWillNeed` proves
its inputs are already present.

`import_delegation_capability` is not the tool for it: it is a
capability-*transfer* protocol requiring canonical JSON and a digest
acknowledging delivered bytes to a funds controller, and it rejects non-NULL
heavy columns — which our cleared-*and-rebuilt* rounds have.

**The new diagnoses are wired to one call site**, the round-pipeline failure.
Other `userFriendlyMessage` callers keep the old mapping.

**Note for 3.1.0.** Local voting-hotkey delegation there derives the blinding
from the stored hotkey secret, so it returns without carving. That is a
reproducibility property, not a recovery path — nothing upstream repairs a
database — and reproducing the blinding is not reproducing the VAN, which also
commits to bundle weight and derived addresses. Carving still covers pre-3.1.0
rounds (these devices), the `with_round_bound_voting_target` path whose
blinding is still sampled, and any wallet whose notes moved.

## Security

Recovered secrets are live voting material. They are logged elided (both ends
kept), never in full. Real captures must never be committed; every fixture here
is synthesised.


secant/Sources/Features/Root/RootStore.swift
```diff
@@ -751,6 +751,10 @@ extension Root {
             try? FileManager.default.removeItem(
                 at: documents.appendingPathComponent(DelegationEscrowFile.name)
             )
+            // Preserved copies of the previous wallet's voting database are
+            // the same wallet-scoped material, and must not cross the reset
+            // boundary either.
+            VotingDatabaseSnapshot.reset()
```
dannywillems: Wait 👀

secant/Sources/Dependencies/DelegationEscrow/DelegationEscrowInterface.swift
```diff
@@ -0,0 +1,56 @@
...
+/// One bundle's delegation secrets, captured at the only moment the app ever
+/// holds them.
+///
+/// `vanCommRand` is sampled from `OsRng` inside the Rust core and is never
+/// derived from the wallet seed, so it cannot be recomputed. It reaches Swift
+/// exactly once, as `VotingPcztResult.vanCommRand`, and is otherwise persisted
+/// only in `voting.sqlite3`. Deleting a round takes `bundles` with it via
+/// `ON DELETE CASCADE`, and the FFI exposes no way to write the value back --
+/// so a wipe is terminal for that round's voting rights unless the value was
+/// copied out first. This is that copy.
```
nuttycom: This suggests to me that hard, cascading deletes are the wrong answer - instead, deletions should be "soft" - set a flag indicating that a particular interaction with the voting round has been invalidated.

nuttycom: Following up on the above, with what the code and the upstream queue actually show.

The case for this is stronger than the doc comment here suggests. On `main` the wipe was gated on a **network** query: `existingBundleCount > 0` → try to recover each bundle's VAN position from the chain → `if recoveredBundleCount > 0 { resume } else { prepareFreshRound() }`. Offline, vote server down, or the delegation not yet confirmed, and zero positions come back, so the round gets cleared. An irreversible local delete authorized by transient remote state. The read-failure path did the same thing: `(try? await votingCrypto.getBundleCount(roundId)) ?? 0` turned a database error into "empty round".

So the invariant worth encoding isn't "prefer soft deletes" — it's that `van_comm_rand` opens a commitment already published on an append-only chain. The chain cannot roll back, so the local row that opens it must not be deletable either.

Two caveats before taking this upstream.

**1. A flag alone can't express it.** `bundles` is `PRIMARY KEY (round_id, wallet_id, bundle_index)`. The rebuild re-inserts bundle 0/1/2 at the same key, so an `invalidated` column collides with the row it is meant to protect. It has to be key versioning — surrogate id plus `superseded_at`, with `CREATE UNIQUE INDEX ... ON bundles(round_id, wallet_id, bundle_index) WHERE superseded_at IS NULL`. And `ensure_bundles` needs a matching "open a new generation" path: today it returns `InvalidInput` on a bundle-count or identity mismatch, and `clear_round` is the only exit from that error. Version the rows without fixing that and data loss becomes a hard stop on round entry instead.

Note that `DelegationWalRecovery.Origin` (`databaseFreeSpace < databaseLive < walFrame(n)`) *is* that generation column — reconstructed by parsing SQLite's physical storage, because the logical schema hasn't got one. That is the strongest argument for the soft-delete position.

**2. Two upstream PRs are already moving on this**, so nothing should be filed cold:

- valargroup/zcash_voting#216 gates `clear_round` behind `test-fixtures`, so the destructive path cannot reach a shipped `libzcashlc` at all. That makes `hasDelegationWorthKeeping` / `roundHoldsUnrecoverableDelegation` here a belt on removed trousers.
- valargroup/zcash_voting#217 adds `recover_delegation_from_forensic_evidence` — precisely the write-back path this file's doc comment says does not exist. It takes `van_comm_rand` as an input (it does **not** recompute it, so the premise here still holds), recomputes the VAN from the hotkey target and bundle weight, verifies the leaf position against a freshly root-validated public tree, and restores atomically.

Its `ForensicDelegationBundle` wants `{bundle_index, total_note_value, address_index, van_comm_rand, van_commitment, van_leaf_position, delegation_tx_hash}` — almost exactly what `DelegationWalRecovery.BundleColumn` carves, with the position and tx hash already available from `delegate_vote`'s `leaf_index` in the coordinator. It also NULLs `alpha`, `rseed_signed`, `rseed_output`, `nf_signed`, `cmx_new`, `padded_note_secrets` and `pczt_sighash`: none of those are ZKP #2 inputs, so once a delegation is confirmed the VAN triple really is the entire unrecoverable set. Independent confirmation that the columns carved here are the right ones.

**Proposal for this PR:** keep the pre-open snapshot, the escrow and the carving as they are — they are the evidence gatherer, and nothing upstream produces the evidence. Retarget the restore so the carved bundle is handed to #217's validated API once it lands and is exposed over FFI, rather than this layer writing rows itself; #217's checks (VAN recomputation plus tree-position verification) are strictly stronger than the canonical-element and origin-ordering heuristics here. Drop the clear-round guard once #216 lands.

Next step is to prototype the supersede-instead-of-delete schema change on a fork and rebase it over #216 and #217 before opening anything upstream — filing it now would collide with work already in flight.

---
*Written with Claude Code assistance.*

.gitignore
```diff
@@ -95,4 +95,13 @@ PartnerKeys.plist
 
 # Design docs, implementation plans, and other working notes. These are
 # conversation artifacts rather than repository history.
-.plans/
+.plans
+.notes
```
dannywillems: Reviewer note: .notes is actually ignored globally on my own machine - can be removed.

zodlTests/VotingTests/CorruptedVotingDatabase.swift
```diff
@@ -0,0 +1,329 @@
...
+                database,
+                "UPDATE rounds SET phase = 3 WHERE round_id = '\(Fixture.roundId)';"
+            )
+
+            if rebuildAfterClearing {
+                // Exactly what `clear_round` does — `bundles` cascades away.
+                // Byte for byte what `clear_round` issues, linked on `schema`.
+                // `bundles` goes with it through ON DELETE CASCADE, which is
+                // the whole incident: `van_comm_rand` is destroyed by a
+                // statement that never mentions it.
```
dannywillems: Add a link to the zcash voting sdk to show it reproduces the same line.

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,863 @@
...
+        case totalNoteValue = 15
+        case addressIndex = 16
+    }
+
+    /// Number of columns in `bundles`; a record with fewer is not one of ours.
+    static let bundleColumnCount = 24
+
+    /// Pallas base field modulus. A genuine blinding factor is a canonical
+    /// little-endian element below it.
+    static let pallasModulus = "40000000000000000000000000000000224698fc094cf91b992d30ed00000001"
```
dannywillems: Note: we should have some centralized file with these values regarding cryptography primitives.

Scripts/e2e/run-delegation-recovery-e2e.sh
```diff
@@ -0,0 +1,208 @@
...
+
+# `test-without-building` reinstalls the app, and the data container is
+# migrated to a fresh path. The planted files follow it (the suite's first test
+# asserts exactly that), but the shell's earlier handle is stale, so resolve it
+# again before looking for what the run wrote.
+DOCUMENTS="$(xcrun simctl get_app_container "$UDID" "$BUNDLE_ID" data)/Documents"
+
+say "Escrow written by the recovery"
+if [ -f "$DOCUMENTS/voting-delegation-escrow.json" ]; then
+    cat "$DOCUMENTS/voting-delegation-escrow.json"
```
dannywillems: This is where the recovered value is getting written in. Should we actually test it contains the right value?

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,948 @@
...
+
+    static let walMagic: [UInt32] = [0x377F_0682, 0x377F_0683]
```
dannywillems: Link to the actual spec for these magic values.

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,948 @@
...
+
+    // MARK: - SQLite file format constants
+
+    /// Every constant below is fixed by the SQLite file format specification,
+    /// <https://www.sqlite.org/fileformat.html>. None is a tuning choice: they
+    /// describe a layout that already exists on disk, so changing one does not
+    /// change behaviour, it makes the parser wrong.
+    enum Format {
```
dannywillems: A reviewer should check these values correspond to the spec, link given below.

secant/Sources/Dependencies/VotingCryptoClient/VotingCryptoClientLiveKey.swift
```diff
@@ -244,6 +245,34 @@ extension VotingCryptoClient: DependencyKey {
                 let rseedSigned: Data = Data(result.rseedSigned)
                 let rseedOutput: Data = Data(result.rseedOutput)
                 let actionBytes: Data = Data(result.actionBytes)
+
+                // Escrow the blinding factor before anything else can touch the
```
dannywillems: This is the entry point.

secant/Sources/Features/Root/RootStore.swift
```diff
@@ -405,6 +405,9 @@ struct Root {
     @Dependency(\.continuousClock) var continuousClock
     @Dependency(\.databaseFiles) var databaseFiles
     @Dependency(\.deeplink) var deeplink
+    #if VOTING_ENABLED
```
dannywillems: For my knowledge:

```
DI is dependency injection. And they're two separate packages, both resolved in this project:

swift-composable-architecture  1.25.5   pointfreeco/swift-composable-architecture
swift-dependencies             1.6.4    pointfreeco/swift-dependencies

TCA (The Composable Architecture) is the state-management library — @Reducer, @ObservableState, Store, TestStore. It's what Root, Settings and the coordinators are built on.

swift-dependencies is a separate, standalone library from the same authors (Point-Free). TCA depends on it rather than containing it, and you can use it in a project with no TCA at all. Everything in the previous answer comes from it, not from TCA:
```

secant/Sources/Features/Root/RootStore.swift
```diff
@@ -749,6 +752,12 @@ extension Root {
             // the same wallet-scoped material, and must not cross the reset
             // boundary either.
             VotingDatabaseSnapshot.reset()
+            // The delegation escrow holds VAN blinding factors for this
+            // wallet's rounds. It is as wallet-scoped as the database it
+            // shadows, so it must not survive the reset boundary either.
+            try? FileManager.default.removeItem(
+                at: documents.appendingPathComponent(DelegationEscrowFile.name)
```
dannywillems: What's this?

dannywillems: Ah, okay, it is in the method `clearDeviceScopedWalletState`.

secant/Sources/Features/Root/RootInitialization.swift
```diff
@@ -106,12 +106,36 @@ extension Root {
...
-                return .run { send in
+                let setup = Effect<Root.Action>
+                    .run { send in
                         try await mainQueue.sleep(for: .seconds(0.5))
                         await send(.initialization(.initialSetups))
                     }
                     .cancellable(id: state.DidFinishLaunchingId, cancelInFlight: true)
 
+                #if VOTING_ENABLED
+                // Delegation recovery, on every cold launch, invisible to the
```
dannywillems, line 117: Starting point here, when the user opens the app.

secant/Sources/Features/Root/RootInitialization.swift
```diff
@@ -106,12 +106,36 @@ extension Root {
...
+                // user. An older build could delete a round's blinding factor
+                // and rebuild the round in its place, which makes that poll
+                // permanently unvotable; this copies the original back out of
+                // the preserved files before anything else opens them.
+                //
+                // Deliberately fire-and-forget and merged rather than chained:
+                // it must never delay launch, and it sends no action, so a slow
```
dannywillems, line 124: Are we fine with logging this? And having a file JSON with the recovery value?

zodlTests/VotingTests/CorruptedVotingDatabase.swift
```diff
@@ -0,0 +1,473 @@
...
+                        (round_id, wallet_id, bundle_index,
+                         note_positions_blob, note_identity_hashes_blob,
+                         van_comm_rand, dummy_nullifiers, rho_signed,
+                         padded_note_data, nf_signed, cmx_new, alpha,
+                         rseed_signed, rseed_output, gov_comm,
+                         total_note_value, address_index, rk,
+                         gov_nullifiers_blob, padded_note_secrets, pczt_sighash,
+                         tx1_effects)
+                    VALUES
+                        ('\(roundId)', '\(Fixture.walletId)', \(index),
```
dannywillems, line 419: Using random value in case the random value that gets wiped, `van_comm_rand`, is the same as another value. There is a lower probability that it recovered the wrong value in this case.

zodlTests/VotingTests/CorruptedVotingDatabase.swift
```diff
@@ -0,0 +1,473 @@
...
+        /// column cannot accidentally produce something that looks right.
+        /// `van_comm_rand` is column 5, immediately after two blobs and
+        /// immediately before another, and several of its neighbours are also
+        /// 32 bytes wide; with fixed filler an off-by-one in the column index
+        /// would return a value the assertion might not distinguish.
+        ///
+        /// Deterministic on purpose. A seeded generator means a failure is
+        /// reproducible, while still giving values with no structure for a
+        /// mistaken read to land on. Same seed, same bytes, every run.
+        private static func randomBlob(_ label: String, _ bytes: Int, seed: Int) -> String {
```
dannywillems: why is this method written this way? No generator exists? Like we could use os rng or smth like this, seeded with a random value that gets logged for reproducibility?

zodlTests/VotingTests/CorruptedVotingDatabase.swift
```diff
@@ -0,0 +1,473 @@
...
+                tx1_effects         BLOB,
+                delegation_tx_hash  TEXT,
+                PRIMARY KEY (round_id, wallet_id, bundle_index),
+                FOREIGN KEY (round_id, wallet_id) REFERENCES rounds(round_id, wallet_id) ON DELETE CASCADE
+            );
+            """
+    }
+}
+
+private extension URL {
```
dannywillems: Note: helper. Not important.

zodlTests/VotingTests/DelegationRecoveryDeviceE2ETests.swift
```diff
@@ -0,0 +1,258 @@
...
+            let size = (try? Data(contentsOf: url).count) ?? 0
+            #expect(size > 0, "\(name) is empty")
+        }
+
+        // A checkpointed log carries only its header, and there would be
+        // nothing left to recover from.
+        let wal = try Data(
+            contentsOf: preserved.appendingPathComponent("voting.sqlite3-wal")
+        )
+        #expect(wal.count > 32, "the planted log holds only a header")
```
dannywillems, line 189: Can this test be more precised?
Like, should we verify that nothing is in the table actually?

Packages/VotingRecovery/Sources/VotingRecovery/Recovery/DelegationRecoveryLiveKey.swift
```diff
@@ -0,0 +1,426 @@
...
+            self.name = name
+            self.databaseURL = databaseURL
+            // SQLite names its sidecars by path suffix, not path extension.
+            self.walURL = databaseURL
+                .deletingLastPathComponent()
+                .appendingPathComponent(databaseURL.lastPathComponent + "-wal")
+        }
+    }
+
+    /// Every copy of the voting database on the device, best first.
```
dannywillems, line 26: We MUST ensure that when we have successfully recovered on the affected devices that we remove this code, in case an external app writes over the database, and this code gets executed. However, I think that we protect this path execution by checking if the wallet requires to be recovered or not.

Packages/VotingRecovery/Sources/VotingRecovery/Recovery/DelegationRecoveryLiveKey.swift
```diff
@@ -0,0 +1,426 @@
...
+    /// 3. The live `Documents/voting.sqlite3` LAST. Its log has almost
+    ///    certainly been checkpointed away, though a checkpoint does not zero
+    ///    the cells it releases: this SQLite is not built with
+    ///    `SQLITE_SECURE_DELETE`, so deleted rows survive in freed pages, in
+    ///    freeblocks and in the unallocated gap.
+    ///
+    /// A database restored from an iCloud or device backup needs no case of
+    /// its own: the preserved set and the live database are both deliberately
+    /// backup-eligible, so a migrated device presents them at these paths.
+    static func sources(includingAbsent: Bool) -> [Source] {
```
dannywillems, line 45: This is a bit long, and probably could be shorter.

zodlTests/VotingTests/DelegationRecoveryDeviceE2ETests.swift
dannywillems: Good catch, and it made the test better — though "nothing in the table" turned out to be the wrong assertion, which is itself worth recording.

This fixture is built with `rebuildAfterClearing: true`, so the table is deliberately NOT empty: `clear_round` deleted the broadcast rows and the app then rebuilt the round. SQL sees three rows holding generation 1. What makes the database *the incident* is narrower: generation 0, the secrets the user actually broadcast, is no longer reachable through SQL at all.

So the precondition now pins exactly that:

```swift
let visible = try source.queryVanCommRands()
#expect(visible == Expected.rebuiltRand, "SQL should see the rebuilt round")
for original in Expected.originalRand {
    #expect(visible.contains(original) == false, ...)
}
```

This is the assertion that was missing. Without it a recovery test could pass by reading the secrets straight out of the table and nothing would notice — the file-size checks alone could not tell the difference.

One deliberate subtlety: it queries the fixture's own copy, never the planted one. Opening a database with a live `-wal` makes SQLite recover and checkpoint it, which would destroy the only place the deleted rows survive — the assertion would consume the evidence the rest of the suite depends on. The planted files are byte-identical copies, so the source answers the same question safely. That reasoning is in a comment at the call site.

Also added alongside it, `openingTheAppDeletesNothingItRecoveredFrom`: the existing byte-comparison test would only catch a *deleted* file incidentally, via a thrown error, and said nothing about entries being added or removed. These three files are the only copy of the deleted delegation, so a recovery that consumed them would work once on a device that could never be asked again — and every other test would stay green, since they each re-plant a fresh fixture first.

de5decfb

zodlTests/TestSupport/SeededGenerator.swift
```diff
@@ -0,0 +1,73 @@
+//
```
dannywillems, line 1: Prompted Claude to have that for future usage, @zcash-harry might be interested in.

zodlTests/TestSupport/SeededGenerator.swift
```diff
@@ -0,0 +1,73 @@
...
+//  SeededGenerator.swift
+//  zodlTests
+//
+//  Reproducible randomness for tests.
+//
+//  General-purpose, and deliberately NOT gated on VOTING_ENABLED or tied to
```
dannywillems: Remove this mention of "NOT gated on VOTING_ENABLED". This is meant to be used generically in the future, whatever the voting code exists or not.

dannywillems: Agreed, and applied in 610aa78f. A file meant to be reached for later should not name a build flag that may not exist by then.

Removed the `VOTING_ENABLED` note, and also the "extracted from the delegation-recovery fixture" provenance line for the same reason. `URL+Suffix.swift` had the identical problem, including a `voting.sqlite3-wal` example, so it got the same treatment (now a generic `db.sqlite3-wal`).

Neither file now mentions voting, delegation, or recovery anywhere.

zodlTests/VotingTests/DelegationRecordSizeTests.swift
```diff
@@ -0,0 +1,119 @@
+#if VOTING_ENABLED
```
dannywillems: Should we gate this recovery feature behind a special flag, like `RECOVERY_VOTING_ENABLED`, which depends on `VOTING_ENABLED`?

secant/Sources/Features/Root/RootInitialization.swift
```diff
@@ -106,12 +106,36 @@ extension Root {
...
                 // https://github.com/Electric-Coin-Company/zashi-ios/issues/704
-                return .run { send in
+                let setup = Effect<Root.Action>
+                    .run { send in
                         try await mainQueue.sleep(for: .seconds(0.5))
                         await send(.initialization(.initialSetups))
                     }
                     .cancellable(id: state.DidFinishLaunchingId, cancelInFlight: true)
 
+                #if VOTING_ENABLED
```
dannywillems: This didn't exist before when @zcash-harry tried.

secant/Sources/Features/Voting/DelegationDiagnosis.swift
```diff
@@ -0,0 +1,179 @@
...
+//  DelegationDiagnosis.swift
+//  zodl
+//
+//  Classifies why a poll's local delegation state is unusable, from what can
+//  be OBSERVED about that state rather than from the text of a database error.
+//
+
+import Foundation
+
+/// What is wrong with a round's local delegation state, and therefore what the
```
dannywillems, line 11: This is added to have a better diagnostic.

zodlTests/VotingTests/DelegationRecordSizeTests.swift
dannywillems: This has been agreed with @nuttycom. Implemented in https://github.com/zodl-inc/zodl-ios/pull/2018/commits/17d3d147a5b0aa5eed10aab7bdcba05458267d54

secant/Sources/Features/Voting/DelegationDiagnosis.swift
```diff
@@ -0,0 +1,201 @@
...
+/// voter should be told to do.
+///
+/// This replaces a single message that covered several unrelated causes. That
+/// message read "Please leave the poll and enter it again to rebuild it", and
+/// leaving and re-entering is what called `initRound`, and so `clear_round` --
+/// so the one piece of advice on offer was correct for one cause and
+/// destructive for another. Splitting the causes is what makes the advice
+/// safe.
```
nuttycom: nit: history of this sort does not belong in the source code documentation

secant/Sources/Dependencies/DelegationRecovery/DelegationRecoveryLiveKey.swift
```diff
@@ -0,0 +1,433 @@
...
+    /// Order is TRUST, meaning "how likely is this copy to still hold the
+    /// superseded page images", which is decided by how much SQLite has been
+    /// allowed to do to it:
+    ///
+    /// 1. `voting_recovery/`, copied before anything opened the database, so
+    ///    its write-ahead log is the only place a freshly cleared round
+    ///    survives whole.
+    /// 2. Any other copy found under Documents. Older than the live file by
+    ///    construction, and nothing has opened it since it was made, so its
+    ///    log is likely intact.
```
nuttycom: This seems to be speculating on the existence of files for which we have no plausible mechanism for their creation, but I guess it's not actively harmful.

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
nuttycom: https://github.com/zcash/pasta_curves/blob/18ee865b9dc9a4a3c6f4f84b0d3db3c887b8970c/src/fields/fp.rs#L110

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1047 @@
...
+        // Database header, spec section 1.3.
+        // <https://www.sqlite.org/fileformat.html#the_database_header>
+
+        /// The database file header occupies the first 100 bytes of page 1,
+        /// ahead of that page's b-tree header. No other page carries it.
+        static let fileHeaderLength = 100
+        /// The first 16 bytes of every database file.
+        static let magic = Array("SQLite format 3\u{0}".utf8)
```
nuttycom: From the spec: 
> The header string: "SQLite format 3\000"

> Every valid SQLite database file begins with the following 16 bytes (in hex): 53 51 4c 69 74 65 20 66 6f 72 6d 61 74 20 33 00. This byte sequence corresponds to the UTF-8 string "SQLite format 3" including the nul terminator character at the end.

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1047 @@
...
+    }
+
+
+    /// Page sizes are powers of two between 512 and 65536, per spec 1.3.
+    /// <https://www.sqlite.org/fileformat.html#the_database_header>
+    private static func isPlausiblePageSize(_ pageSize: Int) -> Bool {
+        pageSize >= Format.smallestPageSize
+            && pageSize <= Format.largestPageSize
+            && pageSize % Format.smallestPageSize == 0
+    }
```
nuttycom: This does not find powers of two; it has false positives. Counting the number of `1` bits (must be == 1) would find a power of two.

nuttycom: Also:
> The value 65536 will not fit in a two-byte integer, so to specify a 65536-byte page size, the value at offset 16 is 0x00 0x01. This value can be interpreted as a big-endian 1 and thought of as a magic number to represent the 65536 page size.

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1053 @@
...
+        /// False when the round was never cleared. Callers must treat this as
+        /// "do nothing" rather than restoring anything.
+        var needsRecovery: Bool { !replacements.isEmpty }
+    }
+
+
+    /// Page sizes are powers of two between 512 and 65536, per spec 1.3.
+    /// <https://www.sqlite.org/fileformat.html#the_database_header>
+    private static func isPlausiblePageSize(_ pageSize: Int) -> Bool {
+        pageSize >= Format.smallestPageSize
```
nuttycom: Since 65536 is encoded as 0x00 0x01, doesn't this exclude that option?

nuttycom: Answered: the page size here has already been decoded before this check is made.

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1058 @@
...
+    /// total order is what lets `plan` pick the original without a clock.
+    enum Origin: Comparable, Hashable, Sendable {
+        /// A deleted cell in the database file: freed pages, freeblocks, or
+        /// the unallocated gap. Oldest.
+        case databaseFreeSpace
+        /// A row still reachable through the database file's b-tree.
+        case databaseLive
+        /// A page image in the write-ahead log, by zero-based frame index.
+        /// Newest, since the WAL holds commits made after the last checkpoint.
+        case walFrame(Int)
```
nuttycom: `Origin`'s fixed order `databaseFreeSpace < databaseLive < walFrame(n)` is not an age order once the WAL has been checkpointed without truncation, or restarted. A PASSIVE auto-checkpoint copies the rebuilt secret B into `voting.sqlite3` and leaves the frames in place; the next writer restarts at frame 0 with a new salt, so the pre-wipe secret A survives only in stale-salt frames at *higher* indices, which `recover(walBytes:)` deliberately walks and stamps with the newest rank. A's freed copy in the database file never decodes (`arewrittenRowLeavesNoDecodableCopyInReleasedSpace` pins that), so `versions` is {B at `databaseLive` or `walFrame(0)`, A at `walFrame(i)`}, `plan` elects `original = B`, `DelegationRecoveryLiveKey` escrows it as `.recovered`, and the escrow's recovered-wins rule then refuses any later live capture. The only opening of the on-chain commitment is discarded. `Replacement.current` would carry A, but no production caller reads it.

No fixture combines `rebuildAfterClearing: true` with a subsequent checkpoint, so nothing covers this.

Suggested direction: order WAL frames by salt generation (frame salts against the header salts, or the commit frame's db-size field) and rank already-checkpointed frames at or below `databaseLive`; and escrow `current` as well, so a wrong pick is survivable.

*Written with Claude Code assistance.*

nuttycom (commented):
Full pass over `DelegationWalRecovery.swift`. Ten inline comments follow, ranked by severity.

Two of them undercut claims the PR makes about itself: the election order is not an age order once a checkpoint has run, so the rebuilt secret can be escrowed as the original and locked in; and free-space carving cannot decode a freed cell as documented, because SQLite's freeblock header overwrites the byte the sweep depends on. The next four share one shape: `plan` keeps a single candidate and discards the rest, so a garbage or legitimately deleted copy at the oldest rank shadows the genuine one with no fallback, ties resolve by dictionary order, and an incomplete carve is escrowed with weight zero. Those argue for one design change rather than six patches: elect only among canonical, complete candidates, report ties instead of guessing, and consult the live round state before calling a bundle wiped.

One thing checked and found correct: `van_comm_rand` is sampled as `pallas::Base` (action.rs) and hashed with Poseidon (governance.rs), so the base-field canonicity check and the `fp.rs` citation are right.

*Written with Claude Code assistance.*

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1058 @@
...
+        // record header-length varint before the start of the page.
+        var index = max(1, headerOffset)
+        while index + 1 < page.count {
+            guard page[index] == bundleRecordSignature[0],
+                  page[index + 1] == bundleRecordSignature[1]
+            else {
+                index += 1
+                continue
+            }
+            let start = index - 1
```
nuttycom: The sweep reconstructs a released record's start as `index - 1`, but SQLite's `freeSpace()` writes a 4-byte freeblock header (next-freeblock, size) over the first four bytes of a freed cell. For a `bundles` cell that is: 2-byte payload-length varint, 1-byte rowid (for rowids < 128), 1-byte record-header length, then `0x81 0x0D`. Byte 3, the header length, is replaced by the low byte of the freeblock size, so `decodeRecord` misaligns and `makeBundle` rejects the row. That is the whole `.databaseFreeSpace` path: recovery from in-page freeblocks and the unallocated gap, which the file header and `recover(databaseBytes:)` promise, recovers nothing for the common case. After a clean close (WAL checkpointed and unlinked) the secret is reported unrecoverable while intact on disk.

`arewrittenRowLeavesNoDecodableCopyInReleasedSpace` already observes exactly this, and `carvesTheDatabaseFileWhenTheLogIsGone` asserts only well-formedness of whatever is found, so the suite is green while the path is dead.

Fix options: parse forward from the signature (header length = 1 + the widths of the 24 serial-type varints) so the clobbered byte is never needed; or, if free-space carving is not going to work, correct the doc contract and the PR description.

*Written with Claude Code assistance.*

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1058 @@
...
+            byBundle[
+                BundleKey(roundId: bundle.roundId, bundleIndex: bundle.bundleIndex),
+                default: []
+            ].append(bundle)
+        }
+
+        var replacements: [Replacement] = []
+        for key in byBundle.keys.sorted() {
+            guard let versions = byBundle[key] else { continue }
+            guard let original = versions.min(by: { $0.origin < $1.origin }),
```
nuttycom: Ties in `Origin` for `original` and `newest` are broken by array position, and that position comes from `deduplicated`'s randomly-seeded `Dictionary` iteration through a sort whose key omits `vanCommRand`. Two distinct `van_comm_rand` values for the same bundle at the same rank (a round wiped and rebuilt twice with a checkpoint between, both copies in free space; or a live cell plus a still-decodable freed cell in one WAL frame image, both tagged `walFrame(i)`) can therefore escrow different secrets on different runs over identical bytes, and nothing reports the ambiguity: `Plan` exposes no candidate count and the LiveKey logs only the one chosen.

`plan` should refuse or report a tie rather than guess, and the sort key should include `vanCommRand`.

*Written with Claude Code assistance.*

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1058 @@
...
+        guard candidate.count == fieldElementLength else { return false }
+        // Stored little-endian; compare big-endian against the modulus.
+        let reversed = Data(candidate.reversed()).hexString
+        return reversed < pallasModulus
+    }
+
+    // MARK: - SQLite b-tree pages
+
+    private static func makeBundle(columns: [RecordValue], origin: Origin) -> RecoveredBundle? {
+        guard columns.count >= bundleColumnCount,
```
nuttycom: `makeBundle` accepts any 32-byte blob as `vanCommRand` and `plan` elects the minimum-`Origin` copy with no canonicity filter, so a partially overwritten freed cell that still decodes to the real `(roundId, bundleIndex)` shadows every genuine copy. Concretely: a row with rowid >= 128 is deleted, the 4-byte freeblock header lands exactly on the 4-byte cell header (record framing intact), and a later `allocateSpace` reuses the freeblock's tail (SQLite carves new cells from the end of a freeblock) and overwrites the `van_comm_rand` region. The garbage row gets `.databaseFreeSpace`, the minimum rank, and becomes `original`. `Plan` discards the other `versions`, so when the LiveKey's post-hoc `isCanonicalPallasElement` check rejects it (about 3/4 of the time) it `continue`s with no fallback and the genuine live/WAL copy `plan` had in hand is never escrowed; the other 1/4 of the time the garbage is escrowed as a recovered secret.

The check has to move to where the alternatives still exist: reject non-canonical values in `makeBundle`, or filter `versions` in `plan` before `min`, so election falls through to the next-oldest genuine copy.

*Written with Claude Code assistance.*

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1058 @@
...
+            // the rebuild `prepareFreshRound` performs. And every surviving
+            // copy sitting in released space with nothing live above it: the
+            // row was deleted and never rebuilt, so it is gone from the
+            // database entirely.
+            //
+            // Neither fires for an untouched round. Its bundles are live, so
+            // they always have a live or WAL origin, and a page rewritten any
+            // number of times still yields exactly one `van_comm_rand`.
+            let distinct = Set(versions.map(\.vanCommRand))
+            let allReleased = versions.allSatisfy { $0.origin == .databaseFreeSpace }
```
nuttycom: `allReleased` treats a bundle whose only surviving copies sit in free space as a wiped-and-never-rebuilt round, and the comment above says the rebuild is the only thing that deletes `bundles` rows. It is not: `deleteSkippedBundles` in `VotingCoordFlowCoordinator` runs `DELETE FROM bundles WHERE bundle_index >= :from_index` on rows that already hold a `van_comm_rand` (stored when the PCZT was built) and never rebuilds them. A Keystone user who signs bundles 0-1 and skips the rest leaves bundle 2 as a freed cell; once that cell survives carving (rowid >= 128, or a multi-page table) the next launch emits `Replacement(original: bundle2, current: nil)`, the LiveKey escrows it as `.recovered`, the escrow makes that sticky, and the coordinator reports POLL-03 (`secretsRecovered`) for a round from which nothing was lost. No test exercises `allReleased == true`.

The verdict needs the live `rounds`/`bundles` state (is this bundle index still expected?) rather than carving alone.

*Written with Claude Code assistance.*

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1058 @@
...
+        if case let .blob(stored) = columns[BundleColumn.govComm.rawValue],
+           stored.count == fieldElementLength {
+            van = stored
+        } else {
+            van = Data()
+            isComplete = false
+        }
+
+        var weight: UInt64 = 0
+        if case let .integer(stored) = columns[BundleColumn.totalNoteValue.rawValue], stored >= 0 {
```
nuttycom: A truncated or non-integer `total_note_value` degrades to `0`, and the only record of that is `RecoveredBundle.isComplete`, which nothing outside this file reads and which `DelegationEscrowEntry` does not carry. A payload cut by overflow just past `gov_comm`, or a sweep hit with `.null` at column 15, is therefore escrowed as `totalNoteValue: 0`, indistinguishable from a genuine zero weight; the LiveKey's admission checks are only `isCanonicalPallasElement` and `van.isEmpty`, the escrow then blocks any later live capture from replacing it, and the VAN self-check `DelegationEscrowInterface` promises is not implemented anywhere. A later capability import would rebuild a VAN over weight 0 that cannot open the on-chain commitment.

Either reject incomplete bundles at admission or carry completeness into the escrow entry.

*Written with Claude Code assistance.*

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1058 @@
...
+        headerOffset: Int = 0
+    ) -> [(columns: [RecordValue], isLive: Bool)] {
+        var records: [(columns: [RecordValue], isLive: Bool)] = []
+        var liveRanges: [Range<Int>] = []
+
+        if page.count > headerOffset, page[headerOffset] == Format.tableLeafPageType {
+            for payload in tableLeafPayloads(
+                page: page, usable: usable, headerOffset: headerOffset
+            ) {
+                liveRanges.append(payload.range)
```
nuttycom: `liveRanges` gets the cell's range before `decodeRecord` has succeeded, so a cell the pointer walk located but could not decode is also suppressed from the signature sweep, whose decode window is strictly larger. And `tableLeafPayloads` validates cell offsets only as `0 < offset < page.count`, so any page image whose first byte happens to be `0x0D` (a freed page, an overflow page, a stale frame) yields bogus live ranges that blanket real records. A bogus cell with a small `payloadLength` starting on a real `bundles` record makes `decodeRecord` return nil (declared header exceeds the window), and the sweep then skips the same bytes. If that was the only live generation and a released copy exists elsewhere, `allReleased` fires and an intact round is reported wiped.

Add a range to `liveRanges` only once it decoded, and bound cell offsets to the cell-content area.

*Written with Claude Code assistance.*

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1058 @@
...
+
+        let database = try Data(contentsOf: databaseURL, options: .mappedIfSafe)
+        rows += recover(databaseBytes: [UInt8](database), roundId: roundId)
+
+        if let walURL, FileManager.default.fileExists(atPath: walURL.path) {
+            let wal = try Data(contentsOf: walURL, options: .mappedIfSafe)
+            // The reserved-region size lives in the database header, so the
+            // log can only be read correctly alongside its own database.
+            let reserved = database.count > Format.reservedSizeOffset
+                ? Int(database[Format.reservedSizeOffset])
```
nuttycom: The reserved-bytes-per-page byte is read from the database header with no magic or geometry validation and handed to the WAL scan as `reservedPerPage`. Nothing upstream guarantees a valid header for the preserved or live sources: `sources()` screens those only through `VotingDatabaseSnapshot.holdsData`, a size > 0 check, and `looksLikeSQLite` applies only to discovered extras. A corrupt `voting_recovery/voting.sqlite3` beside an intact `-wal` is rejected by `geometry`, but its byte 20 still shrinks `usable`, and `localPayloadSize` then truncates the WAL's local payload reads so `van_comm_rand` from an otherwise recoverable log is mis-decoded or dropped.

Gate the reserved byte on `geometry(databaseBytes:)` succeeding, and cross-check the WAL header's page size against the database's.

*Written with Claude Code assistance.*

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1058 @@
...
+
+        let pageSize = Int(readUInt32(walBytes, Format.walPageSizeOffset))
+        guard isPlausiblePageSize(pageSize) else { return false }
+
+        var offset = Format.walHeaderLength
+        let frameLength = Format.walFrameHeaderLength + pageSize
+        while offset + frameLength <= walBytes.count {
+            let imageStart = offset + Format.walFrameHeaderLength
+            let page = Array(walBytes[imageStart..<(imageStart + pageSize)])
+            for (columns, _) in bundleRecords(inPage: page, usable: pageSize)
```
nuttycom: `holdsRoundData(walBytes:)` hand-copies the frame loop of `recover(walBytes:)` but omits the `reservedPerPage` subtraction and the page-1 `headerOffset`, so the probe and the carver disagree about the geometry of the same page, despite `geometry` existing precisely so they cannot. `plan(walURL:)` likewise defaults `reservedPerPage` to 0 and has no callers. No wrong output is reachable today, since both deviations only widen the probe's window and it stays a superset of the carver, but the probe is the gate that short-circuits the LiveKey to `.nothingToRecover`, and a future edit to either loop desynchronizes them silently.

Share one frame walk (a `walGeometry` counterpart to `geometry`) that threads `reservedPerPage` and the frame's page number through both, and delete the uncalled `plan(walURL:)`.

*Written with Claude Code assistance.*

secant/Sources/Dependencies/DelegationEscrow/DelegationWalRecovery.swift
```diff
@@ -0,0 +1,1058 @@
...
+                // with the same `van_comm_rand`: once as
+                // `store_delegation_data` wrote it, and again once
+                // `store_delegation_tx_hash` filled in the hash afterwards.
+                // Keying on the secret alone keeps whichever frame came first
+                // -- always the one WITHOUT the hash -- and the later
+                // generation is dropped here, before `deduplicated` or `plan`
+                // can merge it. Both generations must survive the scan for
+                // that merge to have anything to work with.
+                let key = "\(bundle.roundId)/\(bundle.bundleIndex)/"
+                    + "\(bundle.vanCommRand.hexString)/\(bundle.delegationTxHash ?? "")"
```
nuttycom: This concatenates two interpolated strings with `+`, which `.swiftlint.yml`'s `string_concatenation` rule reports at error severity (and CLAUDE.md forbids). Running `swiftlint lint --config .swiftlint.yml` on this file also reports `type_body_length` as an error: the enum body is 511 lines against the 350 limit, plus warnings for a large tuple in `geometry`, `String(decoding:as:)` on `Data`, and three double blank lines. Only the commented-out build-phase invocation keeps this from failing today.

One interpolation fixes the key; the length wants `Format`, the record decoder, and the primitives split into extensions or separate files.

*Written with Claude Code assistance.*

nuttycom (commented):
Nine of the inline comments above were re-anchored onto the pushed revision; the first copies sat five lines off after the push landed mid-post.

*Written with Claude Code assistance.*

secant/Sources/Features/Voting/DelegationRestore.swift
```diff
@@ -0,0 +1,138 @@
...
+            return nil
+        }
+
+        var bundles: [RecoveredDelegationBundleInput] = []
+        var hashes: Set<String> = []
+        for entry in recovered {
+            guard let hash = entry.delegationTxHash,
+                  isLowercaseHexHash(hash),
+                  hashes.insert(hash).inserted,
+                  entry.vanCommRand.count == 32,
```
nuttycom: Before any recovered bundle can authorize clearing live state, authenticate it independently: reconstruct `VAN(hotkey, round, exact weight, vanCommRand)` and require equality with the recovered `van`. This lets the feature safely recover the maximal usable contiguous subset under the short deadline while rejecting random carve hits, wrong WAL generations, partial overwrites, wrong weights, and wrong-wallet hotkeys. Keep the transaction hash mandatory as evidence that the bundle was submitted. The current tests use unrelated constant VANs; add a real equality test plus an end-to-end partial-recovery vote.

secant/Sources/Features/Voting/DelegationRestore.swift
```diff
@@ -0,0 +1,138 @@
...
+        expectedBundleCount: UInt32?
+    ) -> [RecoveredDelegationBundleInput]? {
+        let recovered = entries
+            .filter { $0.source == .recovered }
+            .sorted { $0.bundleIndex < $1.bundleIndex }
+        guard !recovered.isEmpty else { return nil }
+        guard recovered.enumerated().allSatisfy({ $0.element.bundleIndex == UInt32($0.offset) }) else {
+            return nil
+        }
+        if let expectedBundleCount, recovered.count != Int(expectedBundleCount) {
```
nuttycom: Given that the round ends in days, this need not block recovery if bundles are independently usable: accepting a recovered contiguous prefix can salvage some voting power when a trailing bundle is gone. Please make that an explicit partial-recovery contract rather than claiming the delegation is complete, and add an end-to-end case where an original 3-bundle delegation restores `[0, 1]` and successfully votes. Preserve the original indices; do not renumber across gaps. If downstream voting requires the original total count, then an unknown count remains unsafe.

secant/Sources/Features/Root/RootInitialization.swift
```diff
@@ -106,12 +106,36 @@ extension Root {
...
+                // the preserved files before anything else opens them.
+                //
+                // Deliberately fire-and-forget and merged rather than chained:
+                // it must never delay launch, and it sends no action, so a slow
+                // or failing carve cannot affect what the user sees. What it
+                // did is in the log under "[poll-recovery]".
+                //
+                // It costs almost nothing for the overwhelming majority of
+                // wallets, which have no voting round at all and are turned
+                // away by `holdsRoundData` after one file read.
```
nuttycom, line 135: This uncancelled fire-and-forget task can cross the wallet-reset boundary. Recovery first reads the old database into memory and later calls `delegationEscrow.record`; meanwhile reset directly removes the database, snapshot, and escrow without cancelling or awaiting this effect. If reset lands between those operations, recovery recreates the escrow afterward with the previous wallet’s voting secrets. The escrow actor cannot serialize this because reset bypasses it and the old records are already in the task. Please give recovery a cancellation/ownership boundary that reset waits for (and route reset through the serialized store), so old wallet material cannot be resurrected for the replacement wallet.

secant/Sources/Features/Voting/DelegationRestore.swift
nuttycom: Done, on both sides of the FFI.

**SDK (#40).** Each bundle in the restore request now carries `van`, the commitment its row held. Rust recomputes `construct_van(hotkey address, weight, round, blinding)` per bundle and refuses with `bundle N does not open its commitment; nothing may be cleared` before it reads or writes the round. `zcashlc_voting_van_commitment` exposes the same computation so a caller, or a fixture, can produce a commitment the restore will accept. The transaction hash stays mandatory.

**App.** The restore recomputes every candidate's commitment through the crypto client before offering it, and passes over one that does not open. That matters in practice: against the incident fixture the decoder sometimes yields a second image of an original whose blinding survived but whose commitment and weight the rebuild overwrote. The escrow keeps such images apart by commitment, and the intact one is what restores.

**Tests.** The fixture now stores real commitments for the hotkey the test controls, so the vote-again suite exercises the check end to end; `aBundleThatDoesNotOpenItsCommitmentIsRefused` covers the SDK, `aCandidateWhoseBlindingDoesNotOpenItsCommitmentIsSkipped` the app.

One correction to the framing: a rebuilt row is self-consistent, so commitment equality rejects garbage, partial overwrites, wrong weights and wrong-wallet hotkeys, but does not tell generations apart. The hash and, in the end, the chain's tree do that; a refused leaf marks the candidate so the next restore offers the next one.

_Posted with Claude Code assistance._

secant/Sources/Features/Voting/DelegationRestore.swift
nuttycom: Done, as a prefix contract.

Verified in `zcash_voting` 3.0.0: the capability codec requires bundle indices contiguous from zero, and votes are keyed per `(bundle_index, proposal_id)`, so a prefix is independently usable and a gap is not. The restore therefore offers the longest run from bundle zero the escrow can supply, under the original indices, and the SDK guard was relaxed to match: it refuses a round holding votes, shares, Keystone signatures, or a bundle with an accepted hash the package does not restore, and otherwise accepts a package that stops short of what a failed rebuild left behind or extends a prefix restored earlier. "Already restored" is reported only on exact equality.

`restoringAPrefixLetsThoseBundlesVote` plants a three-bundle delegation, loses bundle 2, restores `[0, 1]` through the real backend, and submits two votes carrying those bundles' commitments.

_Posted with Claude Code assistance._

secant/Sources/Features/Root/RootInitialization.swift
nuttycom: Fixed, with the two halves you describe.

**Cancellation.** The launch-time recovery effect now carries a cancel id, and `.resetZashi` cancels it before the wipe starts, whether or not the SDK wipe can even begin. `resettingTheWalletCancelsARecoveryStillRunning` drives that through the reducer.

**Ownership.** Reset no longer deletes the escrow file behind the store. It calls `DelegationEscrowFile.invalidate`, which bumps a generation counter and then removes the file. A recovery run takes a lease on the generation when it starts and writes through `recordRecovered`, which refuses with `staleLease` once the generation has moved on. Because the reset bumps before it deletes, a write that lands between the two re-checks the generation after persisting and removes what it wrote, so no interleaving lets the old wallet's secrets reappear. `aRecoveryLeaseGoesStaleWhenTheWalletResets` covers the boundary; the run reports `.cancelled` and stops.

_Posted with Claude Code assistance._

secant/Sources/Dependencies/DelegationEscrow/DelegationEscrowInterface.swift
nuttycom: Update, now that the work is in. Most of the plan above was overtaken.

**The two upstream PRs did not land.** valargroup/zcash_voting#216 (gate `clear_round` behind test fixtures) and #217 (`recover_delegation_from_forensic_evidence`) are both closed, unmerged. Nothing here waits on them any more.

**What replaced #217.** The write-back path is `import_delegation_capability`, which already exists in the pinned 3.0.0, wrapped by a guarded FFI in zodl-swift-wallet-sdk#40. The checks #217 promised are there, split across two points: the FFI recomputes each bundle's commitment from the hotkey, weight and blinding and refuses before it reads or writes the round; tree sync then verifies the leaf against the root-validated public tree, and a refused leaf marks that candidate in the escrow so the next restore offers the next one. The app also recomputes the commitment per candidate before offering it, because the decoder can yield a partly overwritten image of a real row.

**What replaced #216.** The app never clears. The SDK clears only inside that FFI, and only when every row the package does not restore is provably useless to the wallet: no votes, shares, or Keystone signatures, and no bundle with an accepted hash the package fails to restore. A package may stop short of what a failed rebuild left behind, or extend a prefix restored earlier.

**The generation column.** The soft-delete argument stands, but the schema was not the place to put it for this release. The escrow now plays that role: every schema-consistent row the decoder finds is kept, keyed by bundle, blinding and commitment, with where it was found and how much that place is trusted. The `Origin` ordering this comment cited is gone with the old carver; #2026's decoder replaced it, and it elects nothing. Supersede-instead-of-delete in `bundles` remains the durable upstream fix, and it is unblocked now that neither #216 nor #217 is in flight.

The doc comment this thread hangs on is corrected on the branch, in the commit that exposed the restore, and goes out with its next push: the FFI does expose a way back, and this escrow is what it reads.

_Posted with Claude Code assistance._

secant/Sources/Dependencies/VotingDatabaseSnapshot/VotingDatabaseRecovery+Records.swift
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    static func blob(_ value: RecordValue) -> Data? {
+        guard case let .blob(data) = value else { return nil }
+        return data
+    }
+
+    static func isCanonicalPallasElement(_ candidate: Data) -> Bool {
+        guard candidate.count == 32 else { return false }
+        return Data(candidate.reversed()).hexString < pallasModulus
```
nuttycom: Is this really using hex string sorting for this? That's disgusting.

Packages/VotingRecovery/Sources/VotingRecovery/Restore/DelegationRestore.swift
```diff
@@ -0,0 +1,183 @@
...
+        opens: (DelegationEscrowEntry) -> Bool = { _ in true }
+    ) -> [RecoveredDelegationBundleInput]? {
+        let candidates = Dictionary(
+            grouping: entries.filter { $0.source == .recovered && $0.rejectedAt == nil },
+            by: \.bundleIndex
+        )
+        var bundles: [RecoveredDelegationBundleInput] = []
+        var hashes: Set<String> = []
+        var index: UInt32 = 0
+        while let bundle = candidates[index]?
```
nuttycom, line 171: **P1:** Restoring the longest available prefix can still erase a usable trailing bundle. The SDK guard currently treats an omitted live bundle with delegation_tx_hash == NULL as safe to clear, but that does not prove the transaction was never broadcast: the app can die after broadcast succeeds and before storeDelegationTxHash persists the hash. In that state the existing bundle may still be recoverable through the normal chain lookup, while this shorter-prefix import clears it permanently. Since recovery is best-effort but must not regress existing wallet behavior, please either strengthen the SDK guard to prove every omitted bundle is unaccepted, or avoid partial-prefix restore until that proof is available.

secant/Sources/Features/Voting/DelegationRestore.swift
```diff
@@ -0,0 +1,183 @@
...
+    /// was a leaf mismatch at all.
+    static func rejectRestoredCandidates(
+        roundId: String,
+        error: Error,
+        escrow: DelegationEscrowClient
+    ) async -> Bool {
+        guard error.localizedDescription.contains(leafMismatchMarker) else { return false }
+        let entries = (try? await escrow.entries(roundId)) ?? []
+        guard let offered = package(from: entries) else { return true }
+        for bundle in offered {
```
nuttycom: **P1:** A leaf mismatch names the exact failing bundle (for example, confirmed delegation bundle 0), but this loop permanently rejects every candidate in the offered prefix. If bundle 0 is wrong and bundle 1 is correct, the next attempt loses the correct bundle-1 candidate too and may restore less voting power—or none—before the poll closes. Please reject only the bundle index identified by the failure, preferably using structured SDK error information rather than matching and parsing a general error string.

Packages/VotingRecovery/Sources/VotingRecovery/Decoder/VotingDatabaseRecovery.swift
```diff
@@ -0,0 +1,477 @@
...
+        )
+    }
+
+    static func recoverAll(
+        databaseURL: URL,
+        walURL: URL? = nil,
+        roundId: String?,
+        walletId: String? = nil
+    ) throws -> Report {
+        let database = try Data(contentsOf: databaseURL, options: .mappedIfSafe)
```
nuttycom, line 232: **P2:** Cold-launch recovery still has unbounded memory cost. Each database and WAL is mapped here and then fully copied into an Array of UInt8 below, and the caller repeats that for every matching source found under Documents. Running this in a fire-and-forget effect avoids blocking launch sequencing, but it does not prevent memory pressure or process termination, so an unusually large database/WAL can regress ordinary wallet startup. Please impose a conservative file-size/source-count budget or process the input incrementally by page.

secant/Sources/Dependencies/DelegationEscrow/DelegationEscrowInterface.swift
```diff
@@ -0,0 +1,180 @@
+#if RECOVERY_VOTING_ENABLED
```
pacu: I don't think using macros for this is the best approach 
It's too invasive. This should be more self contained, be in its own module and then be removed when not needed any more

secant/Sources/Dependencies/VotingCryptoClient/VotingCryptoClientInterface.swift
```diff
@@ -16,6 +16,37 @@ enum VotingTxHashLookup: Equatable, Sendable {
     case present(String)
 }
 
+/// One bundle to restore, in the escrow's own terms.
+struct RecoveredDelegationBundleInput: Equatable, Sendable {
```
pacu: Shouldn't this be behind that flag?

pacu (dismissed):
I want to push back on this compiler flag. It's the wrong tool for the job and reveals poor design. 

Also there are parts of the vote recovery that are not behind it.

secant/Sources/Dependencies/DelegationEscrow/DelegationEscrowInterface.swift
nuttycom: None of this code should be in the app at all. It should all have been in the SDK in the first place. The problem here is with the recovery code, but it's more general. We should have ripped it all out and put it into the SDK back when it was originally contributed, but we didn't and now we're paying that price.

nuttycom: I'll have an agent look at moving stuff into a place where it's better contained within the app, but the code quality of all the voting code is just appalling - I think ultimately we need to get the specifications in order, and then rewrite from good specs with a coherent plan.

secant/Sources/Dependencies/VotingDatabaseSnapshot/VotingDatabaseRecovery+Records.swift
nuttycom: Replaced. The modulus is now a 32-byte big-endian constant, and the check compares the candidate's reversed bytes with `lexicographicallyPrecedes`. The text form was right only by the accident of fixed-width lowercase hex.

_Posted with Claude Code assistance._

secant/Sources/Features/Voting/DelegationRestore.swift
nuttycom: Done. The rejection now marks only the bundle the message names, and it finds that candidate with the same commitment predicate the restore used, so the entry marked is the one that was offered rather than the best-ranked image regardless of whether it opens. Only when the message names no bundle does it fall back to the whole offered set, since nothing then tells them apart.

The message is the pinned crate's, so parsing it is the interim. A classified error report carrying the bundle index is the clean form and belongs on the SDK side; I'll take it as a follow-up alongside the leaf lookup.

_Posted with Claude Code assistance._

Packages/VotingRecovery/Sources/VotingRecovery/Decoder/VotingDatabaseRecovery.swift
nuttycom: Bounded. A run reads at most eight copies, best first, and skips any copy whose database and log together exceed 64 MiB, logging the skip and counting it in the report. The decoder copies the mapped file once, so peak memory stays near twice the budget; a polling database is a few hundred kilobytes, so a copy past the budget is not one recovery was written for. Page-wise incremental decoding would mean rewriting the decoder's WAL replay, which the budget makes unnecessary for the risk named here.

_Posted with Claude Code assistance._

secant/Sources/Dependencies/DelegationEscrow/DelegationEscrowInterface.swift
nuttycom: Agreed, and done at the review level: the flag is gone and the escrow, along with the rest of recovery, now lives in `Packages/VotingRecovery`. This file is the module's, not the app's. See the review reply for the shape and the removal procedure.

_Posted with Claude Code assistance._

secant/Sources/Dependencies/VotingCryptoClient/VotingCryptoClientInterface.swift
nuttycom: Neither behind it nor here any more: the crypto client no longer carries any recovery type or closure. The restore's two SDK calls live in `Packages/VotingRecovery`, built over the backend the app registers once through `VotingRecovery.configure`, which is the only recovery line left in this LiveKey besides the snapshot capture and the live escrow write, each marked for removal.

_Posted with Claude Code assistance._

Packages/VotingRecovery/Sources/VotingRecovery/Restore/DelegationRestore.swift
nuttycom: Taken the second option. The SDK guard (zodl-inc/zodl-swift-wallet-sdk#47, since #40 merged before this landed on its branch; the app pins #47's head) now refuses any package that does not restore every bundle row the round holds, whether or not the row carries a hash, since the hash is stored only after the broadcast returns and its absence proves nothing. A package may still extend a prefix restored earlier, and an empty round still accepts one; the end-to-end prefix test now states that it covers exactly that shape. The changelog no longer promises a partial restore.

On the chain lookup: it is not possible with what the app has. The registration probe needs the stored transaction hash, and witness generation needs the stored leaf position, which is written only after that probe confirms. A hash-less bundle has neither, and the crate has no lookup of a leaf by value. Adding one over the synced tree is a small SDK change that would restore partial recovery over a longer live round and unlock POLL-06 at the same time; that is the follow-up.

_Posted with Claude Code assistance._

LukasKorba (approved):
utACK

PR #2019, opened 2026-08-26
Preserve the voting database before anything opens it

dannywillems:
Preserves the voting database before anything opens it. Split out of #2018 so
it can ship on its own, ahead of any restore path.

## Why this is urgent and standalone

An earlier build could delete and recreate a poll whose delegation had already
been broadcast, discarding `van_comm_rand` — a 32-byte blinding factor sampled
from `OsRng` in the Rust core, never derived from the seed, and unwritable
through the FFI. Once gone, the poll is permanently unvotable.

In WAL mode, frames written before the wipe can still contain the intact row.
After checkpointing removes the WAL, deleted record bytes can also remain in
freed pages of the main database until SQLite overwrites them. Both files are
therefore evidence worth preserving.

The capture runs before `old.close()` and before opening a new SQLite
connection, because those lifecycle operations can checkpoint or unlink the
WAL.

## The logic

> If `voting_recovery/voting.sqlite3` holds no data, copy `voting.sqlite3` and
> any existing `voting.sqlite3-wal` and `voting.sqlite3-shm` sidecars into it.

`VotingDatabaseSnapshot.capture` runs at the very top of
`DatabaseActor.open`. It performs byte copies without opening either the source
or destination through SQLite.

The recovered main database is the idempotence guard: once a non-empty copy
exists it is never replaced, preserving the earliest capture. A zero-byte
leftover does not count as data and is replaced on the next attempt. Emptiness
is determined from file size, never by querying SQLite.

When a WAL exists it is copied first because it is the most volatile artifact.
The main database is still captured when there is no WAL because freed pages
may contain the deleted bundle record required by the later forensic recovery
path. SHM is copied when present.

The copies retain SQLite's canonical names so the trio remains mutually
resolvable. Capture time is recorded by a
`captured-<yyyyMMdd-HHmmss>.txt` marker.

Failures are currently logged and swallowed so capture cannot stop the voting
flow from opening. The copies remain eligible for system backups, like the
source voting database, and are removed with wallet-scoped voting state during
wallet reset.

## Validation

Tested against the affected database shared by the team:

- the first run captured the trio and later runs skipped;
- scanning the preserved WAL reproduced all three bundles' original
  `van_comm_rand`, `gov_comm`, `alpha`, `rk`, and `pczt_sighash`;
- the source trio remained byte-for-byte unchanged.

Unit coverage also pins capture of a main database when no WAL exists. Swift
parsing and diff hygiene pass locally. GitHub CI currently fails during package
resolution before compilation.

That affected database contains live voting material and is not committed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ directly impact users rather than highlighting other crucial architectural updat
 ## [Unreleased]
 
 ### Added
+- [Ironwood] Before opening the voting database, ZODL now preserves a copy of it under `voting_recovery`. If an earlier version already discarded a poll's delegation, that copy keeps what is needed to restore it later — even if you upgrade long after the poll was affected. It is taken only once, so the earliest and most complete copy is the one kept; it is included in device backups so it survives a migration, and is removed when the wallet is reset.
 - [MOB-1466] Advanced Settings now offers "Restart Migration" while a migration is running. It shows how much has already migrated and how much is left, warns that restarting cancels the current plan for good, and asks for a separate confirmation before anything happens. Transfers that already went out are untouched — they stay migrated — and once the plan is cancelled you set up a new one for the remaining balance the same way you did the first time.
 - [MOB-1466] With a privacy migration in progress, keeping the app open now advances the migration automatically — transfers send on schedule without reopening the app.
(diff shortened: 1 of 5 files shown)


Chlup (approved):
utACK

PR #2020, opened 2026-08-26
[MOB-1798] Skip witness rebuild after partial voting recovery

czarcas7ic:
Before, reopening a partially recovered multi-bundle poll refetched tree state and regenerated every witness even though the persisted round had already completed that setup.

After, any recovered bundle proves setup completed, so ZODL reuses the persisted witness work. Rounds with no recovered bundles still run the idempotent completion path to preserve crash recovery.


Chlup (changes requested):
Reviewed the diff at max effort (multi-angle finder pass + independent verification of every candidate + a gap sweep). One correctness issue is blocking; the rest are cleanup, altitude, test-coverage and changelog items.

The change itself is sound in its core claim — witness/tree-state rows do persist in the voting DB, and I could not construct a state where a confirmed delegation coexists with missing witnesses. The blocking issue is a side effect of *how* the skip is applied, not of the idea.

---

## Blocking

### 1. Partial recovery now bypasses the `inconsistentBundleSetup` guard

`secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift:936`

Gating `completeDeterministicRoundSetup` also skips the guard inside it at line 3484:

```swift
guard Int(bundleCount) <= noteChunks.count else {
    throw VotingFlowError.inconsistentBundleSetup(bundleCount: bundleCount, noteChunkCount: noteChunks.count)
}
```

That was the only round-entry check that the persisted bundle count still matches the wallet's current note chunking. On `main` it ran on every partial-recovery resume.

Failure path — round persisted with `existingBundleCount == 3`, a later rescan/reorg leaves `notes.smartBundles().bundles.count == 2` at the fixed snapshot height, bundle 0's delegation is confirmed so `recoveredBundleCount == 1`:

1. No throw at round entry. `.votingWeightLoaded` records `bundleCount: 3` (line 1061).
2. `runDelegationPipeline` ignores the persisted count and recomputes its own — `let bundleCount = UInt32(noteChunks.count)` (line 3764) — delegates bundles 0–1 only, and still sends `.delegationProofCompleted` (line 3882).
3. The vote loop iterates the *persisted* count, `0..<session.bundleCount` == `0..<3` (lines 1764, 1889), and reaches `generateVanWitness(roundId, 2, anchorHeight)` (line 1917) with no VAN row for bundle 2 — throwing **after** bundles 0 and 1 were already broadcast by `submitVoteCommitment`.

Net effect: silently under-delegated voting weight, and a partial ballot that fails late instead of failing closed at entry. The Keystone path fails late too, at the per-bundle prep guard (line 2650) or `reduceKeystoneAllBundlesSigned` (line 2878), i.e. after hardware signing has started.

Cheapest fix that keeps the optimization: hoist the bundle-count consistency check out of `completeDeterministicRoundSetup` so it runs on the resume path regardless of the gate.

---

## Correctness-adjacent

### 2. No behavioural test of the only branch whose behaviour changed

`zodlTests/VotingTests/VotingCoordFlowCoordinatorTests.swift:734`

The added test pins the pure predicate's truth table only. The diff swapped the two branch bodies (`main`: `if delegationReady { witnesses = [] } else { complete… }`), and nothing in `zodlTests/` references `startActiveRoundPipeline`, `votingWeightLoaded`, `recoverDelegationVanPosition` or `clearRecoveryState` — the whole `.run` effect (lines 899–963) is untested. A rebase that resolves this hunk with the old polarity would make setup run exactly when it must be skipped, with every test green.

`interruptedPersistedSetupRetriesOnlyDeterministicWork` (line 753) already owns the harness: `RecoveryOrderRecorder` is wired to `storeTreeState` / `generateNoteWitnesses` / `getTreeState` and emits `get-tree:` / `store-tree:` / `witness:` events. A sibling test with `getBundleCount == 2` and `getDelegationTxHash` present for bundle 0 only, asserting no witness events and that `.votingWeightLoaded(delegationReady: false)` still arrives, would pin the actual change.

### 3. The gate reads a network probe instead of the local fact it stands in for

`secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift:3372`

`recoveredBundleCount` comes from `recoverDelegationVanPosition(…, confirmationTimeout: 0, retryDelay: .zero)` under `try?` (lines 902–907). With that timeout the deadline is `Date()`, so `guard Date() < deadline else { return .notFound }` (line 3964) fires on the first miss — one attempt, no retry — and the live `fetchTxConfirmation` returns nil rather than throwing on network error, HTTP 404 "not yet mined", 5xx, and malformed JSON.

Consequences:

- A force-quit after witness generation but **before** any delegation was submitted gives `recovered == 0` and pays the full 30–120 s rebuild. That is the most likely interruption window, and the one af0932ea targeted.
- The same round with identical persisted state skips on wifi and rebuilds on a flaky link.

The error direction is safe (extra work, never corruption), so this is a mis-targeted optimization rather than a regression — but it means the change often does not fire when it should. `zcash_voting` already exposes the authoritative per-bundle predicate `has_witnesses(round_id, bundle_index)` (`storage/operations.rs:658`), which is offline and deterministic. It is not currently plumbed through the SDK FFI or `VotingCryptoClient`, so adopting it needs an SDK change — worth a follow-up rather than blocking this PR.

---

## Cleanup / altitude

### 4. The new doc comment overstates the hazard

`VotingCoordFlowCoordinator.swift:3369`

> A recovered bundle proves the persisted round completed tree-state and witness setup before delegation submission began.

This reads as *why re-running would be wrong*, but the target function's own doc at line 3465 says it does "only deterministic tree-state and witness work" and "must not clear the round or rebuild delegation authorization" — an explicit re-run-safety guarantee. That holds in the crate too: `store_witnesses` early-returns when the cached count already matches (`storage/operations.rs:708`) and `store_van_position` is an idempotent `UPDATE` (`storage/queries.rs:1346`). The real justification is latency, and the comment never says so — which will make a future witness-repair path harder to argue for.

The step-2 header comment at lines 855–858 ("The Rust backend needs a round row, bundle rows, tree state, and witnesses before Keystone PCZT prep or inline delegation") goes stale with it: on this path step 2 no longer establishes that invariant, it inherits it from a previous launch.

### 5. `cachedWitnesses` is write-only dead state with a stale doc

`secant/Sources/Features/CoordFlows/VotingCoordFlow/RoundSession.swift:42`

Two references repo-wide: the declaration, and the single write at `VotingCoordFlowCoordinator.swift:1061`. Zero reads. The doc claims "Used as authPath input during vote commitment build (Stage 5B)", but the authPath fed to `commitVote` comes from `generateVanWitness(...).authPath` at line 1917.

This matters for reading the diff: the stale comment makes `else { witnesses = [] }` look like it discards something load-bearing, when the branch's only real content is whether `storeTreeState` + `generateNoteWitnesses` write to the DB. Returning `Void` from `completeDeterministicRoundSetup` and dropping the field plus the action payload collapses the hunk to:

```swift
if recoveredBundleCount == 0 {
    try await Self.completeDeterministicRoundSetup(...)
}
```

### 6. `existingBundleCount > 0` is dead at the only call site

`VotingCoordFlowCoordinator.swift:3374`

The call at line 936 is inside `else if Self.shouldResumePersistedRound(existingBundleCount:)` (line 897), and that helper *is* `existingBundleCount > 0` (line 3367). So the term is always true and the predicate reduces to `recoveredBundleCount == 0`, while two places now encode the same threshold and can drift.

`VotingCoordFlowCoordinatorTests.swift:747-750` then pins the unreachable `(0, 0)` case. The zero case is also the wrong default to bake in: `shouldCompleteDeterministicRoundSetup(existingBundleCount: 0, recoveredBundleCount: 0) == false` says "a round with no bundles needs no setup", the opposite of what a fresh round needs — so a future caller reusing this around `prepareFreshRound` / `setupBundles` would silently skip everything.

### 7. Four separate policies now decide which parts of setup to run

`VotingCoordFlowCoordinator.swift:879`

Inside one `.run` effect (lines 826–1013) the decision is taken at line 879 (`proofGenerated` → none), line 918 (all recovered → none + `clearRecoveryState`), line 936 (this new gate), and line 965 (fresh → all). Each new case re-derives the same invariant about what the DB already holds, and only the newest is a named, tested predicate — which reads as though the rule were verified, when what is verified is a tautology over the predicate's own inputs.

Both underlying writes are already `INSERT OR REPLACE`, and the crate models the deeper fix: `PreparedDelegationBundle::ensure_witnesses` (`delegate.rs:594-617`) "persists bundle witnesses when they are missing". Pushing the skip *into* `completeDeterministicRoundSetup` — skip per-bundle work whose witnesses exist, skip `getTreeState` when a tree-state row exists — lets all four sites call it unconditionally, and fixes the `proofGenerated` branch and `prepareFreshRound` at the same time.

### 8. The recovery loop is N sequential network round-trips on the blocking path

`VotingCoordFlowCoordinator.swift:901`

`for bundleIndex: UInt32 in 0..<existingBundleCount` issues one `fetchTxConfirmation` per bundle, strictly serialized, all before `.earlyEligibilityConfirmed` (line 934) hands navigation off. `fetchTxConfirmation` walks every configured vote server, and the non-fast session uses `timeoutIntervalForRequest = 120` (`VotingAPIClientLiveKey.swift:177`; the Tor path ignores request timeouts entirely). On a captive network a 4-bundle round can sit on the polls list for minutes, then conclude "nothing recovered" and start the rebuild anyway.

The probes are read-only — run them concurrently in a `withTaskGroup` keyed by `bundleIndex` and apply the `storeVanPosition` writes sequentially afterwards to keep DB writes and log ordering stable. There is also no `Task.checkCancellation()` in this loop, unlike the sibling at line 2119.

---

## Changelog

### 9. Entry is missing its issue prefix and is written in implementation terms

`CHANGELOG.md:45`

`CLAUDE.md` → `## Changelog`: "Prefix every line with the issue identifier in brackets, e.g. `- [MOB-1321] Short, user-facing description of the change.`" 41 of the 44 bullets under `## [Unreleased]` carry one; the `+` line here does not.

Same rule block: "Write from the user's perspective — what changed for them and why it matters — not the implementation detail." "preserves persisted delegation setup and completed witness work" names internals a user cannot observe, unlike its neighbours ("no longer stays stuck showing 'Sending…'").

Section placement is fine — the line is under `## [Unreleased]` → `### Fixed`, so editing in place loses nothing.

### 10. The claim is broader than what ships

`CHANGELOG.md:45`

> reopening a poll preserves persisted delegation setup and completed witness work needed to retry safely.

Witness work is preserved only when a delegation transaction is found confirmed *on that launch*. Per issue 3, a force-quit after witness generation but before submission still rebuilds everything, and so does a fully-delegated round whose confirmation lookups fail offline.

---

## Checked and cleared

Recording these so they don't get re-litigated — each was chased and refuted with evidence:

- **Permanent loss of the witness repair path.** `store_witnesses` early-returns when the cached count matches, writes are validated against the round's immutable `nc_root`, and an interrupted setup aborts before `.votingWeightLoaded` — so `recoveredBundleCount == 0` and repair is still permitted.
- **Duplicate delegation from an under-counted probe.** `runDelegationPipeline` re-probes every bundle with the full 90 s / 2 s poll and `continue`s over confirmed ones (lines 3766–3783).
- **Cancellation hazard from the removed `getTreeState` suspension point.** The double-tap trigger is disabled in `PollsListView.swift:317`, and a double keychain import throws `alreadyImported` with the first value intact.
- **`clearRecoveryState` running before the eligibility guard.** On the pinned `zcash_voting =3.0.0` the clear is conditional on `van_leaf_position IS NULL`, and `storeVanPosition` always precedes the count — so the hashes survive. (This one *would* go live if the pin ever moved back to 2.0.0-rc.3, which clears unconditionally.)
- **Keystone `completedKeystoneDelegationBundleIndices` rewind.** Defended by the `bundleCount == 0 ||` escape hatch at line 634 and re-derived from persisted signatures; also pre-existing.
- **`buildVotingPczt` precompute wedge** at lines 2131 and 2682. Real, but pre-existing on `main` and unaffected by this diff — the precompute loops lack the "Finding #10" probe that `runDelegationPipeline` (lines 3789–3816) documents as mandatory. Filing separately.

Only issue 1 blocks. 2 and 3 are worth doing here; the rest can be split off if you'd rather keep this PR tight.

PR #2021, opened 2026-08-26
[MOB-1798] Stop deleting rounds to make an INSERT succeed

dannywillems:
Removes the `clear_round` call from the app. Targets `release/3.10.2`, independent
of #2018 and #2019.

## Why it was there

`prepareFreshRound` opened with:

```swift
try? await votingCrypto.clearRound(roundId)
try await votingCrypto.clearRecoveryState(roundId)
...
try await votingCrypto.initRound(params, nil)
let setupResult = try await votingCrypto.setupBundles(roundId, notes)
```

`clear_round` deletes the `rounds` row, and `bundles` cascades away with it —
taking `van_comm_rand`, the 32-byte VAN blinding factor. That value is sampled
from `OsRng` in the Rust core, is never derived from the seed, and no exported
FFI call can write it back. Losing it makes the round permanently unvotable
even though its VAN is already on chain.

Nothing wanted that deletion. It was there so the two calls after it would
succeed: `initRound` reaches a plain `INSERT INTO rounds` and `setupBundles` a
plain `INSERT INTO bundles`, neither with `OR REPLACE` or `ON CONFLICT`. The
wipe was a delete-so-I-can-re-insert idiom; the destruction was incidental.

## Why it is unnecessary

`shouldResumePersistedRound` (from #2016) already keeps callers off this path
once bundle rows exist. So every path that legitimately reaches
`prepareFreshRound` has **no bundles** — nothing to delete, and `setupBundles`
cannot conflict.

The only genuine conflict left is the `rounds` row itself, when setup was
interrupted between the two inserts. That row is now reused, and `initRound` is
called only when it is actually absent:

```swift
switch canReuseExistingRoundRow(existingState: existingState, snapshotHeight: snapshotHeight) {
case .none:       try await votingCrypto.initRound(params, nil)
case .some(true): break
case .some(false): throw VotingCoordFlowError.roundParametersChanged(roundId: roundId)
}
```

`existingState` is already computed by the caller via `loadExistingRoundSetup`,
so this costs no extra database read.

A surviving row whose parameters no longer describe the same round is a real
inconsistency — the round's parameters are what any later delegation and proof
bind to. That now throws rather than being resolved by deletion: a mismatch is
something to report, not something to clear.

## `clearRound` is removed from the client

Deleted from `VotingCryptoClient` rather than left unused, so app code cannot
reach it again. Leaving it on the interface would leave the loaded gun on the
table. `zcashlc_voting_clear_round` stays in the FFI for genuine
user-initiated round deletion.

## Scope

Three tests on `canReuseExistingRoundRow`, following the `shouldResumePersistedRound`
pattern from #2016. No Rust changes, no XCFramework rebuild.

## Follow-ups, not in this PR

- The crate already has `ensure_round` — *"Ensures a round exists… Existing
  rounds are left unchanged"* — which would replace the switch above with one
  idempotent call. It is not exposed through the FFI, and as written it
  compares only the network, not the full parameter set, so exposing it should
  come with that widened.
- `clear_unsigned_delegation_setup_fields` is also unexposed. It would give the
  app a real "replan the unsigned setup" operation for the case where bundles
  *do* exist — the thing `clearRound` was being misused for. Its `WHERE` clause
  already refuses to touch a row carrying a `delegation_tx_hash` or
  `van_leaf_position`.
- This makes the `prepareFreshRound` guard in #2018 redundant; that commit
  should be dropped from #2018 if this lands. The recovery work in #2018 is
  unaffected and still needed for rounds already damaged by 3.10.1 and earlier.

Not built here — no Xcode in the environment this was written in.


Chlup (changes requested):
## Summary

The diagnosis is correct and I verified it end-to-end against the pinned `zcash_voting = "=3.0.0"` sources rather than taking the description on trust: `clear_round` really is a bare `DELETE FROM rounds` (`storage/queries.rs:419`), `bundles` really does declare `ON DELETE CASCADE` (`001_init.sql:41`), `van_comm_rand` really is `pallas::Base::random(&mut rng)` (`action.rs:427`) with no FFI writer, and `insert_round`/bundle setup really are plain `INSERT`s while neighbouring tables use `INSERT OR REPLACE`. So the "delete-so-I-can-re-insert" reading of the old code holds, and removing the closure from the client is the right instinct.

Requesting changes on the implementation, not the direction. The blocking issues are: the new error path creates a state the app cannot leave, and the guard test that carries this PR's "never again" promise can pass green having checked nothing.

---

## Blocking

### 1. `.parametersChanged` is a permanent dead end, and the string promises a remedy that cannot work

`VotingCoordFlowCoordinator.swift:3463`

The throw reaches `catch:` at `:1028` → `.pipelineFailed` → `:1117-1130` sets `state.rootScreen = .error(message)`. `VotingCoordFlowView.swift:157-162` renders `.error` **without** a `recoveryAction` (unlike `.configError` just below it), so `VotingErrorView` offers only "Got it" → `.dismissFlow`. Reopening re-runs the identical comparison against the identical row and throws again. `initRound` is skipped on this branch, `clearRound` is gone from the client, and the only remaining removal of the row anywhere is the full-wallet-wipe `removeItem(at: votingDbURL)` in `RootStore.swift:745`.

The new string says: *"Close it and open it again, and contact support if the problem continues."* Closing and reopening is exactly the loop.

On `main` this state self-healed — `clearRound` + unconditional `initRound` deleted and re-inserted the row at the new height — so this PR **introduces** the dead end rather than preserving one.

It is also disproportionate on this exact path: `prepareFreshRound` is reached only when `existingBundleCount == 0` (`:897` + `shouldResumePersistedRound` at `:3375-3377`), and `van_comm_rand` is a `bundles` column. There is provably nothing to protect where the new code now hard-fails.

The 0-bundle precondition is easy to reach, incidentally: `ensureRoundRow` inserts, then `guard bundleCount > 0, eligibleWeight > 0 else { … return false }` at `:3508` returns early for any user below `ballotDivisor`, leaving a permanent 0-bundle row behind.

Suggested: reuse-with-warning when `bundleCount == 0`, or give `.error` a recovery action. At minimum the string must not promise a remedy the code makes impossible.

### 2. The source-guard test passes green after scanning zero files

`VotingRoundsTablePreservationTests.swift:164-173`

I ran this rather than reasoning about it. `FileManager.default.enumerator(at:includingPropertiesForKeys:)` on a **nonexistent** directory returns a **non-nil** enumerator yielding **zero** URLs, so the `guard let walker … else { return [] }` is dead code and never fires:

```
[nonexistent-deep]   enumerator: NON-NIL, total=0, swift=0
[real-sources]       enumerator: NON-NIL, total=708, swift=538
```

Replicating the exact test body:

```
[REAL source tree]   scanned=535 offenders=0 -> PASS
[ABSENT source tree] scanned=0   offenders=0 -> PASS
```

Identical green result from 535 files and from zero. `#filePath` is baked in at compile time, so this triggers the first time the bundle runs where the tree is not at that path — CI split across `build-for-testing` / `test-without-building` runners, a relocated checkout, a device run. A guard for an invariant the suite itself calls unrepairable would then be silently protecting nothing.

Today's CI is fine (single job, simulator shares the host filesystem), so this is latent, not live. The fix is one line, and the sibling test already has it: `#expect(declarations.contains { $0.hasPrefix("var initRound") })`, commented *"The client is still the real one, not an empty file we mis-parsed."* That defence was applied to one test of the pair only.

### 3. The whitelist declares `deleteSkippedBundles` safe, but it destroys `van_comm_rand` directly

`VotingRoundsTablePreservationTests.swift:92-93` — *"`clearRecoveryState` and `deleteSkippedBundles` are deliberately absent from the forbidden list: neither touches the `rounds` table."*

Literally true, but the suite grounds its own existence in `van_comm_rand` (lines 11-14), which lives on `bundles` and is reached from `rounds` only via the cascade. `deleteSkippedBundles` → `VotingRustBackend` → `delete_bundles_from` runs:

```sql
DELETE FROM bundles WHERE round_id = :round_id AND wallet_id = :wallet_id AND bundle_index >= :from_index
```
(`queries.rs:2395`)

The only guard there (`:2385`) is `round_has_imported_capability_bundles`, which checks `note_positions_blob IS NULL` — imported-capability atomicity. Nothing checks `delegation_tx_hash` or `van_leaf_position`, so locally prepared Keystone rounds are unprotected.

The call site is not provably safe either. `clear_recovery_state` deletes **all** `keystone_signatures` but nulls `delegation_tx_hash` only where `van_leaf_position IS NULL`, so recovery can yield a non-contiguous resolved set (e.g. `{0, 2}` → prefix 1); `DelegationSigningStore.swift:478-479` gates only on `resolvedPrefix > 0`; `deleteSkippedBundles(roundId, 1)` then deletes bundle 2 whose VAN is on chain. Corroborating: `VotingCoordFlowCoordinator.swift:3031-3032` already filters `completedKeystoneDelegationBundleIndices` to `< signedCount`, i.e. the code anticipates on-chain delegations above the prefix and silently discards the record.

The underlying hazard is pre-existing and out of scope here. What is new and wrong is the sentence telling the next reader that a `van_comm_rand`-destroying operation is safe.

---

## Worth fixing

### 4. The comment added above the resume-path `clearRecoveryState` overstates safety

`VotingCoordFlowCoordinator.swift:919-931`. The call is unchanged from `main`; the PR adds the comment.

`clear_recovery_state` runs four statements (`queries.rs:2826-2865`): two **unguarded, whole-round** `DELETE`s (`share_delegations`, `keystone_signatures`) and two position-guarded `UPDATE`s. The SDK's own test `clear_recovery_state_preserves_recorded_vc_position` (`recovery.rs:710-760`) pins the asymmetry, noting *"the unconditional lanes still empty out"*.

Unlike the identical call in `prepareFreshRound`, this site is reached only when `existingBundleCount > 0`, so the cascade-emptiness reasoning the comment borrows from the interface doc does not apply here. "The confirmations themselves survive" is false for share-delivery confirmations: a VC-tree-confirmed vote survives with `submitted == true` while all its share rows are gone — which is what `undeliveredShareProposalIds` (`:3283-3302`) flags, feeding synthetic recovery drafts at `:1735-1741`.

I could not construct a reachable state where `proofGenerated == false` coexists with confirmed votes, so I can't show live harm — you're better placed to say. Either way the comment's justification is delegation-phase reasoning applied to vote-share-delivery state.

### 5. A reused row never refreshes `ea_pk` / `nc_root` / `nullifier_imt_root`

`VotingCoordFlowCoordinator.swift:3439`

To be clear about what I checked and rejected: the scary version of this is **not** true. `ea_pk` cannot silently corrupt a tally — `RoundAuthenticator.swift:68` verifies the chain-reported `ea_pk` against an Ed25519 signature binding round id to EA key on every session fetch, and `ea_pk` is a public proof input (instance offsets 9-10) absent from every submission payload, so a stale key means a rejected `cast-vote` POST. `nc_root` and `nullifier_imt_root` fail loudly too (`witness.rs:216`, `operations.rs:309`, `operations.rs:799`).

What remains is recoverability. After an admin-signed EA rotation, or a root changing at a stable height, the stale row is `.reusable`, is never rewritten, and — since every delete path is gone — can no longer self-heal. The user gets an opaque, permanent failure where `main` silently recovered. The code documents the narrowed comparison at `:3427-3433`; I'd just like it confirmed as an accepted trade rather than an oversight.

### 6. The new check guards the branch with nothing to lose, not the one holding the bundles

`VotingCoordFlowCoordinator.swift:897`

`shouldResumePersistedRound` takes every round **with** bundles down a path that passes the session's `snapshotHeight` straight into `completeDeterministicRoundSetup` without ever comparing it to `existingState.snapshotHeight`. With 3 bundles at `snapshot_height = 100` and a session reporting 101, `storeTreeState` overwrites `cached_tree_state` via `INSERT OR REPLACE` (`queries.rs:1740`) **before** `generateNoteWitnesses` trips the height check in `validate_cached_tree_state_for_round` (`witness.rs:205-211`) — destroying a valid cached tree state on the way to failing.

The effort is inverted relative to the stakes: the validated branch has no `van_comm_rand`; the unvalidated one does. Pre-existing path, but this PR is what establishes the check and chose where to put it.

### 7. `ensureRoundRow` has three parameters encoding one truth

`VotingCoordFlowCoordinator.swift:3450`

`snapshotHeight` is exactly `params.snapshotHeight`, and `roundId` is exactly `params.voteRoundId.hexString` — `VotingCryptoClientLiveKey.swift:52-57` derives both the row's identity and its height from `params` alone. So classification reads the loose values while the insert is built entirely from `params`, with nothing enforcing agreement. A caller passing `params` at 100 with `snapshotHeight` 101 inserts a row at 100, and every later open classifies `.parametersChanged` — uncorrectable, per #1.

The foot-gun is already firing harmlessly in the tests: every call pairs `roundId: "round-1"` with `params` whose `voteRoundId.hexString` is `4a4a…`. Nothing detects it.

Suggest dropping both loose parameters and deriving them from `params` inside. `classifyExistingRoundRow`'s two-argument shape is fine — it's directly unit-tested and correct at that level.

### 8. The suite's headline invariant is false as written

`VotingRoundsTablePreservationTests.swift:8` — *"a `rounds` row, once written, is never deleted or rewritten by the app."*

The row is rewritten during ordinary operation: `operations.rs:1063` and `:1161`, both inside `build_and_prove_delegation`, call `advance_round_phase` → `UPDATE rounds SET phase = :phase` (`queries.rs:192`). Neither guard covers rewriting, and none could, since the phase must advance. Narrow the claim to deletion, which is what the tests and the `van_comm_rand` rationale actually support.

### 9. The `VotingStorageClient` exemption is wrong in both directions

`VotingRoundsTablePreservationTests.swift:160-162`

**False negative:** the comment says matching the directory means *"a rename of the file alone cannot silently widen the exclusion"*, but a directory match is strictly *broader* than a filename match. Every file added to `Dependencies/VotingStorageClient/` is exempt forever — including a future persistence helper calling `backend.clearRound(roundId:)`, in the one directory whose name contains "Storage".

**False positive:** `VotingStorageClient.clearRound` (`VotingStorageClientInterface.swift:20`) is real, fully wired, and has zero callers anywhere — the entire dependency is unconsumed today. Its first consumer must live outside that directory, so `await votingStorage.clearRound(roundId)` matches the receiver-agnostic substring test and fails with *"App code must never delete a voting round"* — about three `removeValue(forKey:)` calls on in-memory dictionaries.

Measured aside: a `func clearRound(` declaration in app code also trips the scan, while a `var clearRound:` declaration does not — the inverse of what a reintroduction guard wants. Matching a receiver (`votingCrypto.` / `backend.`) fixes both directions.

### 10. `hasPrefix("var ")` misses any modifier or attribute

`VotingRoundsTablePreservationTests.swift:132-140`

All 52 declarations in `VotingCryptoClientInterface.swift` are bare `var `, so today's pass demonstrates nothing about robustness. `private var clearRound:`, `public var clearRound:`, `private(set) var`, `@available(*, deprecated) var` are all invisible, and both negative assertions pass vacuously with the `initRound` canary still green.

This is the one guard the compiler does **not** back up — re-adding the property compiles fine. Two things blunt it: the likeliest reintroduction (a straight re-add or `git revert` of the deleted line) *is* caught, and any actual *use* trips the sibling scan. So the gap admits only a declared-but-unused member — which offends the test's own "an available closure is an invitation" rationale, but isn't the destructive outcome.

---

## Structural / nits

### 11. `VotingCoordFlowError` duplicates `VotingFlowError`

`VotingCoordFlowCoordinator.swift:4106` vs `VotingHelpers.swift:365`, an enum literally headed *"Pipeline errors (used by VotingCoordFlow)"*. Same `LocalizedError` + `errorDescription` shape, same `#if VOTING_ENABLED`, 16 throw sites in this very file, and both reach the user through the same `catch:` → `.pipelineFailed`. Decisively, both are thrown from the *same function*: `prepareFreshRound` throws the new one via `ensureRoundRow` and `VotingFlowError.inconsistentBundleSetup` via `completeDeterministicRoundSetup` (`:3562`).

I checked the two reasons not to merge and both fail. `VotingFlowError` lacks `Equatable`, but all associated values are `UInt32`/`Int`/`String` so synthesis applies — and `ShareDelegationError` (`VotingAPIClientInterface.swift:38`) already uses `LocalizedError, Equatable, Sendable`. And `VotingErrorMapper.userFriendlyMessage` switches only over `ShareDelegationError`, so adding a case forces no update. It's a ~6-line move plus one word.

### 12. Three tests and the fixtures are duplicated across both test files

`anAbsentRoundIsInsertedAndNothingElseHappens` (`:30`) ↔ `absentRoundRowIsInsertedExactlyOnce` (`VotingCoordFlowCoordinatorTests.swift:777`); `aSurvivingRoundIsReusedAndNothingIsWritten` (`:38`) ↔ `existingRoundRowIsReusedWithoutInserting` (`:761`); `unrecognisedRoundIsReportedRatherThanRemoved` (`:65`) ↔ `changedParametersThrowWithoutTouchingTheDatabase` (`:796`). Same fixtures (100/nil/101), same assertions (1/0/0), differing only in `RecoveryOrderRecorder` vs `LockIsolated`. The claimed extra assertion power from `@DependencyClient` is present in both, since both build the client identically. `roundState`/`roundParams` are byte-identical bodies.

`ensureRoundRow` now has six call sites across two suites. Dropping the three duplicates deletes the fixture duplication with them and leaves the new file as the pure source-guard suite its own doc says is its distinct value. (If they're kept, `zodlTests/AddKeystoneHWWalletTests/ZcashAccountsTestFixture.swift` is the existing shared-fixture precedent.)

### 13. The invariant is enforced one layer above where the danger lives

`VotingRustBackend.clearRound` is still `public` and `zcashlc_voting_clear_round` still exported, while every closure in `VotingCryptoClientLiveKey.swift` already opens with `let backend = try await dbActor.backend()`. The next person hitting `UNIQUE constraint failed: rounds.round_id` writes `try backend.clearRound(roundId:)` inside a LiveKey exactly as the original code did, with only a test-target grep in the way. The SDK's own doc at `rounds.rs:291` still recommends *"`zcashlc_voting_clear_round` and a complete re-import"*.

Related, and I think this is the real shape of the fix: the crate already has `VotingDb::ensure_round` (`round/mod.rs:244`), and the sibling table already gets the idempotent treatment — `zcashlc_voting_setup_bundles` calls `ensure_bundles`. Only `rounds` was given the non-idempotent `INSERT` + destructive `DELETE` pair, which is what forced the delete-to-insert idiom in the first place. Exposing `ensure_round` (with its network-only comparison widened) would collapse the enum, classifier, error case, localized string and both guard tests into one unconditional call. Guarding the `DELETE` in SQL (refuse when bundles carry `van_comm_rand`) would make the grep test unnecessary rather than load-bearing.

### 14. The guard scan is ~4x slower than it needs to be

Measured: `secant/Sources` is 538 files / 4.08 MB / 97,269 lines. At `-Onone` (the config the test target builds in): enumerate 3 ms, read all files 10 ms, **scan 290 ms** — 97% of the test is line scanning, not I/O, and it finds zero hits every run. Two contributors: `"\(symbol)("` is interpolated inside the innermost loop (293,112 needle allocations where 3 would do, ~26 ms), and `strippingComment` copies every `Substring` and runs a bridged non-literal search on all 97,704 lines to filter zero candidates (~49 ms).

A whole-file `contents.range(of: needle, options: .literal)` prefilter, entering the per-line loop only for files that hit (needed only for line numbers in the failure message), runs the same check in 74 ms and stops scaling with total app source size.

### 15. The localized key uses a bespoke `voting.` namespace

Counted over the catalogue (1397 keys): keys starting `voting.` = **1**, keys starting `coinVote.` = **297**. Call sites: `String(localizable: .voting…)` = **1**, `.coinVote…` = **190**. The structural sibling `VotingFlowError` puts all nine of its localized cases under `coinVote.store.error.*`, so the consistent name is `coinVote.store.error.roundParametersChanged`. Cheap to change now, awkward once translations ship against the key.

---

## Things I checked and did *not* find

Flagging these so they don't get chased:

- **`ea_pk` does not silently corrupt votes.** See #5 — it's signature-bound at session fetch and is a public proof input absent from every payload. A stale key means a rejected POST, not an undecryptable tally entry.
- **`#if VOTING_ENABLED` compiling away in Release is harmless.** It is Debug-only for `zodlTests`, with no project-level or xcconfig fallback — but `ENABLE_TESTABILITY` is Debug-only too, so a Release test build fails outright rather than passing green with zero tests. The wrapper is also the repo-wide convention (all five existing voting test files) and is required to compile.
- **A missing localization key cannot degrade at runtime.** `String(localizable:)` accessors are generated per key by `xcstrings-tool-plugin`, so a renamed or deleted entry is a compile error at the call site.
- **`await #expect(throws:)` is used correctly.** Current non-deprecated API, `Equatable` is satisfied, and synthesized equality genuinely compares the `roundId` payload — the assertion is not degenerate.
- **`clearRound` had no other caller** on `main` besides `prepareFreshRound`, so removing the closure breaks nothing.
- **The `preClearKeystoneSignatures` save/restore machinery is dead**, but that predates this PR — the `shouldResumePersistedRound` gate already on `main` is what killed it.

## Build state

CI is red for a stale-base reason rather than anything in the diff: both jobs die at `xcodebuild: Could not resolve package dependencies: '/Users/runner/work/zodl-ios/zcash-swift-wallet-sdk' doesn't exist`. The branch forked before `a12e66b6` ("Switch to new path for local SDK") and still points at `../zcash-swift-wallet-sdk`; `main` now uses `../zodl-swift-wallet-sdk`. The PR is also marked `CONFLICTING`, with `CHANGELOG.md` the only conflicting file — a rebase onto current `main` fixes both.

Worth noting nothing has compiled this branch yet: the description says it was written without Xcode, and CI never got past package resolution. All of the above is static analysis plus targeted experiments; the first real `xcodebuild test` run is still ahead.

Minor: the description says it targets `release/3.10.2`, but the base is `main` and that branch no longer exists on the remote (`release/3.10.3` does). Base looks right; the text needs updating. The note about dropping the redundant guard commit from #2018 is also moot now that #2018 is a draft that likely won't merge.

zodlTests/VotingTests/VotingCoordFlowCoordinatorTests.swift
```diff
@@ -726,6 +726,121 @@ import Testing
...
+    @Test func absentRoundRowIsClassifiedForInsert() {
+        #expect(
+            VotingCoordFlow.classifyExistingRoundRow(existingState: nil, snapshotHeight: 100)
+                == .absent
+        )
+    }
+
+    /// A row left behind by setup interrupted between `initRound` and
+    /// `setupBundles` is reusable. This is the case the old `clearRound` call
+    /// existed to paper over.
```
str4d: The AI is including a bunch of unnecessary justification in its doc comments (this is just one example), which is meaningless given that `clearRound` has been deleted and useless going forward (it just creates noise). Justifications for changes go into commit messages.

dannywillems: Agreed with this. Making a human pass now.

secant/Sources/Dependencies/VotingCryptoClient/VotingCryptoClientInterface.swift
```diff
@@ -303,7 +302,35 @@ struct VotingCryptoClient {
...
         _ proposalId: UInt32
     ) async throws -> (bundle: VoteCommitmentBundle, vcTreePosition: UInt64)?
-    /// Clear recovery state for a round (keystone sigs, TX hashes).
+    /// Forgets a round's *unconfirmed* submission attempts so they can be
+    /// retried, while preserving everything already confirmed.
+    ///
+    /// The name suggests a general reset; it is narrower and deliberately
+    /// asymmetric. Backed by `clear_recovery_state`, it runs four statements:
+    ///
+    /// - `DELETE FROM share_delegations` — the whole round, unguarded.
```
dannywillems: This is unnecessary prose. The single-line comment existing before this commit was enough. Revert this change.

secant/Sources/Dependencies/VotingCryptoClient/VotingCryptoClientInterface.swift
```diff
@@ -303,7 +302,35 @@ struct VotingCryptoClient {
...
+    /// and a vote that reached the vote-commitment tree keeps its record: the
+    /// two `UPDATE`s are guarded precisely so confirmed history is not
+    /// rewritten as unsent. Imported capability bundles are untouched for the
+    /// same reason — they have no local note selection, so
+    /// `note_positions_blob IS NOT NULL` excludes them.
+    ///
+    /// It never touches delegation secrets. `van_comm_rand`, `gov_comm`,
+    /// `alpha`, `rk` and the rest stay exactly as they were, as do the
+    /// `rounds` row and recorded ballot intent. Nothing here is destructive in
+    /// the way `clear_round` was: everything it removes is either
```
dannywillems: Mentioning clear_round is unnecessary. Remove.

secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -3422,8 +3423,68 @@ extension VotingCoordFlow {
...
+        existingState: RoundStateInfo?,
+        snapshotHeight: UInt64
+    ) -> ExistingRoundRow {
+        guard let existingState else { return .absent }
+        return existingState.snapshotHeight == snapshotHeight ? .reusable : .parametersChanged
+    }
+
+    /// Makes sure a `rounds` row exists for `params`, without ever deleting one.
+    ///
+    /// Inserting is the only thing the previous `clearRound` call achieved:
```
dannywillems: `clearRound` is mentioned here, but it is part of the history. We should remove it.

secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -4046,6 +4113,26 @@ private extension Array where Element == String {
     }
 }
 
+// MARK: - Round setup errors
+
+enum VotingCoordFlowError: Error, Equatable, LocalizedError {
+    /// A surviving round row describes different parameters than the session
+    /// now reports. Surfaced rather than resolved by deleting the round: the
```
dannywillems: > Surfaced rather than resolved by deleting the round: the
    /// delegation material a round accumulates cannot be regenerated, so a
    /// parameter mismatch is something to report, not something to clear.

This is unnecessary AI comment.

zodlTests/VotingTests/VotingCoordFlowCoordinatorTests.swift
```diff
@@ -726,6 +726,121 @@ import Testing
...
+    @Test func interruptedRoundRowIsClassifiedReusable() {
```
dannywillems, line 740: Obvious test. LGTM.

zodlTests/VotingTests/VotingCoordFlowCoordinatorTests.swift
```diff
@@ -726,6 +726,121 @@ import Testing
...
+        #expect(
+            VotingCoordFlow.classifyExistingRoundRow(
+                existingState: roundState(snapshotHeight: 100),
+                snapshotHeight: 100
+            ) == .reusable
+        )
+    }
+
+    @Test func changedSnapshotHeightIsClassifiedAsChanged() {
```
dannywillems, line 749: LGTM.

zodlTests/VotingTests/VotingCoordFlowCoordinatorTests.swift
```diff
@@ -726,6 +726,121 @@ import Testing
...
+            roundId: roundId,
+            phase: .initialized,
+            snapshotHeight: snapshotHeight,
+            hotkeyAddress: nil,
+            delegatedWeight: nil,
+            proofGenerated: false
+        )
+    }
+
+    private func roundParams(snapshotHeight: UInt64) -> VotingRoundParams {
```
dannywillems: This should be more generalized, using proptest.

zodlTests/VotingTests/VotingRoundsTablePreservationTests.swift
```diff
@@ -0,0 +1,116 @@
...
+    ///
+    /// `VotingStorageClient` is excluded because its identically named
+    /// `clearRound` drops entries from an in-memory actor — hotkeys, sessions
+    /// and delegation registrations held for the current session only. It never
+    /// reaches SQLite, so it is not the operation this guards.
+    @Test func noAppSourceCallsARoundDeletingOperation() throws {
+        let forbidden = ["clearRound", "deleteRound", "zcashlc_voting_clear_round"]
+        var offenders: [String] = []
+
+        for file in try Self.appSwiftSources() where !Self.isInMemoryStore(file) {
```
dannywillems: Removing this, as we don't actually care. It is mostly a `git grep`.

zodlTests/VotingTests/VotingRoundsTablePreservationTests.swift
```diff
@@ -0,0 +1,116 @@
+#if VOTING_ENABLED
```
dannywillems: Remove the whole file. This is part of the git history...

secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -3433,17 +3468,33 @@ extension VotingCoordFlow {
...
         )
-        try await votingCrypto.initRound(params, nil)
+
+        switch classifyExistingRoundRow(existingState: existingState, snapshotHeight: snapshotHeight) {
+        case .absent:
+            try await votingCrypto.initRound(params, nil)
+        case .reusable:
+            break
+        case .parametersChanged:
+            // Reached only when the round has no bundles yet (a round with
```
dannywillems, line 3485: I don't know how we should handle this case. This isn't happening right now.

zodlTests/VotingTests/VotingCoordFlowCoordinatorTests.swift
```diff
@@ -726,6 +726,46 @@ import Testing
         #expect(!isDelegationSigningTop(state))
     }
 
+    // MARK: - Round setup never deletes
```
dannywillems, line 729: These are obvious tests.

Chlup (approved):
utACK

dannywillems (commented):
utACK b6bb16a2c92e41dde18c0aa10d103734afda8a66

nuttycom (approved):
utACK; I'm not entirely happy with the way that this has to be woven in here instead of having a more robust state machine but this is the correct minimal fix AFAICT.

PR #2023, opened 2026-08-26
[MOB-1798] Recover ambiguous voting submissions

czarcas7ic:
Before, a software-wallet delegation retry could produce different transaction bytes and a different hash, so exact-hash recovery could not recognize the already-landed VAN. Vote retries could also recompute their selection inputs or switch share mode after restart.

After, ZODL keeps exact-hash recovery as the fast path, then reconciles the persisted VAN against a pinned commitment-tree scan. It locks vote choices and share mode before submission and verifies recovered cast-vote leaf positions before continuing.

Validated with the zodl-internal simulator test build and 71 voting tests.


secant/Sources/Dependencies/VotingCryptoClient/VotingCryptoClientLiveKey.swift
```diff
@@ -474,6 +474,10 @@ extension VotingCryptoClient: DependencyKey {
                 try backend.markVoteSubmitted(roundId: roundId, bundleIndex: bundleIndex, proposalId: proposalId, txHash: txHash)
                 publishState(backend: backend, roundId: roundId)
             },
+            resetVotingSessionState: { roundId in
+                let backend = try await dbActor.backend()
+                try backend.resetSessionState(roundId: roundId)
```
czarcas7ic, line 482: Dependency note: this call requires zodl-swift-wallet-sdk commit 3a3a0b5c38acc4fd46d2738c0d633282b59d6930. Use that SDK revision for integration and release builds until a release tag containing the API is available, then switch the sibling ../zcash-swift-wallet-sdk checkout to that tag.

It is needed, otherwise generating a Keystone QR code and exiting will cause an error on retry.

https://github.com/zodl-inc/zodl-swift-wallet-sdk/pull/21

nuttycom: Update: zodl-swift-wallet-sdk#21 was rebased onto current `candidate/4.1.0` and its history repaired (two commits did not build in isolation — see the review comments there), so `3a3a0b5c` no longer exists. The pin for integration and release builds is now the current PR #21 head, `37e2817af1fe3a0f216b6aa3232269f6a7fb877e`.

No code change is needed in this PR: `resetSessionState(roundId:)` is unchanged by the rework. For awareness, two other surfaces did change there — the never-shipped keys-taking sighash readback became the keyless `getStoredPcztSighash(roundId:bundleIndex:)`, and a per-bundle `clearKeystoneSignature(roundId:bundleIndex:)` was added — neither is used here (#2015 consumes both).

PR #2025, opened 2026-08-27
[MOB-1798] Recover Base64-decoded voting leaf indexes

czarcas7ic:
Before: some vote-server compatibility responses could turn a decimal delegation leaf index such as `3753` into non-ASCII text. The transaction was confirmed, but ZODL rejected the position and stopped submission.

After: iOS restores the decimal value only when normal parsing fails and the response is non-ASCII, while malformed ASCII remains rejected. Every delegation confirmation path uses the shared parser.


secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -3137,6 +3137,25 @@ extension VotingCoordFlow {
         return nil
     }
 
+    /// Some deployed `/tx` handlers opportunistically Base64-decode CometBFT
```
dannywillems, line 3140: If the server changes the response format, a good practice is to introduce a version field, to differentiate the different versions. Versioning at the endpoint level could be also a solution to support multiple versions.

IIUC the CHANGELOG entry that is modified, server responses have been changed.

nuttycom (approved):
utACK 703f098d5926a3af6de60ebc2946107f9db8c4d4

secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -3137,6 +3137,33 @@ extension VotingCoordFlow {
...
+    /// event text. A non-ASCII value is recovered only when it re-encodes to
+    /// the canonical decimal the server emits; all other values fail closed.
+    static func delegationVanPosition(from confirmation: TxConfirmation) -> UInt32? {
```
dannywillems, line 3143: Reviewer note - it seems that the related server code is located at:
- https://github.com/valargroup/vote-sdk/blob/main/api/handler.go#L374-L380
- https://github.com/valargroup/vote-sdk/blob/main/api/handler.go#L286-L298

dannywillems: I think that it would be more appropriate to actually normalize the values on the server side, to avoid every client to require supporting this change.

dannywillems: Additionally, a corresponding parsing method exists in Rust: https://github.com/valargroup/zcash_voting/blob/main/zcash_voting/src/confirmation.rs#L355-L370.

czarcas7ic: https://github.com/valargroup/vote-sdk/pull/444

Done here but wont be out till after this vote

PR #2034, opened 2026-08-27
[MOB-1810] Defer vote-server health checks off the polls-list load path

Chlup:
## Problem

Opening the Voting flow awaited a parallel `GET /shielded-vote/v1/status` probe of **every** configured vote server inside `configureURLs` before the rounds list was even fetched. One operator currently drops most Tor connections, and over Tor the probe is effectively unbounded: `TorClient.httpRequest` hands the request to the Rust FFI and ignores `URLRequest.timeoutInterval`, while the app pins `retryLimit: 3` — so a single dead operator cost 3× arti's internal timeouts and `probeAll` waited for the slowest server. After that, a 60 s background loop kept re-probing all servers forever. The probe data fed nothing: `healthyServers()` had zero call sites.

## Change

- `ServerHealthTracker.configure(serverURLs:fetcher:)` replaces `initialize` and **never probes**; the polls list now waits only on config + rounds.
- The 60 s loop is deleted. Probing is a one-shot, coalescing background sweep (`startProbeSweep()`), fired at active-poll entry (`.roundTapped` active branches and `.viewMyVotesTapped` for active rounds) and at batch-submission start, where it overlaps the delegation proof. Nothing ever awaits probe results, and a sweep cancelled by a config reload can neither read the fresh config nor record stale results into it.
- The health data is finally consumed — as **advisory ordering only**: `delegateShares` target selection and `resubmitShare`'s walk order prefer servers with a closed/half-open circuit (shuffled within groups). No server is ever filtered out; the share-spread privacy rule and POST-failure pruning are untouched, and an empty or all-open health view degrades to the previous pure-shuffle behavior. This mirrors how the other wallet implementations treat helper health.
- Corrected two misleading comments: the claim that per-request timeouts govern the Tor transport, and the stale "health is intentionally not consulted" note on the share path.
- One standalone commit fixes a pre-existing `zodlTests` compile failure on Xcode 26.6 (`#require` over multi-line throwing calls in `VotingDatabaseSnapshotTests`; hoisted into locals — test-only, also broken on `main`). It unblocked this branch's own test runs and is cherry-pickable if wanted separately.

## Testing

- New `ServerHealthTrackerTests` (10): configure-never-probes, per-server sweep coverage, coalescing, reconfigure-cancels-sweep (plus a 25-iteration flake hunt on that race), circuit-breaker semantics incl. cooldown re-admission.
- New `ShareTargetOrderingTests` (6): ordering partition properties, healthy-first target selection, dynamic backfill through a healthy-target POST failure (never-veto property), resubmission walk order.
- 5 new coordinator tests: flow init never sweeps; `.roundTapped` active / `.viewMyVotesTapped` active sweep; finalized rounds don't; batch submission sweeps before delegation.
- Full `zodlTests` suite: **1530 tests / 213 suites — TEST SUCCEEDED** (Xcode 26.6, iPhone 17 Pro simulator).


secant/Sources/Dependencies/VotingAPIClient/ServerHealthTracker.swift
```diff
@@ -36,37 +39,75 @@ actor ServerHealthTracker {
...
-        servers = Dictionary(uniqueKeysWithValues: serverURLs.map { ($0, ServerState()) })
+    init(cooldownInterval: TimeInterval = 30) {
+        self.cooldownInterval = cooldownInterval
+    }
+
+    // MARK: - Configuration
+
+    /// Replace the server map and probe fetcher for a freshly loaded service
+    /// config. Cancels any in-flight sweep (its target set is stale) and resets
+    /// every circuit to closed. Never probes — the polls list must never wait
```
nuttycom, line 74: nit, for follow-up: the terminology used in this utility (probe, sweep, etc.) are jargon to the point of being unclear - I need definitions for these words to understand what's being probed, what's being swept, and why; same thing for "fetcher".

secant/Sources/Features/CoordFlows/VotingCoordFlow/VotingCoordFlowCoordinator.swift
```diff
@@ -427,9 +438,23 @@ extension VotingCoordFlow {
                 } else {
                     statusPolling = .none
                 }
+                // MOB-1810: this entry point lands on the same active-round
+                // review screen as `.roundTapped`'s voted branch, so it needs
+                // the same background health sweep at poll entry — advisory
+                // ordering input for the share-resubmission walk; nothing
+                // awaits it.
+                let startHealthSweep: Effect<Action>
+                if state.allRounds.first(where: { $0.id == roundId })?.session.status == .active {
```
dannywillems, line 447: Reviewer note: only probing if the current round is the active one.

secant/Sources/Dependencies/VotingAPIClient/VotingAPIClientInterface.swift
```diff
@@ -86,5 +86,12 @@ struct VotingAPIClient {
     /// Query the Cosmos SDK TX endpoint for a confirmed transaction and its ABCI events.
     /// Returns nil if the TX is not yet in a block (404 or network error).
     var fetchTxConfirmation: @Sendable (_ txHash: String) async throws -> TxConfirmation?
+    /// Kick off a one-shot background health sweep of the configured vote
+    /// servers. Returns as soon as the sweep is spawned; never waits for probe
+    /// results. Submission effects fire this unconditionally as an advisory
+    /// hint, so `testValue` overrides it to a silent no-op in
+    /// VotingAPIClientTestKey — unimplemented-endpoint reporting would flag
+    /// every test that merely passes through those effects.
+    var startHealthProbeSweep: @Sendable () async -> Void
```
dannywillems, line 95: I'm wondering if the type is actually correct.

secant/Sources/Dependencies/VotingAPIClient/VotingAPIClientTestKey.swift
```diff
@@ -3,6 +3,13 @@ import ComposableArchitecture
 import Foundation
 
 extension VotingAPIClient: TestDependencyKey {
-    static let testValue = Self()
+    static let testValue: Self = {
+        var value = Self()
+        // Fire-and-forget advisory hint invoked unconditionally by submission
+        // effects — silent here so unrelated tests stay green; every other
+        // endpoint keeps the macro's unimplemented reporting.
```
nuttycom, line 10: This is unintelligible; test-only code but still.

secant/Sources/Dependencies/VotingAPIClient/VotingAPIClientInterface.swift
Chlup: What do you mean?

secant/Sources/Dependencies/VotingModels/VotingServiceConfig.swift
```diff
@@ -183,6 +183,14 @@ extension VotingServiceConfig {
...
             throw VotingConfigError.decodeFailed("pir_endpoints must contain at least one entry")
         }
+        var seenVoteServerURLs = Set<String>()
+        for endpoint in voteServers where !seenVoteServerURLs.insert(endpoint.url).inserted {
+            throw VotingConfigError.decodeFailed("vote_servers contains duplicate url: \(endpoint.url)")
+        }
+        var seenPirURLs = Set<String>()
+        for endpoint in pirEndpoints where !seenPirURLs.insert(endpoint.url).inserted {
+            throw VotingConfigError.decodeFailed("pir_endpoints contains duplicate url: \(endpoint.url)")
+        }
```
nuttycom, line 193: Why do we treat duplication as an error, instead of just deduplicating?

secant/Sources/Dependencies/VotingAPIClient/ServerHealthTracker.swift
```diff
@@ -36,37 +39,75 @@ actor ServerHealthTracker {
...
-        await probeAll()
+    // MARK: - Probe Sweep
+
+    /// One-shot background sweep of every configured server. Returns as soon as
+    /// the sweep Task is registered — callers never wait on probe results.
+    /// While a sweep is in flight further calls coalesce into it and return
+    /// nil; a call after completion starts a fresh sweep. Unconfigured (no
+    /// fetcher, empty map) is a no-op returning nil. The returned Task exists
+    /// so tests can await sweep completion.
+    @discardableResult
```
dannywillems, line 95: Oh, I see, the return value can be actually discarded, and that's why the type is Void. Is it a practice we accept in this codebase?

secant/Sources/Dependencies/VotingAPIClient/ServerHealthTracker.swift
```diff
@@ -36,37 +39,75 @@ actor ServerHealthTracker {
...
-    /// to a non-Tor `URLSession.shared`; that fallback would leak the
-    /// device's IP to every vote server every 60s if a future refactor ever
-    /// called probeAll before initialize.
+    /// The in-flight one-shot probe sweep, nil when idle. `startProbeSweep`
+    /// coalesces onto it; `configure` cancels it because its target set is stale.
+    private var sweepTask: Task<Void, Never>?
+    /// Monotonic sweep identity so a finished (or cancelled) sweep can only
+    /// clear its own registration, never a successor started after a
+    /// reconfigure.
+    private var sweepGeneration: UInt64 = 0
```
dannywillems, line 55: Reviewer note: okay, this is only a counter.

secant/Sources/Dependencies/VotingAPIClient/ServerHealthTracker.swift
Chlup: Sorry I am lost. On line 96 there is a function which is not Void: `func startProbeSweep() -> Task<Void, Never>? {`

nuttycom (approved):
utACK cead8ebc4d217bb3d3d3947f7810356db58f0d15 - most of the surface area here is actually test coverage; the functional changes look straightforward.

secant/Sources/Dependencies/VotingAPIClient/ServerHealthTracker.swift
```diff
@@ -36,37 +39,75 @@ actor ServerHealthTracker {
...
+            await self?.probeAll()
+            await self?.clearSweepTask(generation: generation)
+        }
+        sweepTask = task
+        return task
+    }
 
-        // Start background probing (replaces any existing loop)
-        startBackgroundProbing()
+    private func clearSweepTask(generation: UInt64) {
```
dannywillems, line 108: And this is used to actually clear a specific task when it actually finished.
`startProbeSweep` uses the nil value to actually start a new task.

dannywillems (commented):
Optimistic ACK

Issue #1591, opened 2025-12-09
[iOS Tech Debt] Refactor instances of .publisher in the TCA


true-jared:
The TCA acts weird with .publisher. For instance, generation of QR codes is an async operation that lacks modern API - Apple hasn't updated it. So the code is wrapped inside .publisher. Lately I found out it's not reliable anymore - .run works reliably. We have several publishers in place, even for streams.

PR #2036, opened 2026-08-27
[MOB-1818] Show the Max chip as loading while spendable value is unknown

### LINKED ISSUE #2035: [MOB-1818] Max chip reads as unavailable while spendable value is still being determined
Mirror of Linear **[MOB-1818](https://linear.app/zodl/issue/MOB-1818/stale-tip-spendable-mask-zeroes-the-whole-shielded-balance-making)** — filed here so the PR can reference it. Linear remains the source of truth.

## Problem

Right after launch the SDK masks the entire shielded spendable balance to zero until it has confirmed a fresh chain tip ([#1591] stale-tip protection). During that window the Send screen shows two controls that contradict each other:

- **Balance row — correct.** `SendFormView.swift:44` renders `WalletBalancesView`, which shows `AvailableBalanceView(balance:showIndicator:)` with `showIndicator` driven by `isProcessingZeroAvailableBalance`. The amount is replaced by a `ProgressView()` — "still working it out".
- **Max chip — wrong.** `isMaxButtonEnabled` requires `spendability != .nothing && shieldedBalance.amount > 0`, both false under the mask, so `ZashiMaxChip` simply dims. It reads as *unavailable* rather than *still being determined*, right next to a balance that is visibly loading.

Product direction: while the chain tip is stale, the Send screen should look exactly as it does when the app is syncing and still determining the spendable value.

## Fix

Drive the chip's in-flight presentation from the same condition as the balance spinner. `ZashiMaxChip` already supports `isInFlight` (spinner, and non-tappable via `.disabled(!isEnabled || isInFlight)`), so no new component, string or asset is required.

Also logs the error that the `sendMaxAmount` `catch` was discarding — every failure surfaced the same generic toast, which is what made the original report impossible to diagnose.

## Context

The SDK-side observability that proved this mechanism is zodl-inc/zodl-swift-wallet-sdk#26. The mask was caught firing on a normal cold start, hiding the wallet's entire 10.49785 shielded balance for ~1s.

pacu:
Closes #2035

## Problem

Right after launch the SDK masks the entire shielded spendable balance to zero until it has confirmed a fresh chain tip ([#1591] stale-tip protection). During that window the Send screen shows two controls that contradict each other:

| control | under the mask | reads as |
|---|---|---|
| balance row (`AvailableBalanceView`) | `showIndicator: isProcessingZeroAvailableBalance` → spinner | "still working it out" ✅ |
| Max chip (`ZashiMaxChip`) | `isMaxButtonEnabled` false → dimmed | "unavailable" ❌ |

`isMaxButtonEnabled` requires `spendability != .nothing && shieldedBalance.amount > 0` — both false while masked — so the chip greys out next to a balance that is visibly loading.

This is the app-facing half of a bug that surfaced as "my orchard change is stuck in confirmation pending and I can't send max". The funds were never stuck: the note in question was 65,544 blocks deep with every gate in `is_spendable` passing. The mask was hiding it, transiently and silently.

## Change

Per product direction — while the chain tip is stale, the Send screen should look exactly as it does when the app is syncing and still determining the spendable value.

- `SendFormStore.State.isSpendabilityBeingDetermined` mirrors the condition that drives the balance spinner (`walletBalancesState.isProcessingZeroAvailableBalance`).
- `SendFormView` passes it into the chip's `isInFlight` alongside the existing `isMaxRequestInFlight`.

**No new component, string or asset.** `ZashiMaxChip` already renders a spinner for `isInFlight` and already refuses taps while it is set (`.disabled(!isEnabled || isInFlight)`), so the interaction stays correct — the chip is not tappable while the value is unknown, exactly as before, it just stops claiming to be unavailable.

Also logs the error the `sendMaxAmount` `catch` was discarding. Every failure mode surfaced the same generic toast, which is precisely what made the original report undiagnosable.

## Testing

Built `zodl-testnet` for the iOS simulator (`ARCHS=arm64`, local FFI) — `BUILD SUCCEEDED`. The only warning on the touched file is a pre-existing `alert(store:)` deprecation, shifted one line by the edit.

The chip is gated to `ZODL_INTERNAL || SECANT_TESTNET` inside `ZashiMaxChip`, so this is internal/testnet-only surface.

## Related

- SDK observability that proved the mechanism: zodl-inc/zodl-swift-wallet-sdk#26
- Linear: MOB-1818, related to MOB-1817

Still open on MOB-1818 and deliberately not in this PR: investigating what widens the ~1s mask window (slow/failed tip refresh, server switch, foreground-return after a long background), and whether the mask should time out rather than hold indefinitely.

https://claude.ai/code/session_01Gm1ZGjUGU1CvkgvFQKhV4a


Chlup (changes requested):
Thanks for chasing this down — the diagnosis in the description is right, and wiring the chip to the same signal as the balance row is the correct direction. Two things need to change before it lands, plus a couple of smaller ones.

---

## 1. `isProcessingZeroAvailableBalance` is not the stale-tip mask — the chip can spin forever

[`SendFormStore.swift:147`](https://github.com/zodl-inc/zodl-ios/blob/4d52e43c673bee5e63a48e6657c5171ff3318b4f/secant/Sources/Features/SendForm/SendFormStore.swift#L147)

The doc comment says "the SDK masks spendable to zero until it has confirmed a fresh chain tip", but the expression it aliases encodes nothing about a chain tip or a clock:

```swift
// WalletBalancesStore.swift:43
var isProcessingZeroAvailableBalance: Bool {
    if shieldedBalance.amount == 0 && ShieldingProcessorClient.isShieldable(balance: transparentBalance, threshold: autoShieldingThreshold) {
        return false
    }
    return totalBalance.amount != shieldedBalance.amount && shieldedBalance.amount == 0
}
```

That is "shielded spendable is zero while total isn't, and transparent is below the auto-shielding threshold". Two cases where it stays true indefinitely:

- **Sub-threshold transparent-only wallet.** The live threshold is `Zatoshi(100_000)` (`ZcashSDKEnvironmentLiveKey.swift:46`). A wallet holding 50,000 zats transparent and nothing shielded: `50_000 >= 100_000` is false, so the early return doesn't fire and the flag is **permanently** true. The chip spins forever, promising a figure that will never arrive. Before this PR it dimmed — which was the truthful state. Our own `WalletBalancesTests.swift:24-26` pins exactly this shape as `true`.
- **Only-unconfirmed shielded funds.** `shieldedSpendableValue` is 0 while `shieldedTotal()` isn't, so the chip spins for the ~12 minutes until confirmation, not the ~1s the description describes.

A spinner is a promise that the value is on its way. It needs a condition that is actually bounded — either narrow it to the mask itself (needs an SDK-side signal), or pair it with a timeout after which the chip falls back to dimmed. At minimum the doc comment shouldn't assert a causal story the code doesn't encode, because the next reader will trust it.

## 2. The same masked zero still lies through "Insufficient funds" on the same screen

[`SendFormStore.swift:166`](https://github.com/zodl-inc/zodl-ios/blob/4d52e43c673bee5e63a48e6657c5171ff3318b4f/secant/Sources/Features/SendForm/SendFormStore.swift#L166)

`isInsufficientFunds` is `amount.amount > shieldedBalance.amount`, and `shieldedBalance` is synced from the masked `walletBalancesState.shieldedBalance` at `SendFormStore.swift:555`. So under the very mask this PR is about:

- `invalidZecAmountErrorText` (`:191`) returns `.sendErrorInsufficientFunds` — a **red error** under both the ZEC and currency fields;
- `isValidForm` (`:151`) is false, so `.disabled(!store.isValidForm)` (`SendFormView.swift:195`) greys out **Review**.

Repro: launch, open Send, paste an address and type an amount inside the mask window. The screen tells the user they don't have the money, on funds that are genuinely spendable — a considerably stronger false claim than a dimmed chip, one row below the chip we just fixed.

I don't think the chip fix is complete while that's still there. If you'd rather split it, fine — but then the PR shouldn't be described as making the screen self-consistent, because it still isn't.

## 3. Swap and Pay have the identical contradiction, untouched

[`SendFormView.swift:108`](https://github.com/zodl-inc/zodl-ios/blob/4d52e43c673bee5e63a48e6657c5171ff3318b4f/secant/Sources/Features/SendForm/SendFormView.swift#L108)

`SwapForm.swift:274-277` renders a `ProgressView()` beside the "Max: …" label whenever `spendability == .nothing`, while `SwapForm.swift:367-374` passes only `isInFlight: store.isMaxRequestInFlight` and `isEnabled: store.isSwapMaxButtonEnabled` — false exactly when `spendability == .nothing` (`SwapAndPayStore.swift:168`). `CrossPayForm.swift:48-56` inherits it via `isPayMaxButtonEnabled`. That is the same spinning-label-next-to-dimmed-chip defect, on two more screens. `ZashiMaxChip`'s own doc comment even points at that SwapForm spinner as the thing its in-flight state matches.

Note the two screens already spell the same idea two different ways — `isProcessingZeroAvailableBalance` here, `spendability == .nothing` there. Fixing this per call site leaves three copies to keep in sync. I'd rather see one derivation (on `WalletBalances.State`, or folded into the chip) that all three read.

## 4. A cancelled max request is logged as a failure

[`SendFormStore.swift:471`](https://github.com/zodl-inc/zodl-ios/blob/4d52e43c673bee5e63a48e6657c5171ff3318b4f/secant/Sources/Features/SendForm/SendFormStore.swift#L471)

Good addition, but it sits on the wrong side of the cancellation guard. `.onDisapear` returns `.cancel(id: state.maxCancelId)` and the effect is `.cancellable(id:)`. TCA's `Send.callAsFunction` guards on `!Task.isCancelled`, so the existing `await send(.maxAmountFailed)` is correctly suppressed on cancellation — the new log line runs before it and is not. Tap Max, then leave the screen before it resolves, and (if `proposeSendMax` propagates cancellation, which the `Task.sleep`-based `onDisapearCancelsInFlightMaxRequestAndClearsFlags` test does) you get `send-max failed to resolve an amount: CancellationError()` at error level for an ordinary dismissal — noise in exactly the channel this line exists to make readable.

```swift
} catch {
    if !(error is CancellationError) {
        LoggerProxy.error("send-max failed to resolve an amount: \(error)")
    }
    await send(.maxAmountFailed)
}
```

## 5. `shieldableTransparentBalanceSuppressesTheProgressState` passes for the wrong reason

[`SendFormMaxButtonTests.swift:355`](https://github.com/zodl-inc/zodl-ios/blob/4d52e43c673bee5e63a48e6657c5171ff3318b4f/zodlTests/SendFormTests/SendFormMaxButtonTests.swift#L355)

`isShieldable` is `balance >= threshold` (`ShieldingProcessorInterface.swift:36`). With `autoShieldingThreshold = .zero`, *every* balance is shieldable, so the early return fires regardless of `transparentBalance` — the test passes identically with `transparentBalance = .zero` and can't distinguish "transparent is shieldable" from "threshold is zero". The two sibling tests in the same commit use `Zatoshi(100_000)` and say why in their comments. Use the live threshold with transparent at or above it, so the hole you're documenting is actually exercised.

## 6. Tautological assertion

[`SendFormMaxButtonTests.swift:312`](https://github.com/zodl-inc/zodl-ios/blob/4d52e43c673bee5e63a48e6657c5171ff3318b4f/zodlTests/SendFormTests/SendFormMaxButtonTests.swift#L312)

`#expect(state.walletBalancesState.isProcessingZeroAvailableBalance)` asserts the literal body of `isSpendabilityBeingDetermined`, so it cannot fail independently of the `#expect` above it. The comment says "if they ever diverge the two controls contradict each other on screen" — with an alias no divergence is representable, and the balance row reads the same property directly (`WalletBalancesView.swift:60`). Either drop the line, or replace it with something that pins the shared condition's shape against a hand-built `WalletBalances.State`.

---

Verified, in case it comes up: the cancelled-max path does **not** fire a spurious toast (TCA guards `Send` on `Task.isCancelled`), and `SendForm.State.shieldedBalance` / `walletBalancesState.shieldedBalance` stay in lockstep — both written under `.balanceUpdated` — so the chip can't end up enabled-and-spinning.

Items 1 and 2 are the blocking ones for me; 3–6 I'm happy to take as follow-ups if you'd rather keep this PR tight, as long as they're tracked.

PR #2038, opened 2026-08-27
Support remote and pinned-local SDK modes in CI

Chlup:
CI previously failed whenever the committed Xcode project referenced the SDK as a local package (`../zodl-swift-wallet-sdk`) — the sibling checkout doesn't exist on runners. This PR makes both SDK modes first-class in CI.

## What changed

- **`.sdk-pin`** (committed, repo root) records which `zodl-swift-wallet-sdk` commit the app builds against: empty when the project references the SDK remotely, a full 40-char sha when it references the local sibling. A new **`Update SDK Pin` build phase** on the three app targets maintains the file automatically on every local build (write-on-change; `-dirty` suffix while the SDK working tree has uncommitted changes; never fails a build).
- **`Scripts/setup-sdk-companion.sh`** + composite action **`.github/actions/setup-sdk-companion`**: validates the pin against the project mode (clear, instructive failures for empty/dirty/malformed pins and mode mismatches) and fetches the SDK at the pinned commit to the exact path the pbxproj declares (read from the project, not hardcoded).
- **New job topology in both workflows** (`pull-request.yml`, `e2e-smoke.yml`):
  - `validate_sdk_pin` — ubuntu gate, fails bad pins in seconds before any macOS minutes burn;
  - `build_ffi` — builds the SDK's Rust core (`libzcashlc`) **once** per push via the SDK's own `Scripts/init-local-ffi.sh` (probes for the new `--arm-simulator` slice option, falls back to `--arm-ios`), with a product cache keyed on the FFI input tree at the pin plus a cargo cache underneath; a shared concurrency group serializes the twin jobs across the two workflows so only one ever builds cold;
  - unit-test / e2e-build jobs consume the built `LocalPackages` artifact and otherwise run unchanged.
- **Removed the vestigial `IPHONEOS_DEPLOYMENT_TARGET=16.0` override** from the e2e build (standalone commit): it was added to fix a TCA compile failure under the runner's then-default old Xcode and was superseded the same day by the select-latest-Xcode step; it also force-lowered 16.6-configured code to 16.0.
- Bats coverage for both scripts (`Scripts/test/`), README section documenting the workflow.

## Notes

- Initial pin `6006366c…` requires the companion SDK change for the sim-only slice to activate later: zodl-inc/zodl-swift-wallet-sdk#27 (CI works either way via the `--arm-ios` fallback). The pin currently lives on an unmerged SDK branch — CI emits the expected durability warning; repoint after that SDK PR lands.
- Verified locally end-to-end: 30/30 bats, full `zodl-internal` simulator build against the pinned SDK with a freshly built FFI, phase writes the pin, worktree stays clean.


LukasKorba (approved):
utACK

PR #2040, opened 2026-08-28
Scripted release branch management, and a check that releases are merged back

### LINKED ISSUE #2039: Scripted release branch management, and a check that releases are merged back
Cutting a release is currently a hand-run sequence documented in
`docs/release-automation.md`: bump the version on `main`, push, branch, push
again. Two problems follow from it.

**The release branch is cut from `main`, so nothing produces a reviewable
diff of the release.** There is no artifact anywhere that answers "what are
users actually getting relative to the last version they had" — only the
intervening development history, which is not the same question.

**Nothing checks that a released branch got merged back.** A release that is
only cherry-picked back leaves its tag unreachable from any live branch, so the
released commit stops being part of the history it shipped from.

Two smaller things surfaced alongside:

- `fastlane/lib/zodl/version.rb` recognises only `release/X.Y.Z`, so the
  preflight's branch-against-version check does not apply to whatever branch a
  release is actually built from during the cycle.
- `CLAUDE.md` says every implemented task must be recorded in `CHANGELOG.md`,
  while the changelog's own preamble scopes it to user-related modifications.
  That leaves the common cases undecided: tooling, CI, tests, refactors.

nuttycom:
Closes #2039

Adds `Scripts/prepare-release.sh`, modelled on the equivalent in
`zodl-swift-wallet-sdk`, and ports that repo family's "release merged back"
check.

## `prepare-release.sh start <remote> <version> [<revision>]`

Cuts two branches and opens the pull request between them:

- **`release/X.Y.Z`** starts out identical to the **previous release tag**.
- **`candidate/X.Y.Z`** starts at the revision being released, and gets the
  CHANGELOG promotion and the version bump on top.

Basing the release branch on the previous tag is the whole point: **the release
PR's diff is then exactly what users receive relative to the last release**,
rather than the intervening development history.

It checks the tree, remote, `gh` login and version tool; fetches tags; finds the
newest `X.Y.Z` tag reachable from the revision; refuses if either branch already
exists locally or on the remote, or if `## [Unreleased]` is empty or already
carries a heading for this version; then creates and pushes `release/X.Y.Z`,
creates `candidate/X.Y.Z`, promotes the CHANGELOG to `## [X.Y.Z] - YYYY-MM-DD`,
sets `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION`, pushes, and opens the PR.

`--issue N`, `--build N` (default 1), `--previous TAG`, `--dry-run`.

The version bump delegates to `.claude/skills/update-app-version/scripts/set_version.py`
rather than reimplementing it, so target selection stays driven by the project's
object graph. That tool needs macOS; `--dry-run` reports the requirement instead
of refusing, and rehearses the whole plan anywhere.

**This changes which branch you build from.** During a release cycle
`release/X.Y.Z` is still the *previous* release, so TestFlight and App Store
builds come from `candidate/X.Y.Z`. `docs/release-automation.md` is updated
accordingly, and `fastlane/lib/zodl/version.rb` now recognises `candidate/X.Y.Z`
so the preflight's branch-against-version check applies to the branch actually
being built — previously only `MARKETING_VERSION` was checked there.

## Release Merged Back

`.github/workflows/release-merged-back.yml` runs on pull requests into a
maintenance branch or `main` and reports any tagged `release/` branch missing
from the line it belongs to. Advisory: missing from its own maintenance line
fails the job, not having reached `main` yet only warns. The maintenance line is
derived from the tag, not the branch name.

Run against this repository's refs today it reports:

```
:grey_question: `release/3.10.2` (`3.10.2`) belongs to `maint/v3.10.x`, which no longer exists; checking main only.
:white_check_mark: `release/3.10.2` (`3.10.2`) is in `main` and downstream.
:grey_question: `release/3.10.3` has no tag at its tip; not a released branch, skipped.
```

The grey question is because our maintenance branch is `maint/3.10.x` without
the `v`. Renaming it to `maint/v3.10.x` puts it back in the chain; the check
still verifies the release reached `main` in the meantime. The workflow triggers
on `maint/**` so it runs either way.

## CHANGELOG discipline

`CLAUDE.md` said every implemented task must be recorded, while the changelog's
own preamble scopes it to user-related modifications — leaving tooling, CI, tests
and refactors undecided. Replaced with the `zodl-swift-wallet-sdk` statement of
the same rules, retargeted from a library's consumers to the people who use the
app, plus an explicit list of what gets no entry. This PR takes no changelog
entry under that rule.

## Test plan

- `bats Scripts/test/` — 44 passing, including the CHANGELOG promotion against
  fixtures and a regression test that the library does not shadow bats' `run`.
- `bundle exec rake test` — 83 passing (run here as plain minitest; Ruby is not
  installed on my machine, so please confirm under the pinned toolchain).
- `shellcheck -x` clean on all three scripts; `zizmor --persona regular` clean on
  the workflow.
- `./Scripts/prepare-release.sh start --dry-run upstream 3.11.0` against current
  `main`: correctly identifies `3.10.2` as the base and warns about the absent
  `plutil` rather than failing.
- `promote_changelog` applied to the real `CHANGELOG.md`: inserts the heading,
  leaves `## [Unreleased]` in place, moves no entries.
- The merged-back check was run against the real refs of this repo,
  `zodl-swift-wallet-sdk` and `zodl-android` before committing.

**Not tested:** a real (non-dry-run) `start`, which needs macOS. The first live
release cut is the real test — do it with `--dry-run` first.

### DIFF
--- a/.claude/skills/make-builds/SKILL.md
+++ b/.claude/skills/make-builds/SKILL.md
@@ -210,7 +210,7 @@ Example — a local SDK checkout sitting on a tagged commit, app not yet tagged:
 ```
 _iOS TestFlight Build (internal)_ — 3.10.1 (1)
 
-App: `release/3.10.1@54812f81`
+App: `candidate/3.10.1@54812f81`
 SDK: `candidate/4.1.0@cafca07a (tag: 4.1.0-rc.1)`
 ```
 
--- a/.claude/skills/make-builds/scripts/build-refs.py
+++ b/.claude/skills/make-builds/scripts/build-refs.py
@@ -12,7 +12,7 @@ live) or as a pinned remote package. The two are reported in the same shape:
     <branch-or-version>@<sha8>[ (tag: X)][ (dirty)]
 
 Usage:
-    build-refs.py --ref release/3.10.1 --sha 54812f81 [--repo /path/to/repo]
+    build-refs.py --ref candidate/3.10.1 --sha 54812f81 [--repo /path/to/repo]
 
 Prints one or two lines to stdout, ready to paste into the reply. The SDK line
 is omitted entirely when no SDK reference can be resolved; the reason goes to
--- a/.claude/skills/update-whatsnew/SKILL.md
+++ b/.claude/skills/update-whatsnew/SKILL.md
@@ -1,8 +1,8 @@
 ---
 name: update-whatsnew
-description: Prepare a ZODL release changelog. Reads the version from the zodl-production target, writes the release entry into every whatsNew*.json file (one per language) — prepending it, or replacing the existing entry if that version is already there — and prints the App Store changelog text for each language. Manual only — invoke with /update-whatsnew and paste the multi-language changelog.
+description: Prepare a ZODL release changelog. Takes the release version from the invocation, or reads it from the zodl-production target, writes the release entry into every whatsNew*.json file (one per language) — prepending it, or replacing the existing entry if that version is already there — and prints the App Store changelog text for each language. Manual only — invoke with /update-whatsnew and paste the multi-language changelog.
 disable-model-invocation: true
-argument-hint: <paste the multi-language changelog (language blocks with Added/Changed/Fixed sections)>
+argument-hint: [X.Y.Z] <paste the multi-language changelog (language blocks with Added/Changed/Fixed sections)>
 ---
 
 # Update What's New + App Store changelog
@@ -35,7 +35,9 @@ user's message:
 
 ARGUMENTS: $ARGUMENTS
 
-It is one block per language. Each block starts with a language name
+The argument may open with the release version (`X.Y.Z`) before the first
+language block; step 1 says how it is used. The changelog itself is one block
+per language. Each block starts with a language name
 (`English`, `Español`, …) and contains sections — a short header line ending in
 `:` (`Added:`, `Changed:`, `Fixed:`, `Removed:`, and their localized forms like
 `Añadido:`, `Cambiado:`, `Corregido:`) followed by one or more items.
@@ -47,10 +49,15 @@ changelog isn't in the argument or the message, ask the user to paste it.
 
 ## Workflow
 
-### 1. Read the release version
+### 1. Determine the release version
 
-Read `MARKETING_VERSION` from the **zodl-production** target (this is what ships
-to the App Store). It takes ~20s — that's normal:
+A version named in the invocation wins: a leading `X.Y.Z` before the changelog
+(`/update-whatsnew 3.11.0 …`), or one stated in the user's message, is the
+release version exactly as given.
+
+Only when no version was named, read `MARKETING_VERSION` from the
+**zodl-production** target (this is what ships to the App Store). It takes
+~20s — that's normal:
 
 ```bash
 xcodebuild -project secant.xcodeproj -target zodl-production \
@@ -64,6 +71,14 @@ version, so this is reliable in practice, just not target-specific):
 ```bash
 grep -m1 'MARKETING_VERSION' secant.xcodeproj/project.pbxproj | sed 's/.*= *//; s/;.*//'
 ```
+
+A checkout-derived version needs a cross-check: during a release cycle only
+`candidate/X.Y.Z` carries the new version, so every other checkout — `main`
+included — still reads the **previous** release. On a `candidate/X.Y.Z` or
+`release/X.Y.Z` branch (`git branch --show-current`), the version read must
+equal the version in the branch name — stop and report a mismatch. On any
+other checkout the version cannot be confirmed here; step 4's dry-run is the
+tripwire.
 
 ### 2. Determine date and timestamp — once, shared by all languages
 
@@ -122,10 +137,17 @@ Read what each dry-run reports:
 
 - **`would add`** — the version is new; it gets prepended.
 - **`would replace`** — the version already has an entry and it will be
-  overwritten in place. This is normal for an amended changelog; just carry on.
-  The entry keeps its original `date`/`timestamp` so an amendment doesn't look
-  like a re-release. **Mention the replacement to the user** when you report, and
-  if they want the date bumped to today, re-run that file with `--refresh-date`.
+  overwritten in place. When the version was named in the invocation, or was
+  read on the release's own candidate branch, this is normal for an amended
+  changelog; just carry on. The entry keeps its original `date`/`timestamp` so
+  an amendment doesn't look like a re-release. **Mention the replacement to the
+  user** when you report, and if they want the date bumped to today, re-run
+  that file with `--refresh-date`. But when the version was derived from a
+  checkout that is **not** the release's candidate branch, an existing entry
+  is the stale-checkout signature — a freshly cut release never already has
+  one, so this is almost certainly the previous release's version about to be
+  clobbered with the new copy. **Stop and ask the user to confirm the
+  version** before writing anything.
 - **`Unchanged`** — the on-disk entry already matches the payload byte for byte;
   nothing will be written. Expected when re-running with an unedited changelog.
 - **A language with no file yet** is a new language. Do **not** guess the
--- a/.claude/skills/update-whatsnew/evals/evals.json
+++ b/.claude/skills/update-whatsnew/evals/evals.json
@@ -9,12 +9,12 @@
       "files": [],
       "assertions": [
         "Both files updated; the new entry is at index 0 of releases (newest-first)",
-        "version equals zodl-production MARKETING_VERSION; date is DD-MM-YYYY; timestamp is an integer epoch",
+        "version equals zodl-production MARKETING_VERSION (no version was named in the invocation); date is DD-MM-YYYY; timestamp is an integer epoch",
         "Section titles are kept verbatim per language (Added:/Changed:/Fixed: and Añadido:/Cambiado:/Corregido:), not translated",
         "The 'Fixed:'/'Corregido:' section has two bulletpoints; the others have one",
         "Existing entries (3.6.0 and older) are byte-for-byte unchanged; output is valid JSON",
         "App Store text: one block per language, section title line then '* ' bullets, blank line between sections",
-        "If an entry for the version already exists, nothing is written and the user is warned (stop-and-warn)"
+        "If an entry for the version already exists, nothing is written and the user is asked to confirm the version first — with no version named and the checkout not on the release's candidate branch, an existing entry is the stale-checkout signature; replace-in-place without asking is right only for an explicit version or one read on its own candidate branch"
       ]
     }
   ]
--- a/.github/scripts/check-release-merged-back.sh
+++ b/.github/scripts/check-release-merged-back.sh
@@ -1,0 +1,136 @@
+#!/usr/bin/env bash
+#
+# Check that every tagged release/ branch has been merged back into its
+# maintenance line, and forward down to main.
+#
+# Usage: check-release-merged-back.sh [pr-base] [pr-ref]
+#
+#   pr-base   branch a pending change targets; that branch is then evaluated
+#             as it would be with the change applied. Omit to check the
+#             branches as they stand.
+#   pr-ref    the pending change; defaults to HEAD (refs/pull/N/merge in CI).
+#
+# A release/ branch counts as released only when a tag points at its tip, so
+# in-flight release branches are skipped. The maintenance line comes from the
+# TAG, not the branch name: in the Swift SDK, release/3.0.0 is tagged 2.8.0-rc.1
+# and belongs to maint/v2.8.x.
+#
+# Branches resolve under refs/remotes/$REMOTE, which defaults to origin -- what
+# CI checks out. Set REMOTE to run this against a local clone whose canonical
+# remote has another name, and run `git fetch` first.
+#
+# Exits 1 when a tagged release is missing from a branch it should be in.
+
+set -euo pipefail
+
+PR_BASE="${1:-}"
+PR_REF="${2:-HEAD}"
+REMOTE="${REMOTE:-origin}"
+
+say() {
+  printf '%s\n' "$*"
+  if [ -n "${GITHUB_STEP_SUMMARY:-}" ]; then printf '%s\n' "$*" >>"$GITHUB_STEP_SUMMARY"; fi
+}
+
+# Maintenance lines in version order, then main. v:refname sorts v2.10.x after
+# v2.9.x.
+CHAIN=()
+while IFS= read -r ref; do
+  CHAIN+=("$ref")
+done < <(
+  git for-each-ref --sort=v:refname --format='%(refname:lstrip=3)' \
+    "refs/remotes/${REMOTE}/maint/v*"
+)
+CHAIN+=('main')
+
+# Resolve a chain branch, substituting the pending change when it targets it.
+resolve() {
+  if [ -n "$PR_BASE" ] && [ "$1" = "$PR_BASE" ]; then
+    git rev-parse "$PR_REF"
+  else
+    printf '%s/%s\n' "$REMOTE" "$1"
+  fi
+}
+
+say '## Release branches merged back'
+say ''
+
+failed=0
+checked=0
+
+while IFS= read -r rel; do
+  [ -n "$rel" ] || continue
+  ref="${REMOTE}/${rel}"
+
+  tags="$(git tag --points-at "$ref" | tr '\n' ' ')"
+  tags="${tags% }"
+  if [ -z "$tags" ]; then
+    say ":grey_question: \`${rel}\` has no tag at its tip; not a released branch, skipped."
+    continue
+  fi
+
+  # [v]MAJOR.MINOR.PATCH[-anything] -> maint/vMAJOR.MINOR.x. The optional
+  # trailing part covers both a pre-release suffix (2.8.0-rc.1) and the build
+  # number some repos append (3.9.3-2393).
+  tag="${tags%% *}"
+  ver="${tag#v}"
+  major="${ver%%.*}"
+  rest="${ver#*.}"
+  minor="${rest%%.*}"
+  maint="maint/v${major}.${minor}.x"
+
+  # Its own line is the obligation. Anything downstream of it is merge-forward
+  # lag, which the merge-forward jobs already track, so it warns rather than
+  # fails; otherwise the merge-back PR itself would report red for main.
+  # A retired line is simply absent from the chain.
+  own=''
+  downstream=()
+  seen=0
+  for b in "${CHAIN[@]}"; do
+    if [ "$b" = "$maint" ]; then seen=1; own="$b"; continue; fi
+    if [ "$seen" -eq 1 ]; then downstream+=("$b"); fi
+  done
+  if [ "$seen" -eq 0 ]; then
+    own='main'
+    downstream=()
+    say ":grey_question: \`${rel}\` (\`${tag}\`) belongs to \`${maint}\`, which no longer exists; checking main only."
+  fi
+
+  checked=$((checked + 1))
+
+  lagging=()
+  for b in ${downstream[@]+"${downstream[@]}"}; do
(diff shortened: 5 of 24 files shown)


Chlup (changes requested):
Careful, well-documented PR — the reasoning comments in `release-lib.sh` and the choice to base the release branch on the previous tag are both right, and the bats coverage is a real addition. Requesting changes on a handful of items, most of which are small; the grouping below is by cost-to-fix, not just by severity.

Findings were verified by reproducing them in throwaway clones, so each one below has an observed failure rather than a reading of the code. Where I reproduced something, I say what I saw.

---

## Must fix before merge

These either damage the canonical repo, or produce a release that is wrong in a way nobody sees.

### 1. `Scripts/test/prepare_release_args.bats:70` — the test suite cuts and pushes a real release

```
run "$PREPARE" start upstream v3.11.0
```

No `--dry-run`, and the only assertion is about absent text — `$status` is never checked, so the test reports `ok` whatever the script does. `v3.11.0` strips to `3.11.0`, passes `RELEASE_TAG_RE`, and `--build` defaults to 1, so parsing completes and the script runs to the end.

Reproduced: in a clone whose only added state was a remote named `upstream`, `bats Scripts/test/` printed all 44 `ok` lines while it pushed `release/3.11.0` and `candidate/3.11.0` to that remote, switched HEAD to `candidate/3.11.0`, rewrote `CHANGELOG.md`, edited 9 `MARKETING_VERSION` lines in `project.pbxproj`, and made two commits. Only `gh pr create` failed, and only because the sandbox remote was not a GitHub slug.

This is armed by the docs in this same PR: `upstream` is the remote name in all five places `<remote>` is illustrated, and `docs/release-automation.md:333` prescribes exactly `bats Scripts/test/`. The file header at line 3 and `docs/release-automation.md:325` both state the suite needs no network, remote, or token.

**Fix:** add `--dry-run` and assert `$status`.

### 2. `Scripts/prepare-release.sh:121` — `--dry-run` after the positionals is silently discarded

The option loop `break`s at the first non-option word (line 117) and there is no `$# -gt 3` check; `$4` onward is never read.

Reproduced: `start upstream 3.11.0 HEAD --dry-run` performed a full live cut — both branches pushed, working tree switched to the candidate, both commits made — with nothing in the output mentioning the ignored argument.

The three-argument form is not much better: `start upstream 3.11.0 --dry-run` assigns `revision="--dry-run"` and dies with a bare `fatal: Needed a single revision`. `start upstream 3.11.0 --previous 3.10.3` fails the same way and silently drops the tag — an ordering `usage_start:60-62` invites by describing `--previous` in terms of `<revision>`.

**Fix:** reject extra positionals (and ideally keep parsing options after them).

### 3. `Scripts/prepare-release.sh:166` — `--previous` is never checked for ancestry

Lines 164-168 run only `git rev-parse -q --verify "refs/tags/${prev_tag}"`. The auto-detected path is safe — `git tag --merged` guarantees ancestry — so this is the only hole, and `version_le` does not close it.

Reproduced using the exact case `docs/release-automation.md:234` recommends `--previous` for — cut `3.10.4` from `origin/maint/3.10.x` with `--previous 3.10.3`, where `3.10.3` is not an ancestor of that branch:

- the three-dot PR diff **hid** 802 insertions across 9 files present only on the base side, including `project.pbxproj`, `whatsNew.json`, `whatsNew_es.json` and `MigrationManagerLiveKey.swift`;
- merging the PR auto-merged cleanly, and `git diff release/3.10.4 candidate/3.10.4` then showed 42 insertions / 802 deletions;
- merged `release/3.10.4` records `MARKETING_VERSION = 3.10.3` / build 2, while `candidate/3.10.4` — the branch actually built and uploaded — records `3.10.2` / 1.

So the tag signed on the release branch identifies source that was never built, and none of the discrepancy is visible in the PR diff.

**Fix:** `git merge-base --is-ancestor "refs/tags/$prev_tag" "$rev_sha"` after the existence check.

### 4. `Scripts/prepare-release.sh:171` — the base is the newest *reachable* tag, not the newest released

`release_tags_merged_into` returns the newest release tag reachable from `<revision>`, and nothing compares that against the newest tag that exists.

This is live on today's refs. Tag `3.10.3` exists but is an ancestor of neither `origin/main` nor `origin/maint/3.10.x` — the only ref containing it is `refs/remotes/origin/release/3.10.3`. So `prepare-release.sh start origin 3.11.0` from `main` right now bases `release/3.11.0` on `3.10.2` and opens a PR whose body asserts "its diff is exactly what users receive relative to the last release", while the release it produces drops all 13 commits and 9 files that 3.10.3 shipped.

Nothing catches it: `version_le 3.11.0 3.10.2` passes, both CHANGELOG guards pass, and the merged-back workflow never runs on this PR because its base is `release/X.Y.Z`, not `main` or `maint/**`.

**Fix:** compare the detected tag against the newest existing release tag and refuse (or require `--previous`) when they differ.

### 5. `docs/release-automation.md:162-164` — the step-5 snippet can publish the tag on the wrong branch

```bash
git switch release/3.8.0 && git pull --ff-only upstream release/3.8.0
git tag -s 3.8.0 -m "Release 3.8.0"
git push upstream refs/tags/3.8.0        # this fires linear-release.yml
```

`prepare-release.sh:296-297` deliberately leaves the operator on `candidate/X.Y.Z`, and switching to the release branch changes both `CHANGELOG.md` and `project.pbxproj`, so any stray local edit makes `git switch` refuse. The `&&` correctly stops the pull — and lines 2 and 3 then run anyway, signing and pushing `3.8.0` onto `candidate/3.8.0`'s tip.

That is not recoverable by re-tagging: the tag is on the remote, `linear-release.yml` has already fired and cut a Linear release from the wrong range, and `check-release-merged-back.sh:65` then finds no tag at `origin/release/3.8.0`'s tip and reports the release as `not a released branch, skipped` — so it drops out of the very check this PR adds.

**Fix:** chain all four commands with `&&`, or split the tag into its own verified step.

---

## One-line fixes — no reason to defer

### 6. `Scripts/prepare-release.sh:261` — open the release PR as a draft

`.github/workflows/e2e-smoke.yml` has no `branches:` filter on its `pull_request` trigger, and neither gate stops the new long-lived candidate→release PR: the draft gate passes because `gh pr create` is called without `--draft`, and `paths-ignore` is evaluated against the PR's full diff, which for a release PR is the whole release delta (measured `3.9.5...3.10.2` = 30 non-ignored files) — so even a docs-only fix pushed to the candidate fires the suite. The concurrency group is per-PR, so it cannot dedupe against the fix PR's own run.

The workflow states its own cost at lines 35 and 366-372: up to 75-minute runs, ~$2.30 USDC + ~$0.05 ETH gas per fire. Under the old flow no such PR existed (#2030 confirms fixes went in as PRs into `release/3.10.2`, one run each); now every fix costs two.

Adding `--draft` fires it exactly once, since the trigger already lists `ready_for_review`.

### 7. `Scripts/prepare-release.sh:204,209` — the CHANGELOG guards read the wrong working tree

Both preconditions read `CHANGELOG.md` from whatever is checked out; line 224 then replaces the tree with `git switch -c "$candidate_branch" "$rev_sha"`, and line 231 edits the new file.

Reproduced both shapes with an explicit `<revision>`:

- a revision already carrying a `## [3.11.0]` heading passed `changelog_has_version` — whose own comment says "a second one for the same version must not be added" — producing a candidate with two `## [3.11.0]` headings;
- a revision with an empty Unreleased section passed the other guard while the script printed "the Unreleased section has entries to promote", shipping `## [3.11.0] - <date>` with nothing under it.

In both runs `release/3.11.0` was already pushed before the damage. The default `<revision>` = HEAD case is safe (`require_clean_tree` covers it), so the trigger is an explicit revision — a documented positional, and the natural `start upstream 3.11.0 origin/main` from a stale local branch hits it.

**Fix:** move both checks after the `git switch`.

### 8. `CLAUDE.md:96-98` — What's New does not come from `CHANGELOG.md`

The added line says `CHANGELOG.md` "is also where the App Store 'What's New' copy comes from". It isn't: that comes from `secant/Resources/WhatsNew/whatsNew*.json` via `Zodl::WhatsNew` (`fastlane/lib/zodl/whats_new.rb:41-86`), maintained by the `update-whatsnew` skill — and `docs/release-automation.md:188` and `:311`, edited in this same PR, say so. Grepping the tree for `CHANGELOG` finds no consumer outside the release scripts' own promotion logic.

The practical cost: the new "Cutting a release" narrative and the rewritten five-step flow enumerate everything a release needs and never mention `whatsNew*.json`, while `CLAUDE.md` now asserts the CHANGELOG covers it. The gap surfaces at `--submit-review`, which hard-fails on a missing What's New entry — after the binary is uploaded.

**Fix:** drop the sentence, and add the `/update-whatsnew` step to the five-step flow.

### 9. `Scripts/test/release_lib.bats:239` — the assertion can never fail

```
[[ "$output" != "hello" ]]
```

Intended as "the command was not executed", but the describe line is always present, so `$output` can never equal `"hello"` exactly. Break `run_cmd` so it prints the describe line *and* falls through to `"$@"` — the classic missing-`else` regression — and `$output` becomes `"  would run: echo hello\nhello"`; line 238's substring match still passes and line 239 is still true. The suite stays green while `--dry-run` has started pushing branches for real.

Given that two of the must-fix items above are `--dry-run` failing to be a rehearsal, this is the one assertion that would have caught the primitive breaking. `[[ "$output" != *"hello"* ]]` is sound.

---

## Worth deciding deliberately

### 10. `Scripts/prepare-release.sh:221` — `release/X.Y.Z` is pushed before any work that can fail

After the push, five steps can still abort: the CHANGELOG promotion (die at 232), the version bump (die at 247), either `git commit` (237 / 251, which `set -e` kills on "nothing to commit"), and the candidate push (254). Only `gh pr create` gets a recovery path (265-280). For the others the script dies with the release branch orphaned on the remote, and its own guard at 196-200 then refuses the retry.

The die at 247-249 compounds it: its message says only "`project.pbxproj` is unchanged", and its comment justifies leaving the tree dirty so a re-run "starts from a clean tree rather than dying at `require_clean_tree`" — but by then HEAD has already moved to the candidate and a CHANGELOG commit exists, neither of which it mentions. There is no `trap` to unwind any of it.

Doing all local work first and finishing with `git push --atomic "$remote" "$release_branch" "$candidate_branch"` collapses this to one irreversible step and dissolves item 11 below along with most of the "delete it there and re-run" recovery text. It is a restructure rather than a one-liner, but it is cheaper now than after the first live cut.

### 11. `Scripts/prepare-release.sh:251` — a no-op version bump aborts the script silently

`set_version.py`'s `cmd_set` has no already-at-target short-circuit; it rewrites unconditionally and returns 0. Reproduced on a pbxproj already at 3.11.0 / build 1: the tool printed `MARKETING_VERSION: 3.11.0 -> 3.11.0` and `Updated …`, md5 unchanged, `git status` empty — so the `elif !` guard at line 242 never fires. `git add` stages nothing and line 251 exits 1 with only `nothing to commit, working tree clean`, no `error:` line, leaving `release/3.11.0` on the remote and HEAD on the candidate.

Reachable in practice: `CURRENT_PROJECT_VERSION` is permanently 1 across all 9 configurations here and `--build` defaults to 1, so only `MARKETING_VERSION` has to match — e.g. re-cutting from a prior candidate's sha, which the script's own "delete and re-run" text at 188-191 invites.

Neutralised entirely by item 10.

### 12. The merged-back workflow is inert where it matters and noisy where it isn't

Two separate defects, which together are a question about whether this workflow should ship in this PR:

**`check-release-merged-back.sh:42`** — the chain is built only from `refs/remotes/$REMOTE/maint/v*`, but this repo's branch is `maint/3.10.x`. `git for-each-ref … 'refs/remotes/origin/maint/v*'` returns nothing, so `CHAIN=('main')` and `resolve()` is only ever called with `main`. Running the script both ways gives byte-identical output:

```
$ .github/scripts/check-release-merged-back.sh
$ .github/scripts/check-release-merged-back.sh maint/3.10.x origin/maint/3.10.x
… :x: `release/3.10.3` (`3.10.3`) is **not** merged back into `main`.   (exit 1 both times)
```

So on a PR into `maint/3.10.x` the pending change is never substituted, and because `maint/v3.10.x` is absent from the chain the `seen -eq 0` branch rewrites `own` to `main` — meaning the merge-back PR the check exists to encourage is judged against `main` and reported red. The PR description proposes renaming the branch by hand; the workflow comment at lines 22-25 documents `maint/**` as deliberate breadth while the script it invokes cannot see those branches.

**`prepare-release.sh:220`** — cutting `release/X.Y.Z` from the previous tag leaves that tag pointing at the new branch's tip, and `git tag --points-at` peels annotated tags, so it sees it. Immediately after a cut: `git tag --points-at upstream/release/3.11.0` → `3.10.2`. This invalidates the script's own premise at its lines 13-14 ("in-flight release branches are skipped") — the skip can never apply to branches this PR's tooling produces. Reproduced against a mirror of the real refs, it emits `:x: release/3.11.0 (3.10.3) is not merged back into maint/v3.10.x` — a hard failure row for a branch created seconds earlier that has released nothing — and increments `checked`, inflating the "All N tagged release branches are merged back" count. It persists for the whole release cycle.

`continue-on-error: true` means neither blocks anything. But a check that is simultaneously silent where it matters and noisy where it doesn't will get ignored, which is the failure mode you can't recover from later. Either fix both, or split the workflow into its own PR.

---

## Follow-ups

Real, but none of them need to hold this up.

- **`Scripts/lib/release-lib.sh:91`** — under `pipefail`, `grep` exiting 1 on no match makes the pipeline fail, so `prev_tag="$(release_tags_merged_into "$rev_sha" | tail -1)"` trips `set -e`. Reproduced with all X.Y.Z tags removed: output stops at `==> Determining the release base`, exit 1, **zero bytes on stderr**. The `die "no release tags are reachable…"` at `prepare-release.sh:172-175` is therefore unreachable — when the pipeline succeeds, grep matched, so `tail -1` is never empty. Triggers on a first release or a shallow CI checkout.
- **`docs/release-automation.md:223` and `:283`** — `--issue <N>` is documented as "the PR body closes it", and `prepare-release.sh:259` echoes "body would close issue #N". GitHub ignores closing keywords entirely unless the PR merges into the **default branch**; the base here is `release/X.Y.Z`, so no link is even created. (`usage_start:54-55` is accurate as written.)
- **`.github/scripts/check-release-merged-back.sh:76`** — the tip tag is not filtered by a release-tag shape, and `${tags%% *}` takes the alphabetically-first of several tags. Reproduced: `beta4-rc2` at a release tip yields `maint/vbeta4-rc2.beta4-rc2.x`; a `0.0.1-99-mainnet` build tag alongside a real one yields `maint/v0.0.x` and reports the branch green against the wrong line. This repo multi-tags 11 commits with `0.0.1-NN-mainnet`/`-testnet` pairs, so it is one stray tag away. `RELEASE_TAG_RE` already exists in `release-lib.sh` for exactly this.
- **`Scripts/lib/release-lib.sh:63`** — `echo "  would run: $*"` loses per-argument quoting. Real output: `would run: git commit -m Promote CHANGELOG for 3.11.0`, and `would run: git commit -m Bump version to 3.11.0 (1)` — the second is a bash syntax error if pasted. Neither is the command that runs.
- **`Scripts/prepare-release.sh:145`** — `GH_REPO` comes from a pure sed transform with no shape check and is first *used* at `gh pr create`, after both pushes. `repo_slug_from_url "../upstream.git"` → `upstream`; a URL with a port → `2222/zodl-inc/zodl-ios`. A `grep -qE '^[^/]+/[^/]+$'` at line 145 moves it into preflight.
- **`Scripts/prepare-release.sh:19`** — `cd "$(git rev-parse --show-toplevel)"` then `. "Scripts/lib/release-lib.sh"`. `cd ""` returns 0, so `set -e` does not catch a failed `rev-parse` outside a repo; and invoking the script by absolute path from another checkout sources *that* checkout's library and operates on its CHANGELOG, remote and tags (reproduced). The library's own header at line 5 documents the script-relative idiom this script does not use.
- **`Scripts/lib/release-lib.sh:124`** — `promote_changelog` does `mktemp` (0600) + `mv`, leaving `CHANGELOG.md` at mode 0600 in the working tree for the rest of the cut. Git records only the exec bit, so it never leaves the machine — a nit, but a one-line fix.
- **`.github/workflows/release-merged-back.yml:57`** — the explicit `git fetch --tags origin '+refs/heads/*:refs/remotes/origin/*'` is redundant. At the pinned SHA, `getRefSpecForAllHistory()` returns exactly `['+refs/heads/*:refs/remotes/origin/*', '+refs/tags/*:refs/tags/*']` when `fetchDepth <= 0`. The comment above it documents a requirement checkout has already met, so the next reader will preserve the line.
- **`.github/workflows/release-merged-back.yml:38`** — `continue-on-error: true` is at **job** level, so it also swallows the check never having run. A failed checkout or a rate-limited fetch produces a green run with an empty step summary and nobody learns the check stopped working. Moving it onto the `Check release branches` step keeps the advisory semantics for findings only.
- **Cleanup**, if you're in these files anyway: `project_marketing_version` (`release-lib.sh:160`) has no caller outside its own bats test, and is the third `MARKETING_VERSION` reader after `marketing_version_at` and `set_version.py show`. `start_pr_body` is generated twice on the failure path (`prepare-release.sh:260` and `:270`) and the `mktemp` file is never cleaned; writing it once up front and passing `--body-file "$f"` collapses both. The `start` option table exists in three copies (`usage_start:53-63`, `docs/release-automation.md:221-227` and `:282-288`) which have already drifted — the two doc copies say "the PR body closes it", and only one lists `-h, --help`. The `maint/vX.Y.x` convention is encoded five ways across three files with no shared helper.
- **`docs/release-automation.md`**: the `release/` → `candidate/` rewrite was applied to steps 2-4 but not to the `--submit-review` example (`:177`), the `--dry-run` example (`:254`), the prose at `:258`, or the troubleshooting rows at `:298` and `:312` — the last of which answers the error that building from `release/X.Y.Z` produces by telling you to run `bump` on it, which corrupts the PR base. Line 67 still names only `release_args.bats`. `fastlane/README.md`, which `:9` points readers to first, still documents the whole deleted flow and never mentions `prepare-release.sh`.
- **`CLAUDE.md`**: `promote_changelog` moves every `### Added`/`### Changed`/… subsection under the new version heading, leaving `## [Unreleased]` completely empty — but the rewrite dropped the old "create that section if it isn't there", so the instruction to file under a subsection is unsatisfiable for the first entry of each cycle. Separately, nothing says where an entry goes for a fix that lands on `candidate/X.Y.Z` after the promotion; the stated default files it under the *next* release. And `CHANGELOG.md`'s own preamble still states the older, softer policy this section supersedes.

---

## Checked and cleared

Recording these so they don't get raised again:

- **The `RELEASE_BRANCH` widening is sound.** `from_branch` has one consumer (`Fastfile:79` → `Preflight::Context#branch_version` → the equality check at `preflight.rb:35`). Under the narrow regex `candidate/X.Y.Z` returned nil and the check was *skipped*; widening it turns the check on for the branch every release build now comes from. Nothing that previously passed now fails except a genuine mismatch.
- **Commit messages.** `CLAUDE.md:80` prescribes `[#<issue_number>]`, and none of the five commits here carry one — but 0 of the last 100 non-merge commits on `main` use that form either, and the closest precedent (the SDK-pin scripts + bats + CI PR) landed 13 commits with no identifier. The rule is stale, not this PR.
- **No CHANGELOG entry is correct** under the new rules this PR installs.
- **bats hygiene is clean** — each `@test` runs in its own process, `setup()` runs per test, `$BATS_TEST_TMPDIR` is per test, and nothing leaks between tests.
- **bash 3.2 safety** — `${downstream[@]+"${downstream[@]}"}`, `${#lagging[@]}` on an empty array, `sort -V` with the `~` pre-release mapping, and `[ -n "$issue" ] && echo` under `set -e` all behave correctly on macOS 3.2.57.
- **`github.base_ref` reaches the script through `env:`**, the injection-safe pattern, and its format matches the `%(refname:lstrip=3)` chain entries.
- **`set_version.py --project` default matches `PBXPROJ`**, so `git add` and `git checkout --` target the right file.

Chlup (approved):

Issue #32, opened 2021-09-23
Onboarding Designs (initial implementation)


geffenz:

pacu:
@geffenz  I think this is done actually, isn't it?

geffenz:
Yep, approved!

PR #2042, opened 2026-08-29
[MOB-1831] Preserve spendable balance during automatic server switching

zcash-harry:
## Issue

MOB-1831

## Summary

- Run automatic server benchmarking and the durable local balance read at the same time.
- Apply an automatic winner only after the local balance read completes.
- Keep the last concrete local balance visible while the synchronizer is replaced.
- Replace the retained balance when the new synchronizer publishes state.
- Defer automatic switches during payments, voting, migration, server setup, and background work.
- Keep manual server changes immediate.

## Root cause

An automatic server winner can replace the synchronizer during startup. The replacement starts with empty in-memory state. Send then loses its known spendable balance until wallet state reloads.

## Result

On cold start, the server race does not block the durable database read. If the race finishes first, its winner waits for a completed local balance read. An empty balance map is a valid completed read. The result does not depend on the selected account or a nonzero balance.

The automatic winner can apply while the app is visible. The UI keeps the last concrete local balance during the short rebuild.

## SDK dependency

- Pin SDK PR #32 at `a3d35a37517f5926a56a8a147e1c40473f9c5241`.
- Use the repository companion-SDK workflow so CI builds the matching Rust FFI.
- Adapt the existing anchor-error check to the SDK `RedactedRustError.message` API.

SDK PR: https://github.com/zodl-inc/zodl-swift-wallet-sdk/pull/32

## Verification

- SDK offline tests: 979 passed, 0 failed.
- iOS focused tests: 20 passed, 0 failed.
- SDK lint passed.
- SDK protobuf check passed.
- SDK pin validation passed.
- Xcode project, package JSON, and Git diff checks passed.
- The local Bats suite was not run because `bats` is not installed.

## Manual QA

1. Enable automatic server selection.
2. Terminate the app.
3. Cold-start and immediately open Send.
4. Confirm that the saved spendable balance appears without a loading gap.
5. Keep the app visible and let a qualified winner apply.
6. Confirm that Send keeps its balance during synchronizer replacement.
7. Confirm that manual server changes remain immediate.

# Author

- [x] Automated tests added
- [x] Changelog updated
- [x] Compatible SDK pin added
- [x] Branch rebased and squashed
- [ ] Manual device test completed

# Reviewer

- [ ] Checklist review
- [ ] Automated tests
- [ ] Manual test


Chlup (changes requested):
Requesting changes. The blocking item is that the test target does not compile, so nothing in this PR is currently verified by CI. Beyond that, the second commit appears to revert the first, which leaves the symptom the branch is named for unfixed.

Findings are ordered by severity. Line numbers refer to the head of this branch.

---

## Blocking

### 1. `zodlTests` does not compile — SDK import is missing `@testable`

`zodlTests/UtilTests/AutoServerSelectionClientTests.swift:4` imports the SDK without `@testable`:

```swift
@preconcurrency import ZcashLightClientKit
```

The new test then constructs two types whose initializers are **internal** to `ZcashLightClientKit`:

- `AccountUUID.init(id: [UInt8])` — declared without `public` (SDK: `Sources/ZcashLightClientKit/Account/Account.swift:41`)
- `Account` — a `public struct` with no explicit init, so its memberwise init is internal (SDK: `Sources/ZcashLightClientKit/Model/WalletTypes.swift:13`)

`@testable` is a per-file import attribute, so a sibling file's does not help. This is `error: 'init(id:)' is inaccessible due to 'internal' protection level`, which fails the whole target — 191 test files, ~1600 tests never run.

All 49 other test files in the repo that build an `AccountUUID` use `@testable @preconcurrency import ZcashLightClientKit`. This file is the only exception.

Worth noting why this got through: the PR description lists `xcrun swiftc -frontend -parse` as the verification step. That is syntax-only and structurally cannot catch an access-control error. Per CLAUDE.md the test target runs via the `zodl-internal` scheme; if the local SDK checkout is blocking that, it is worth resolving before merge rather than around it.

**Fix:** add `@testable` to line 4.

---

## Correctness

### 2. The second commit reverts the first — the Send-balance fix is gone

This may be an unintended rebase or amend artifact, so flagging it rather than assuming intent:

```
git show ceb63ccc --stat   # WalletBalancesStore.swift +20, WalletBalancesTests.swift +52
git show 1bdf2f9a --stat   # WalletBalancesStore.swift -20, WalletBalancesTests.swift -52
git diff d76c9fca..HEAD --stat  # both files absent — net zero
```

`ceb63ccc` ("Fix Send balance during automatic server switch") made `.updateBalances` publish the in-memory snapshot before awaiting the database:

```swift
let cachedBalance = sdkSynchronizer.latestState().accountsBalances[account.id]
return .run { send in
    if let cachedBalance { await send(.balanceUpdated(cachedBalance)) }
    ...
}
```

`1bdf2f9a` deleted that and restored the original await-database-first ordering. The changelog line was rewritten to match — from *"Opening Send immediately after a cold start no longer leaves the spendable balance loading…"* to the implementation description now on the PR.

The replacement (a Root navigation gate) cannot cover the manual path: `ServerSetupStore.applyServerSwitch:381-386` calls `switchToEndpoint` under `switchWaiting` with no hydration term at all. Combined with finding 3, it does not cover automatic switches after the first either.

If dropping the balance fix was deliberate, could the reasoning go in the PR description? The two commits fix different bugs at different layers, and the first is the one that matches the branch name.

### 3. The hydration latch never re-arms, so only the first switch per process is gated

`hydratedBalanceAccountID` has exactly three writers: set at `RootInitialization.swift:276` and `:1310`, cleared only at `:1025` inside `.initialization(.initializeSDK)`.

There is no `else` branch clearing it when an emission's `accountsBalances` no longer contains the key — and the SDK does emit empty maps (the SDK's `SlipstreamSynchronizer.prepare` sends `summary?.accountBalances ?? [:]`, and `SynchronizerState.zero` is `accountsBalances: [:]`).

A warm foreground routes `willEnterForeground` → `.retryStart` (`RootInitialization.swift:155-157`), never `.initializeSDK`, and `.retryStart` re-sends `.refreshAutomaticServer` at `:858`. So a cold start latches once, the app is backgrounded and foregrounded for days, and every later benchmark winner applies with the hydration term permanently true — including right after a previous `applySwitch` reopened the engine.

### 4. `accountsBalances[id] != nil` is satisfied by spendable-masked zero balances

`WalletSummary.withSpendableMasked()` (SDK: `Rust/WalletSummary+FFI.swift:49-74`) rewrites every account's `spendableValue`/`unshielded` to `.zero` **in place** and removes no keys, so `!= nil` stays true. It applies whenever the tip is not fresh (SDK: `SlipstreamSynchronizer.swift:733-734`; legacy path `ZcashRustBackend.swift:1222-1225`), and `tipFresh` is reset by every switch — `engine.reopen` allocates a fresh handle with `tip_fresh: AtomicBool::new(false)`.

On cold start the first non-empty map is the masked one, and the hook at `RootInitialization.swift:1308-1311` deliberately re-reads exactly that snapshot. The displayed value is the masked one (`WalletBalancesStore.swift:231` → `SendFormStore.swift:543`).

So the check means "a wallet summary exists", not "the balance is hydrated" — and the gap is widest precisely when the current server is slow, since that is when the tip refresh does not happen.

### 5. Absent balance entries still render as an authoritative zero

Separate from the reverted fix, `.synchronizerStateChanged` forwards the raw Optional and `.balanceUpdated` collapses it:

```swift
// WalletBalancesStore.swift:269
return .send(.balanceUpdated(latestState.data.accountsBalances[account.id]))
// WalletBalancesStore.swift:231
state.shieldedBalance = accountBalance?.shieldedSpendableValue ?? .zero
```

`BalancesStore.swift:186-199` duplicates this, and `SendFormStore.swift:542-543` mirrors it into `isInsufficientFunds` (`:158-161`). The sibling `.updateBalances` arm (`:219-225`) never sends nil and `SmartBannerStore.swift:1639` uses `guard let`, so this collapse is the outlier rather than the house style.

Treating an absent entry as "unknown, keep the last snapshot" would fix cold start, the automatic switch, the manual Save, and any future teardown at once — and would make the Root gate unnecessary for this symptom.

One related note on the stated root cause: on the shipping engine (`useSlipstreamSynchronizer` defaults `true`, `secant/Sources/Models/WalletConfig.swift:25`), the SDK's `SlipstreamSynchronizer.switchTo` calls `stopPolling()` directly and never emits a state that blanks balances. The window this PR really closes is cold start, where `prepare()`'s `?? [:]` leaves the map empty. The PR description's mechanism ("a switch temporarily removes the selected account and balance") holds only on the non-default legacy synchronizer.

### 6. The deferral log contradicts the gate it is diagnosing

```swift
// RootStore.swift:640 — missing the `selectedWalletAccount != nil` term the gate has
let walletHydrated = state.hydratedBalanceAccountID == state.selectedWalletAccount?.id
```

On a cold start before `.loadedWalletAccounts` selects an account, both sides are nil, so this logs:

```
[AutoServerSelection] Candidate deferred (bgTask: false, serverSetup: false, sensitiveFlow: false, hydrated: true)
```

Every printed gate reads clear while the switch was in fact refused. This is the only field diagnostic for the new behaviour, so the first report of "automatic server selection stopped working" cannot be diagnosed from logs.

The underlying cause is that the gate at `:226-231` relies on `nil == nil` being repaired by a trailing `&& selectedWalletAccount != nil` placed *after* the term it corrects. One named computed property used by both the gate and the log would remove the trap and the divergence together:

```swift
var isSelectedAccountBalanceHydrated: Bool {
    guard let id = selectedWalletAccount?.id else { return false }
    return hydratedBalanceAccountID == id
}
```

### 7. Re-deferring a candidate restarts its 15-minute TTL

`onChange` (`RootStore.swift:566-573`) re-sends only the endpoint:

```swift
return .send(.autoServerCandidateReady(pending.endpoint))   // benchmarkedAt does not travel
```

If the gate closes again before that action is reduced, the guard at `:634` builds a fresh `PendingServerCandidate(endpoint:, benchmarkedAt: date.now())`. Repeat the cycle and `pending.isExpired` at `:568` can never fire against `pendingCandidateTTL` (15 min, `AutoServerSelectionInterface.swift:36`), so an hours-old benchmark is eventually applied as fresh. The new hydration term makes this bounce considerably more frequent.

**Fix:** carry `benchmarkedAt` through the re-send.

### 8. The gate is point-in-time; the 60-second rebuild window is unguarded

`canApplyAutoServerSwitch` is read once, then:

```swift
// RootStore.swift:646
return .run { _ in _ = await autoServerSelection.applySwitch(candidate) }
```

No `.cancellable(id:)`, no ID, and `applySwitch` re-validates automatic-mode / endpoint / migration pinning but never navigation state. It runs under `withTimeout(serverSwitchTimeout)` = 60s (`TransactionGuard.swift:81`). Grepping `secant/Sources` for `isSwitching|switchInFlight|InFlight|isRebuilding` finds no in-flight marker anywhere, so a user entering Send one tick after the gate was read hits the whole rebuild unprotected.

Publishing "a switch is in flight" from `AutoServerSelectionClient`/`TransactionGuard`, so balance consumers hold their last snapshot, is what actually protects that window.

### 9. A wallet that has never reached lightwalletd can never open the gate

Narrower than it first looks, and the obvious framing is wrong — `fully_scanned_height < 0` is the `WalletSummary::none()` sentinel, not a "never scanned" signal, and once a summary exists every account gets an `AccountBalance::ZERO` entry regardless of scan progress. An established wallet whose server later degrades still hydrates offline.

What survives: `get_wallet_summary` returns `None` without scan progress, and progress needs either a scanned `blocks` row or a completed subtree whose `subtree_end_height` is written only by `put_shard_roots` ← lightwalletd's `GetSubtreeRoots`. So the gate is permanently false for a fresh install or restore whose configured server was dead from first launch — reachable, since the wallet is created offline from the bundled checkpoint (SDK: `Initializer.swift:488-525`), init completes, and `.refreshAutomaticServer` fires.

Pre-PR the gate was just `bgTask == nil && !isServerSetupVisible && !isSensitiveFlowActive`, so the switch would have applied. This is exactly the case where automatic selection is the user's only escape. (Verified statically only.)

### 10. `.retryStart` can benchmark with no state-stream subscriber

`didEnterBackground` cancels `state.CancelStateId` (`:232`). On the next foreground, `.retryStart` re-subscribes only conditionally:

```swift
if startedSyncThisPass {
    await send(.initialization(.registerForSynchronizersUpdate))   // :844-846
}
...
await send(.refreshAutomaticServer)                                 // :858 — unconditional
```

When `visitKind() == .send`, sync is deliberately not started, so there is no subscriber to `stateStream()` for that whole foreground — and `.synchronizerStateChanged` is one of only two setters of the latch. The latch usually survives from the previous foreground, which is what keeps this from biting every time; but when it is stale, the app pays for the benchmark with no possible trigger to apply it.

---

## Test coverage

### 11. Four pre-existing gate assertions became vacuous

`Root.State.initial` never sets `selectedWalletAccount` and the `@Shared(.inMemory)` default is nil, so the new `selectedWalletAccount != nil` term makes `canApplyAutoServerSwitch` false before any test touches `path`, `signWithKeystoneCoordFlowBinding`, or `serverSetupViewBinding`.

Line 152 was correctly flipped to `XCTAssertFalse`, but lines **163, 183, 190 and 199** were left asserting something the hydration term now guarantees on its own. Delete `!isSensitiveFlowActive` and `!isServerSetupVisible` from the gate entirely and the suite still goes green — while shipping a switch that rebuilds the synchronizer mid-Send or mid-vote.

There is no `TestStore`-level test of `.autoServerCandidateReady` anywhere, so these state assertions were the whole safety net. A fixture that pins an account and sets `hydratedBalanceAccountID` would restore them.

### 12. The regression test for the reported bug was added and deleted in the same PR

`ceb63ccc` added `updateBalancesPublishesCachedSnapshotBeforeDatabaseRefreshCompletes()` (52 lines) pinning that `.updateBalances` emits the cached snapshot before the database refresh resolves. `1bdf2f9a` removed all 52 lines.

The replacement suite exercises only the in-memory `canApplyAutoServerSwitch` boolean and never touches `WalletBalances`, `sdkSynchronizer`, or a switch. So a change that reorders `.updateBalances` back — or that removes the Root gate — leaves everything green.

### 13. The new test leaves the gate's account-equality unpinned

```swift
#expect(!state.canApplyAutoServerSwitch)
state.hydratedBalanceAccountID = account.id
#expect(state.canApplyAutoServerSwitch)
```

Only the nil→match transition is covered. Weaken the gate to `hydratedBalanceAccountID != nil && selectedWalletAccount != nil` and both assertions still pass — while a switch now applies immediately after a Zashi→Keystone switch using hydration recorded for the *previous* account. The equality is the only thing that closes the gate on an account switch, since `HomeStore.swift:190/219/226` and `AddHWWalletStore.swift:184` write the shared account without touching the latch.

A case that latches account A, selects account B, and expects the gate closed would pin it.

### 14. The new suite mutates process-global `@Shared` state without isolation

```swift
state.$selectedWalletAccount.withLock { $0 = account }
defer { state.$selectedWalletAccount.withLock { $0 = nil } }
```

`@Suite(.serialized)` only orders tests *within* the suite; other suites still run concurrently against the same `InMemoryStorage`. ~20 other files write this same key. Either direction fails: another suite's write between the two `#expect`s fails this test with no local cause, or this `defer` nils another suite's in-flight account. The `defer` also hardcodes `nil` rather than restoring the previous value.

The repo already has the convention, documented in a comment at `RootIronwoodAnnouncementGateTests.swift:26-30` and used at 107 call sites:

```swift
withDependencies { $0.defaultInMemoryStorage = InMemoryStorage() } operation: { ... }
```

---

## Conventions

### 15. Changelog entry

Two CLAUDE.md rules:

- *"Prefix every line with the issue identifier in brackets, e.g. `- [MOB-1321] …`"* — the entry uses `[Automatic server]`, though the PR title supplies MOB-1831.
- *"Write from the user's perspective — what changed for them and why it matters — not the implementation detail."* — the diff is self-documenting here: `ceb63ccc` wrote a symptom a user recognises, and `1bdf2f9a` replaced it with "no longer rebuilds the synchronizer before … the local balance snapshot have hydrated … endpoint application waits for a safe, hydrated state". Compare the neighbouring entries, which describe the button, the screen and the wait.

Also worth updating alongside: CLAUDE.md's Transaction-guard paragraph enumerates this gate's terms ("sensitive flow on screen, Server Setup visible, or a background task") and is now incomplete, as is the `var body` doc comment at `RootStore.swift:561-563`, which lists the gate's inputs as "(path, bindings, bgTask, settings path)".

---

## Checked and cleared

For the record, so these do not get re-raised:

- **`sdkSynchronizer.latestState()` in a reducer** — no new concurrency or Sendable issue; ~20 pre-existing call sites already read it this way, and it is a synchronous non-isolated property read.
- **`latestState` in tests** — no breakage. Every Root test that can reach `.loadedWalletAccounts` uses `.mocked(...)` or `.noOp`, both of which supply `latestState`, and all run with `exhaustivity = .off`.
- **The `onChange` wrapper** — correct. `_OnChangeReducer` snapshots `toValue(state)` eagerly before and after the base reduce, and `initializationReduce()` is composed inside `core`, so both new writers are observed.
- **Account switching stranding the latch** — the mechanism is real (no switch path writes `hydratedBalanceAccountID`), but Slipstream's 2-second poll emits unconditionally, so the gate reopens well inside the TTL. A one-line write in `accountSwitchedEffect` would still tidy it.
- **The `"Must run above the selectedWalletAccount guard…"` comment** — not displaced; the new block was inserted above it and it still sits on `presentIronwoodAnnouncementIfNeeded`.
- **The three `.initializeSDK` exits that clear the latch before `isInitializingSDK = true`** — all documented dead ends recoverable only by relaunch, so the cleared latch has no consequence.
- **Line length and file length** — no new violations.

secant/Sources/Features/Root/RootStore.swift
```diff
@@ -610,26 +601,29 @@ struct Root {
...
                 guard state.bgTask == nil, !state.isServerSetupVisible else { return .none }
                 return .run { send in
-                    if let best = await autoServerSelection.findBestServer() {
-                        await send(.autoServerCandidateReady(best))
+                    async let candidate = autoServerSelection.findBestServer()
+                    async let localBalances = try? sdkSynchronizer.getLocalAccountBalances()
+                    let (best, snapshot) = await (candidate, localBalances)
+
+                    guard let best else { return }
+                    guard snapshot != nil else {
```
zcash-harry, line 614: If there is no snapshot, the winner candidate is dropped. Open to change if there is a better option than this.

secant/Sources/Features/WalletBalances/WalletBalancesStore.swift
```diff
@@ -217,29 +217,39 @@ struct WalletBalances {
...
+                // Use the unmasked local snapshot for display continuity. The SDK's visible
+                // balances can be spendable-masked until the new server reports a fresh tip.
+                let cachedBalance = sdkSynchronizer.latestState().localAccountsBalances[account.id]
                 return .run { send in
-                    if let accountBalance = try? await sdkSynchronizer.getAccountsBalances()[account.id] {
-                        await send(.balanceUpdated(accountBalance))
-                    } else if let accountBalance = sdkSynchronizer.latestState().accountsBalances[account.id] {
-                        await send(.balanceUpdated(accountBalance))
+                    if let cachedBalance {
+                        await send(.balanceUpdated(cachedBalance))
```
zcash-harry, line 225: Publishes the SDK’s cached local balance first. Then  refreshes from the database and replaces it only when the value changed.

secant.xcodeproj/project.pbxproj
```diff
@@ -1359,15 +1359,14 @@
...
-			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "https://github.com/zodl-inc/zodl-swift-wallet-sdk.git";
-			requirement = {
-				kind = revision;
-				revision = 7e44c48aed237202043010775ec7383c45c519fe;
-			};
+/* Begin XCLocalSwiftPackageReference section */
+		3464A8FA303C3A54005A09AD /* XCLocalSwiftPackageReference ".ci/zodl-swift-wallet-sdk" */ = {
+			isa = XCLocalSwiftPackageReference;
+			relativePath = ".ci/zodl-swift-wallet-sdk";
```
Chlup: Remote SDK should be used. So first merge the SDK PR for this and then change this back to remote with specific commit. If if there is a reason to use local SDK use `../zodl-swift-wallet-sdk` path for it.

Chlup (approved):
utACK

PR #2044, opened 2026-08-30
Fix e2e driver-stability CI failures on maint/3.10.x

zcash-harry:
## Summary
Ports the three-commit CI fix from #1981 onto `maint/3.10.x` directly, so any build/PR based on this branch gets the fix without waiting on #1981 to land on `main`.

- Pin CI Xcode to 26.3 (`.github/workflows/*.yml`)
- Fix a Swift 6.0 type-check timeout in `MigrationArrivedWindowCaptionTests` that only shows up under Xcode 26.3's older compiler
- **The actual fix**: pin the e2e simulator to the iOS 26.2 runtime explicitly (Xcode's toolchain selection alone doesn't change which Simulator runtime gets used — CoreSimulator runtimes are shared system-wide across every installed Xcode), and pin Maestro CLI to 2.7.0 instead of floating on "latest"

Root cause: Xcode 26.5/26.6 both ship the iOS 26.5 Simulator runtime, which has a confirmed upstream Maestro/XCUITest driver-stability bug (mobile-dev-inc/maestro#3318, #3254) causing repeated e2e-smoke driver crashes and hangs. Verified fixed across a full green phase-1 + phase-2 e2e run on #1981.

## Test plan
- [x] Cherry-picked cleanly from #1981's validated commits (`6ba7f94f`, `581d6d6b`, `9eae4812`)
- [x] Local YAML validation + Swift syntax check pass
- [ ] CI on this PR should confirm the pin holds on `maint/3.10.x`'s own tree

### DIFF
--- a/.github/workflows/e2e-smoke.yml
+++ b/.github/workflows/e2e-smoke.yml
@@ -133,16 +133,28 @@ jobs:
     permissions:
       contents: read
     steps:
-      - name: Select latest Xcode
-        # macos-15 runners ship multiple Xcodes; default-selected may be 16.x.
-        # XCUITest from older Xcode driving an iOS 26 sim causes maestro tap/
-        # input quirks (empty seed fields). Pin to the newest Xcode available.
+      - name: Select Xcode (pinned, avoiding iOS 26.5 driver bug)
         run: |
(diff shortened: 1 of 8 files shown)


Chlup (approved):
utACK

PR #2049, opened 2026-08-31
[MOB-1831] Evaluate benchmark results before automatic server switches

Chlup:
## What

Automatic server selection no longer switches blindly to the benchmark winner. The switch decision now lives in the SDK (`evaluateServerSwitch`): the app switches only when the new server is meaningfully faster than the current one — at least **200 ms and 25%** better in the benchmark — or when the current server fails its health checks. A ~5 ms win no longer tears down the synchronizer.

## Why

`switchTo(endpoint:)` rebuilds the whole synchronizer, which is why the transaction guard and the Root gating machinery exist. Paying that cost for a marginal (and, on single-shot measurements, mostly noise-driven) improvement was all downside. The benchmark's timings never left the SDK, so the evaluation had to move there rather than into the app.

## How

- `findBestServer` keeps only its app-level duties — the Automatic-mode flag check and the migration-pinning candidate filter (N4) — and delegates the decision to `sdkSynchronizer.evaluateServerSwitch(current, candidates, …)`, returning its answer as-is (`nil` = stay).
- New `SDKSynchronizerClient.evaluateServerSwitch` closure (interface + live passthrough + test-key mirrors).
- **Unchanged:** `applySwitch` and its re-validation, all Root gating/TTL machinery, and the Server Setup screen — an explicit "Automatic + Save" still simply picks the measured best.
- `AutoServerSelectionConstants.candidateCount` is gone: the decision path benchmarks every candidate, so "current missing from results" reliably means unhealthy rather than "lost a top-3 spot by 2 ms".

## Tests

- New Swift Testing suite `AutoServerSelectionFindServerTests` (4 cases: flag-off short-circuit, nil propagation, candidate propagation, delegation of current/candidates/constants).
- Superseded XCTest `findBestServer` cases removed; `applySwitch` and Root gating tests untouched and green.
- Decision-rule coverage (13 table-driven cases) lives with the policy in the SDK PR.

## Pairing / CI

Depends on the paired SDK PR: zodl-inc/zodl-swift-wallet-sdk#31. `main` now consumes the SDK as a remote package (`upToNextMinorVersion` from 4.1.0), so the pairing is expressed the remote way: this branch points the `zodl-swift-wallet-sdk` package reference at the exact revision `d784165d` (the SDK branch head this branch was built and tested against) and `.sdk-pin` stays empty, as remote mode requires. Once the SDK side merges and ships in a release, restore the version-range requirement.


LukasKorba (approved):
utACK

PR #2050, opened 2026-09-01
Switch to the remote SDK at 4.1.0

Chlup:
Flips the app's `ZcashLightClientKit` dependency from the pinned-local companion checkout to the remote `zodl-swift-wallet-sdk` package, now that 4.1.0 is tagged (the previous local pin was at `4.1.0-rc.3`).

## What changed

- **`secant.xcodeproj`**: the `XCLocalSwiftPackageReference` to `.ci/zodl-swift-wallet-sdk` is replaced with an `XCRemoteSwiftPackageReference` to `https://github.com/zodl-inc/zodl-swift-wallet-sdk.git`, with an up-to-next-minor requirement from **4.1.0**.
- **`Package.resolved`**: pins `zodl-swift-wallet-sdk` at 4.1.0 (`e948e76`).
- **`.sdk-pin`** is emptied. The pin file records the companion checkout's commit only in local mode; in remote mode both the "Update SDK Pin" build phase and CI's validation (`setup-sdk-companion.sh`: "remote: pin must be empty") expect it empty. Leaving the stale rc pin in place would fail CI's validate step and dirty every builder's working tree on their first build.

Dependency wiring only — no source changes, and no changelog entry since the changelog tracks user-facing modifications.


LukasKorba (approved):
utACK

PR #2057, opened 2026-09-03
Release 3.11.0

Chlup:
Release `3.11.0`, following `3.10.3`.

The base of this pull request is `release/3.11.0`, which starts out
identical to `3.10.3`. Its diff is therefore exactly what users receive
relative to the last release, rather than the intervening development history.

On top of the released content this branch carries:

- the CHANGELOG's Unreleased section, promoted to `## [3.11.0]`;
- `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION`, set across every
  non-test target.

## Do not merge before the build has shipped

TestFlight and App Store builds for this release are made from
`candidate/3.11.0`, the head of this pull request:

    ./Scripts/release.sh --variant internal-testnet --ref candidate/3.11.0 \
        --version 3.11.0 --build <n>

Fixes found in testing land on `candidate/3.11.0` and appear here. This
pull request stays a **draft** while builds are in progress; marking it ready
for review is part of declaring the release final, and merging it is what makes
it final. Tag `release/3.11.0` afterwards, then merge it back into
`maint/v3.11.x` and forward to `main` so the tag stays reachable. Make each
merge with `--no-commit` and check that `## [3.11.0]` in the CHANGELOG
still matches the tag before committing — entries added on the target branch
during the release cycle merge into the published section without a conflict.


LukasKorba (approved):
utACK

PR #2071, opened 2026-09-05
[MOB-1848] Package of fixes to improve overall performance and stability

Chlup:
This PR lands the iOS side of the sync-contention series (Linear MOB-1848) on `main` in one merge commit. The branch is the top of the reviewed stack plus one merge of `main` that resolves the SDK pin and the instructions file.

## What the series fixes

- **MOB-1853** Automatic server switching no longer restarts a sync that is in progress; when the wallet's own recovery gives up, the app rebuilds the connection (to a better server when one exists) instead of spinning forever, shows a "Sync has stalled" banner with Retry, and keeps the banner honest while a retry runs or the app goes to the background.
- **MOB-1854** Sync resumes after a migration broadcast even when the resume request arrives while the previous start is still finishing; backgrounding during a restart cannot leave that restart running.
- **MOB-1855 / MOB-1856** The transaction list no longer empties or sticks on placeholders while the wallet catches up, and the app stays responsive during a long history refresh.
- **MOB-1857** Insufficient-funds and failed-payment messages are correct again.
- **MOB-1858** Sending no longer waits behind unrelated network work; the transaction guard covers only the broadcast, and transactions the app created but could not hand over are released to the SDK's background resubmission.
- **MOB-1859** Opening the wallet no longer generates a fresh receive address for every account on each load.
- **MOB-1860** Leaving a voting screen while a proof is being prepared stops that work.
- **MOB-1862** Balances: a loading balance for a previous account can no longer be shown as the new account's; the breakdown shows the same spendable and pending amounts as the home screen; funds awaiting confirmation no longer leave the balance spinning; swapping into ZEC no longer waits for spendability.
- **MOB-1863** A sent transaction shows "Sent · awaiting confirmation" once a server has accepted it.
- **MOB-1869** Send, Swap and Pay wait for the spendable balance the way the home screen does instead of reporting insufficient funds while it is still being worked out.
- **MOB-1848** The SDK pin moves to the SDK `main` commit that carries the matching SDK series (landed via zodl-swift-wallet-sdk #45).

`CHANGELOG.md` carries the user-facing entries under Unreleased.

## How it was reviewed

The work was reviewed as 21 stacked pull requests, one per change, each approved in place: #2083–#2102, #2104, #2105. Review fixes landed on the affected level and were merged forward, so this branch is exactly the sum of those PRs. The stacked PRs were retargeted to `main` so that GitHub marks them merged when this PR lands.

## How to merge

Merge with a merge commit. The stacked PRs close as merged by themselves and the level branches are deleted afterwards. This branch deliberately does not contain the two manual version-bump commits that were on the integration branch before; version bumps stay with the release process.

## The merge of `main`

`main` released 3.12.0 while the stack was under review. The merge commit keeps the released changelog sections untouched and re-homes the stack's entries under Unreleased; `.sdk-pin` keeps the series' SDK pin, `CLAUDE.md` takes `main`'s stub, and the three transaction-guard notes the series added to the old `CLAUDE.md` moved into `AGENTS.md`.

## Test plan

- Build for testing and the whole `zodlTests` target against the pinned SDK, built locally from source: 1829 Swift Testing tests in 245 suites passed, XCTest 37 tests, 0 failures.
- CI runs on this head.


LukasKorba (approved):
utACK

PR #2083, opened 2026-09-07
[MOB-1853] Apply automatic server switches only when sync is idle and cancel the refresh at background (stack 01/21)

Chlup:
> **Stacked PR 01/19** of the sync-contention series (umbrella #2071). Base: `main`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Also carries the trivial segment(s) merged just before it: repin-sdk-contention-tip.
>
> Same change as #2059 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Linear: MOB-1853.

**What:** automatic server switches now apply only while the synchronizer is idle (`upToDate`, `stopped`, `unprepared`, `error`), tracked from the state stream; a candidate that arrives mid-sync is parked and re-applied by the existing gate observer when sync settles. The automatic refresh is cancelled when the app goes to background, so a benchmark started before backgrounding cannot switch servers later. When the SDK reports that a sync pass stalled and its own reconnect did not help (second attempt or give-up), the app treats the sync as idle and runs a benchmark, so a healthier server can be chosen without waiting for the next foreground. The stall signal has its own event channel, separate from the throttled transaction events.

**Why:** the gate never checked whether a pass was in flight, so a switch could tear down a sync in progress; and once switches are idle-only, a stalled pass would otherwise block the only recovery an app that gives the engine no alternate servers has.

**Test plan**
- `RootAutoServerIdleGateTests` (new, 10 tests): candidate parked while syncing and applied once on `upToDate`; `nil` status keeps the gate closed; background cancels the refresh effect; attempt-1 stall is logged only (no benchmark call); attempt-2 and give-up open the gate and trigger a refresh; the stalled flag clears on `upToDate` and on progress. `AutoServerSelectionClientTests`, `RootAutoServerCandidateTests` and the four `RootTransactionsTests` suites pass.
- Manual (testnet): while a restore is syncing, change nothing and watch the log — a benchmark winner is logged as deferred with `idle: false` and applied after the sync reaches `upToDate`.


LukasKorba (approved):
utACK

PR #2084, opened 2026-09-07
[MOB-1854] Guard retryStart against re-entry (stack 02/21)

Chlup:
> **Stacked PR 02/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1853-idle-gated-server-switch-stack-01`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2060 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Linear: MOB-1854.

**What:** the `retryStart` pipeline (migration visit, synchronizer start, stream registration, automatic server refresh) now runs one at a time. A `retryStart` that arrives while one is in flight is logged and dropped; the in-flight pipeline performs the same work. The flag is cleared by a finishing action on both exit paths of the effect and reset when the app goes to background, so a dropped send can never wedge the next foreground.

**Why:** `retryStart` is sent from the foreground transition, the background task, the start-failure retry and the migration-gate resume, so two start pipelines could overlap and each could tear down and restart the sync. Cancelling the first pipeline was rejected: the SDK's `start()` has no cancellation points, so the first start would complete anyway and the second would restart the pass.

**Test plan**
- `RootRetryStartReentrancyTests` (new, 3 tests, real suspension gate on `start`): two back-to-back `retryStart` call `start` once; a third after the finishing action proceeds; after background a `retryStart` proceeds even if the previous pipeline never finished. Full `zodlTests`: 1597 tests passed (2 pre-existing known issues).
- The `SDKSynchronizerClient.mocked(start:)` test helper now accepts an `async throws` closure to match the real dependency; existing callers are unaffected.


LukasKorba (approved):
utACK

PR #2085, opened 2026-09-07
[MOB-1855] Keep the transaction list through spurious empty fetches and clear invalidation on failure (stack 03/21)

Chlup:
> **Stacked PR 03/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1854-retry-start-reentrancy-stack-02`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2061 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Linear: MOB-1855.

**What:** `.fetchedTransactions` no longer replaces a non-empty transaction list with an empty fetch result while the synchronizer is not up to date and neither list is invalidated by an account switch. On that path the rows are kept, the event is logged, both lists still receive `transactionsUpdated`, and the 30 s reconciliation poller is computed from the kept rows instead of the empty array. A failed `getAllTransactions` now sends a new `transactionsFetchFailed(accountUUID:)` action that clears both lists' loading placeholders and re-arms the poller, after the same account-provenance guard the success path uses. The seed fetch logs its early return when no account is selected yet.

**Why:** during a restore or a restart of the sync engine the database can briefly report zero rows while the engine still holds unreconciled transactions; overwriting the list with that result reads as data loss. The emptiness test is made on the raw on-chain rows rather than the merged array, because the merged array also carries one synthetic row per in-flight swap-to-ZEC, which would defeat the guard. A thrown fetch used to leave the loading placeholders on screen indefinitely and arm no poller. The provenance guard on the failure path prevents a stale failure for an account the user just switched away from marking the new account's lists as loaded while the old rows are still shown.

**Test plan**
- `RootTransactionsEmptyFetchTests` (new, 6 tests): empty fetch while syncing keeps the rows and keeps the poller; empty fetch while up to date empties the list; empty fetch while a list is invalidated replaces the list; fetch failure clears both `isInvalidated` flags and schedules the poller; failure for a non-selected account changes nothing; empty on-chain fetch with an in-flight swap-to-ZEC keeps the on-chain rows. The four sibling `RootTransactionsTests` suites still pass (29 tests).


LukasKorba (approved):
utACK

PR #2086, opened 2026-09-07
[MOB-1856] Read transaction outputs once per row and coalesce concurrent refreshes (stack 04/21)

Chlup:
> **Stacked PR 04/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1855-transaction-list-empty-fetch-guard-stack-03`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2062 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Linear: MOB-1856.

**What:** the transaction-state mapping in the live synchronizer dependency reads each transaction's outputs once and derives the recipients from them, instead of issuing a second per-row query. Transaction-list refreshes are now coalesced: while one `getAllTransactions` fetch is in flight, further `fetchTransactionsForTheSelectedAccount` requests only mark the state dirty, and a single follow-up fetch is dispatched when the in-flight one completes (success or failure). The account-switch path resets the coalescing flags before it cancels the running fetch, because actions sent from a cancelled effect are dropped and would otherwise leave the gate closed. The comment that used to justify the missing `cancelInFlight` now explains the coalescing and why the account-provenance guard stays.

**Why:** during catch-up sync every throttled synchronizer event re-dispatched a full-history refresh, and each refresh queried the wallet database twice per transaction. Those fetches piled up while the sync engine was writing the same database, which is one of the contention sources behind the stalls and the sluggish UI on wallets with long histories. The SDK's `getRecipients` is defined as the outputs mapped to their recipient, so the mapping result is unchanged.

**Test plan**
- `RootTransactionsCoalescingTests` (new, 3 tests, real suspension gate on the mocked fetch): three dispatches before completion make one call and exactly one follow-up afterwards, and the in-flight fetch's own payload still lands; the failure path coalesces the same way; an account switch while a fetch is parked starts the new account's fetch immediately and the stale completion changes nothing. Assertions await the dispatch's `StoreTask`, not wall-clock sleeps; passed in three consecutive runs.
- The pre-existing test that asserted two concurrent same-account fetches both land is removed: coalescing makes that scenario impossible by design, and the property it protected (the running fetch's own result lands after a later dispatch) is now asserted in the new suite.
- `SDKSynchronizerClient.mocked(getAllTransactions:)` now takes an `async throws` closure to match the real dependency; existing callers are unaffected.
- The seven suites under `RootTransactionsTests` plus `RootRetryStartReentrancyTests`: 40 tests passed.


LukasKorba (approved):
utACK

PR #2087, opened 2026-09-07
[MOB-1857] Detect insufficient funds by error kind and never swallow a send failure (stack 05/21)

Chlup:
> **Stacked PR 05/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1856-coalesce-transaction-refresh-stack-04`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2063 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Linear: MOB-1857.

**What:** `ZcashError.isInsufficientBalance` now recognises the SDK's typed `rustProposalInsufficientFunds` case first and keeps the two legacy text matches as a fallback. In the Send form, a failed send or payment request always presents the failure alert; previously the alert was shown only for the `.send` confirmation type.

**Why:** the SDK redacts Rust error text, so the string match no longer fires for a real insufficient-funds proposal and the user saw a generic failure instead of the dedicated message. For the `.requestPayment` path no reducer presented the failure at all, so tapping Send could silently do nothing. The coordinator that also observes this failure only pops the navigation path and does not present an alert, so there is no double alert.

**Test plan**
- `ZcashErrorInsufficientBalanceTests` (new, 3 tests): the typed case is detected, the legacy message text is still detected, `rustProposalScanRequired` is not.
- `SendFormSendFailedAlertTests` (new): a failure with the `.requestPayment` confirmation type sets the alert.
- The existing `SendFormAddressValidationTests` and `SendFormMaxButtonTests` still pass; 17 tests across the four suites.


LukasKorba (approved):
utACK

PR #2088, opened 2026-09-07
[MOB-1858] Hold the submission guard only across the broadcast, with an acquisition timeout (stack 06/21)

Chlup:
> **Stacked PR 06/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1857-typed-insufficient-funds-stack-05`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2064 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Linear: MOB-1858. Two commits.

**What**
1. `TransactionGuard` gains `acquire(timeout:)`: the waiter is registered and a deadline task is armed under actor isolation; on expiry the deadline removes the waiter only if it is still queued and fails the acquisition with `TransactionGuardBusyError` ("The wallet is busy with another network operation. Try again in a moment.", en + es in the string catalog). If a `release()` already handed the waiter ownership, the deadline is a no-op, so ownership can neither leak nor be released twice. `TransactionGuardClient.withSubmission(timeout:_:)` exposes it; the existing `withSubmission`, `switchIfIdle` and `switchWaiting` are unchanged.
2. The two send paths in the live synchronizer dependency (proposal-based and PCZT) run proposal creation and proving before taking the guard and hold it, with a 30 s acquisition timeout, only across the store-and-submit step. A busy guard maps to the existing `.failure` result with a distinct code. The voting API client now acquires the guard per attempt inside its retry loop, so back-off sleeps no longer hold it. The repository `CLAUDE.md` describes the new scope and timeout.

**Why:** the guard is one process-wide mutex shared by broadcasts and server switches. Holding it across zk proving and across retry sleeps made a user's send wait behind work it had nothing to do with, and a stuck holder blocked every submission in the app with no way out. Proving runs through the Rust backend on the wallet database and the submit plan store, neither of which a server switch tears down, so only the broadcast needs exclusion. Racing the old `acquire()` against an external timer could have leaked the guard when a release handed ownership to the parked waiter as the timer fired, which is why the deadline lives inside the actor.

**Test plan**
- `TransactionGuardTests` (8 new tests, 16 total): a held guard makes `withSubmission(timeout: 100 ms)` throw the busy error within 300 ms and a fresh submission then succeeds; a release at 90 ms racing a 100 ms deadline leaves the guard free whichever wins; cancellation of a parked timed waiter; FIFO order after a timed-out waiter; the send seam proves before acquiring and releases after submitting (event log).
- Wider regression run: 152 Swift Testing tests in 12 suites plus 5 XCTest tests, 0 failures.

**Notes for reviewers:** the busy text does not reach the Send failure screen because that screen shows fixed copy and the confirmation reducer overwrites the failure description before the support report reads it; that is pre-existing and left for a follow-up. The Flexa payment path shows its generic failure message for a busy guard; a one-line special case on the new failure code would recover the busy text there if wanted.


LukasKorba (approved):
utACK

PR #2089, opened 2026-09-07
[MOB-1859] Stop writing a fresh diversifier on every account load (stack 07/21)

Chlup:
> **Stacked PR 07/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1858-scoped-submission-guard-stack-06`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2065 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Linear: MOB-1859.

**What:** loading the wallet's accounts no longer derives a "next private address" for every account. The live synchronizer dependency returns accounts with an empty address stash, and the three places that replace the shared account list merge each account's existing stash back by id. The shared account list is now the source of truth for that stash: one helper writes a stash value into the list entry and mirrors it into the selected-account copies, and it is used by the Root load path, by the Receive and Swap screens' refills, and by the promotion step that consumes a stash (which now also clears it in the list, so an account switch can never re-show an address already displayed). After a load, Root refills only the accounts whose stash is still empty, in the background, cancellable, never awaited on the load path. The transaction-existence check is unchanged because the SDK offers no single-transaction lookup.

**Why:** deriving the next address persists the diversifier index, which is a wallet-database write. It ran for every account on every load, including twice on cold start, while the sync engine was writing the same database, behind a 15 s busy timeout. That made opening the wallet slow during catch-up and added contention that the Receive screen fix (MOB-1803) had already identified. Keeping the stash in the shared list means the writes happen once per account and their results are never discarded.

**Test plan**
- `WalletAccountStashTests` (pure merge helper, 5 tests) and `RootLoadedWalletAccountsStashTests` (2 tests): the load path calls address generation only for accounts without a stash, the refilled address is observable in the list entry and in the selected-account copies, and a second load with both stashes present makes zero generation calls.
- `HomeReceiveRotationTests` gains a test that promotion clears the list entry's stash and the refill writes it back.
- `SwapAndPayQuoteRotationTests`, `AddKeystoneHWWalletTests`, `RootTransactionsAccountSwitchTests`, `RootInitializeSDKColdStartFetchTests`, `SmartBannerLadderReadinessTests` still pass; 41 tests across the 8 suites.


LukasKorba (approved):
utACK

PR #2090, opened 2026-09-07
[MOB-1860] Cancel an abandoned voting proof instead of letting it run boosted in the background (stack 08/21)

Chlup:
> **Stacked PR 08/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1859-no-diversifier-write-on-load-stack-07`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2066 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Linear: MOB-1860. Voting only.

**What:** the delegation-proof stream in the voting crypto dependency now keeps the handle of the detached task that produces it, cancels that task when the stream is terminated by its consumer, checks for cancellation before starting the proving work, and checks once more right before entering the Rust FFI call. A proof that is already inside the FFI cannot be interrupted; the fix prevents the next orphan from starting and stops the abandoned producer from yielding into a stream nobody reads. The stream construction is extracted into an internal factory so the wiring is testable with a spy prover; the live value passes the real backend work into it. The QoS boost and the key-generation task are untouched.

**Why:** the coordinator consumes the stream under `cancelInFlight`, so leaving a screen or restarting a proof cancelled only the consumer. The producer kept running an all-cores proof under the interactive QoS boost while the next proof started, and each cycle added another orphan.

**Test plan**
- `VotingCryptoClientLiveKeyTests` (new, 3 tests): cancelling the consumer while the prover is parked is observed by the prover within 100 ms; cancelling before the producer starts leaves the prover either never invoked or invoked and cancelled, with no elements emitted; an uncancelled run yields the progress values and the completed proof in order. Assertions await latches, not fixed sleeps; passed in five consecutive runs.
- `VotingCoordFlowCoordinatorTests` (78 tests) still pass.


LukasKorba (approved):
utACK

PR #2091, opened 2026-09-07
[MOB-1862] Balance breakdown parity, updating signal, and pending-funds spendability (stack 09/21)

Chlup:
> **Stacked PR 09/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1860-cancel-abandoned-voting-proof-stack-08`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2067 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Linear: MOB-1862. Three commits.

**What**
1. The balance breakdown sheet publishes from the same unmasked local balance snapshots the home balance already uses, and its instant read prefers the local read with the same fallback order. A replayed empty synchronizer state no longer publishes a zero balance into the breakdown.
2. The SDK's "spendable value is masked" flag is mirrored into the wallet-balances state. While it is set, the available-balance row shows its updating indicator, the Send form is not valid, "insufficient funds" is not shown, and the Max chip shows its in-flight state. No new component, string or asset.
3. "Processing zero available balance" is redefined in both balance stores as "the spendable value is still being determined": the flag is masked, or a sync is running and no concrete balance has been published yet for the selected account. Funds that are merely pending confirmation therefore count as something to spend, with a spendable amount of zero and no spinner. The spendability value is re-derived whenever those inputs change, not only when a balance is published, so the indicator and the Send gates cannot lag behind the flag. The shielding-threshold early return is removed because no shielding path read this predicate.

**Why:** after the home balance moved to unmasked local snapshots, the breakdown sheet still read the masked visible balances and mapped a missing entry to zero, which reached "everything spendable" on a replayed empty snapshot. The old "shielded spendable zero while total non-zero" rule drove a spinner that never cleared for funds pending confirmation and blocked the Send form. The SDK now says explicitly when the spendable value is masked, so the UI can show "updating" only while that is true.

**Test plan**
- `BalancesTests` (extended) and new `WalletBalancesMaskedSpendableTests`: replayed zero state publishes nothing; masked state mirrors the flag, turns the indicator on, makes the Send form invalid and not "insufficient"; unmasked returns to normal; a pending self-shield yields "something to spend" with no spinner; a masked state with no local entry and an account switch mid-sync both re-derive spendability.
- Full `zodlTests` target: 1644 tests passed, 2 pre-existing known issues.

**Notes for reviewers:** `WalletBalances.State.autoShieldingThreshold` is now write-only; it is left in place because removing it would delete an existing eligibility test, and is flagged as a follow-up.


LukasKorba (approved):
utACK

PR #2092, opened 2026-09-07
[MOB-1863] Show sent transactions as awaiting confirmation once a server accepted them (stack 10/21)

Chlup:
> **Stacked PR 10/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1862-masked-spendable-breakdown-stack-09`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2068 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Linear: MOB-1863.

**What:** `TransactionState` gains an `isAcceptedBySubmitter` flag. When the live synchronizer dependency maps a transaction whose computed status is "sending", it asks the SDK for the transaction's submission status and sets the flag when a server has accepted the broadcast. The status label shows "Sent · awaiting confirmation" (en) / "Enviado · esperando confirmación" (es) for such rows instead of "Sending". Rows in any other status are untouched and make no extra query.

**Why:** the app derives "sending" from a sent transaction having no mined height, and the mined height is only learned by scanning the block the transaction landed in. Submission does not ride the sync wire, so after a successful broadcast the row said "Sending" for as long as the scan was stalled, even though the chain very likely already had the transaction. The SDK now records which server accepted a submission, so the label can say so.

**Test plan**
- `TransactionStateTests` (3 new, 22 total): sending and accepted shows the new label; sending and not accepted still shows "Sending"; a mined transaction ignores the flag.
- The suites under `RootTransactionsTests` still pass (37 tests), covering the `Equatable` change on the model.


LukasKorba (approved):
utACK

PR #2093, opened 2026-09-07
[MOB-1848] Whole-branch review follow-ups for the sync contention series (stack 11/21)

Chlup:
> **Stacked PR 11/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1863-sent-awaiting-confirmation-stack-10`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2069 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Six small commits from the whole-branch review of the integration branch, each tagged with the ticket it belongs to.

**What**
- `[MOB-1862]` The Swap and Pay form gains the same masked-spendability gates as the Send form: while the SDK masks the spendable value the form is not valid, "insufficient funds" is not shown (including the cross-pay variant that drives the red field border and the "You'll pay" label), and the Max chip shows its in-flight state.
- `[MOB-1858]` The send seam's busy log carries the caller's prefix (the PCZT path was logged under the software-send prefix); the seam tests now distinguish bounded from unbounded acquisition so a regression to the unbounded guard fails a test; a wall-clock assertion is removed; the 30 s send acquisition timeout is kept and its trade-off against the 60 s server-switch bound and unbounded migration broadcasts is documented.
- `[MOB-1859]` A failed background address generation no longer clears an address stash another path wrote: the shared refill helper skips the callback on failure, and the Receive/Swap live-fill paths skip the stash update when their second generation fails.
- `[MOB-1853]` Two comments corrected: backgrounding cancels only the server benchmark (an in-flight apply runs to completion under its own timeout), and the automatic-switch gate documentation lists its synchronizer-idle inputs.

**Why:** these are the cross-task findings of the final review: the two forms disagreed during the mask, the guard's diagnostics and tests had gaps, the new stash source of truth could be clobbered by a failed generation, and two comments overstated what the code does.

**Test plan**
- Touched suites green: `TransactionGuardTests`, `RootLoadedWalletAccountsStashTests`, `HomeReceiveRotationTests`, `SwapAndPayQuoteRotationTests`, `SwapAndPayMaxButtonTests`, `WalletBalancesMaskedSpendableTests`, `SendFormMaxButtonTests`. New tests were written first and failed for the intended reason before each fix.
- The full `zodlTests` target is re-run on the integration tip after this merge.


LukasKorba (approved):
utACK

PR #2094, opened 2026-09-07
[MOB-1862] Remove the write-only shielding threshold from the wallet-balances state (stack 12/21)

Chlup:
> **Stacked PR 12/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/final-review-fixes-stack-11`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Also carries the trivial segment(s) merged just before it: repin-sdk-merged-tip.
>
> Same change as #2072 (already merged into the integration branch); description reproduced below.

Part of MOB-1848 (Slipstream sync contention fixes). Follow-up to the whole-branch review; no user-visible change, so no CHANGELOG entry.

**What:** `WalletBalances.State.autoShieldingThreshold` and its assignment on appear are removed. After MOB-1862 redefined `isProcessingZeroAvailableBalance` as "the spendable value is still being determined", nothing read the property; the spendability derivation takes the threshold from the SDK environment directly. The breakdown sheet's `Balances.State.autoShieldingThreshold` is untouched because `isShieldableBalanceAvailable` still uses it.

**Tests:** the processing test in `WalletBalancesTests` no longer sets a threshold on its two states. The `WalletBalancesEligibilityTests` suite (one test) is deleted: it existed only to pin the removed shielding-threshold rule, and once the property is gone its case is the transparent-only shape that `WalletBalancesTests.isProcessingZeroAvailableBalance` already asserts, alongside the two genuinely unresolved states. The sibling `ShieldingEligibilityTests` suite in the same file stays.

**Verification**
- Full `zodlTests` target → 1653 tests in 225 suites passed, 2 pre-existing known issues (one test and one suite fewer than before, as intended).


LukasKorba (approved):
utACK

PR #2095, opened 2026-09-07
[MOB-1858] Release created transactions to background resubmission when the broadcast is missed (stack 13/21)

Chlup:
> **Stacked PR 13/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1862-remove-dead-shielding-threshold-stack-12`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Also carries the trivial segment(s) merged just before it: repin-sdk-audit-fixes.
>
> Same change as #2074 (already merged into the integration branch); description reproduced below.

Part of MOB-1848. Fixes the audit finding that a send which created and proved its transaction but could not hand it to a server was reported as a plain failure and left stranded.

**What:** when the submission guard stays busy past its timeout, or the send is cancelled after proving, the app now releases the exact created transactions to the SDK's background resubmission with the endpoints the user chose, and reports them as created but not yet confirmed. The confirmation screen takes its existing pending route with new copy explaining that the transaction will be handed to a server as soon as the wallet finishes its current network work; the same copy is used on the software send, the Keystone (PCZT) send and the Swap and Pay flow. An empty proving result keeps the old failure. The comment that promised a clean retry is rewritten; the endpoint selection used for submission is shared with the release path.

**Why:** proving stores the transaction in the wallet database and reserves its inputs. The SDK's background resubmission deliberately skips transactions the app never released, so a busy timeout after proving left a transaction that would neither be broadcast nor retried while a fresh attempt could find its inputs unavailable. Releasing the same transaction id cannot produce a duplicate payment.

**Test plan**
- `TransactionGuardTests` (seam): busy guard after a non-empty prove → release called once with the created transactions and intended endpoints, zero submits, result carries the created ids with the busy reason; cancellation while waiting for the guard → release; empty prove keeps the failure shape.
- `MultiServerSubmitRoutingTests` (new): the busy reason routes to the new pending copy on the software, Keystone and Swap and Pay paths; timeout and nil reasons unchanged.
- `xcodebuild … -scheme zodl-internal` build → `** BUILD SUCCEEDED **`; 73 tests in 10 suites passed.


LukasKorba (approved):
utACK

PR #2096, opened 2026-09-07
[MOB-1853] Rebuild sync after the SDK's stall recovery gives up, bounded per foreground (stack 14/21)

Chlup:
> **Stacked PR 14/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1858-release-created-transactions-stack-13`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2075 (already merged into the integration branch); description reproduced below.

Part of MOB-1848. Fixes the audit finding that a terminal recovery failure in the SDK had no effective way back to a running sync from the app.

**What:** when the SDK reports that its stall recovery has given up, the app now rebuilds the connection instead of only requesting a server benchmark: it picks a better server when automatic selection is on and one qualifies under migration pinning, otherwise the configured current server, and asks the SDK to rebuild its engine at that endpoint and start a pass, under the same transaction guard and switch timeout the automatic switch uses. A changed endpoint is persisted as the automatic switch does. At most two rebuilds run per foreground session; the counter resets on background, the rebuild is skipped during a background task or while Server Setup is open, and a parked server candidate from an earlier benchmark is dropped when a rebuild starts so it cannot replay a second switch later. After the budget is exhausted the SDK's error state stays visible. A second stall attempt without give-up keeps today's benchmark path.

**Why:** after a failed reopen the SDK had no pass and possibly no engine handle; a benchmark alone could not bring sync back (manual mode has no candidate, and switching to the current server was a no-op that also only restarted a running pass). The SDK gained `restartSync(at:)` for exactly this case; this change wires it in with a bound so it can never become an automatic retry loop.

**Test plan**
- `RootTerminalStallRebuildTests` (new): one rebuild per give-up; the two-per-foreground bound and its reset on background; skip while Server Setup is visible; a parked candidate is dropped and never replays through the automatic switch.
- `RootAutoServerIdleGateTests`: a second attempt without give-up still only benchmarks.
- `AutoServerSelectionRebuildTests` (new, live dependency): manual mode → configured endpoint; automatic with a candidate → candidate and preference persisted; automatic with no candidate or a pinning-excluded candidate → configured endpoint; a thrown restart → false; a busy guard → false.
- `xcodebuild … -scheme zodl-internal` build → `** BUILD SUCCEEDED **`; focused suites: 21 Swift Testing tests plus 5 XCTest tests passed.


LukasKorba (approved):
utACK

PR #2097, opened 2026-09-07
[MOB-1855] Clear the previous account's rows on switch and keep failed fetches truthful (stack 15/21)

Chlup:
> **Stacked PR 15/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1853-terminal-stall-rebuild-stack-14`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2076 (already merged into the integration branch); description reproduced below.

Part of MOB-1848. Fixes the audit finding that a failed refresh right after an account switch could show the previous account's transactions as the new account's history.

**What:** the shared transaction rows now record which account they belong to. Switching accounts clears the rows together with the lists' invalidation flags, so nothing renderable from the previous account survives the switch; the refreshed account's own rows set the account again when they land. A failed refresh for the selected account notifies the lists only into a truthful state (same-account rows stay; any foreign rows are cleared first, defensively), the empty-fetch keep-guard applies only to same-account rows, and the reconciliation poller is armed only from vetted rows. The empty-fetch test suite's time limit is raised to the sibling suites' three minutes and its two sleep-gated positive assertions now wait on observable state, which is what tripped the CI unit job.

**Why:** the failure path already dropped a stale failure for a previous account, but it never checked whether the rows it was about to reveal belonged to the newly selected one; both lists cleared their placeholders and rendered the old account's history, and a later legitimate empty result for the new account was then rejected by the keep-guard.

**Test plan**
- `RootTransactionsAccountProvenanceTests` (new, driving the real account-switch action): A has rows, switch to B, B's fetch fails → neither list shows A's rows; B returns an empty result while syncing → B stays empty; a stale failure for A after the switch changes nothing; a same-account failure keeps rows and the poller.
- `RootTransactionsEmptyFetchTests` passed twice in a row inside the new budget; the other `RootTransactionsTests` suites pass — 41 tests in 7 suites.


LukasKorba (approved):
utACK

PR #2098, opened 2026-09-07
[MOB-1862] Let incoming swaps request quotes while local spendability is masked (stack 16/21)

Chlup:
> **Stacked PR 16/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1855-transaction-rows-provenance-stack-15`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Also carries the trivial segment(s) merged just before it: reword-field-incident-comments.
>
> Same change as #2078 (already merged into the integration branch); description reproduced below.

## Summary

Swapping another asset **into** ZEC never spends the wallet's own ZEC, yet the quote form treated it like an outgoing swap and stayed disabled for as long as the SDK was still confirming local spendability (for example right after a server switch or a restore). This lets the incoming flow request a quote as soon as the destination and amount are filled in, while outgoing swaps and CrossPay keep waiting until spendability is known — the same split the insufficient-funds check already made.

- `SwapAndPay.State.isValidForm` exempts only the incoming (swap-to-ZEC) experience from the "spendability still being determined" gate.
- A test-only amount seam lets the form tests drive the parsed amount without the shared number formatter (interim; tracked under MOB-1873).

## Tests

Three new `SwapAndPayMaxButtonTests` cases: incoming + masked → valid; outgoing + masked → still invalid; incoming with an empty address → still invalid. The first fails on the previous gate.

Verified locally with the `zodl-internal` scheme: build succeeded; 33 tests in 3 suites passed.


LukasKorba (approved):
utACK

PR #2099, opened 2026-09-07
[MOB-1862] Attribute balance responses to their account and show the breakdown as updating while masked (stack 17/21)

Chlup:
> **Stacked PR 17/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1862-incoming-swap-gate-stack-16`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2079 (already merged into the integration branch); description reproduced below.

## Summary

Balance responses now carry the account they were looked up for and the request they answer. After switching accounts, a balance that was still loading for the previous account can no longer be stored as the new account's balance, and a response to an earlier request can no longer overwrite a later one.

- `WalletBalances` and the balance breakdown (`Balances`) tag every balance send with the account id and a per-request generation; the handler drops anything that no longer matches the selected account or the current request.
- Selecting a different account re-derives spendability immediately, before any balance for it arrives, so the previous account's "everything spendable" state does not linger on the new one.
- While the SDK still masks the spendable value, the breakdown shows the spendable row with a progress indicator and the "syncing" header copy instead of presenting the carried-forward number as final.
- The shared unmined-migration pending figure, which the breakdown subtracts from its Pending row, is now reset when the account changes and when a failed fetch clears rows that belonged to another account, so it cannot reduce the newly selected account's pending amount.

## Tests

New `WalletBalancesAccountProvenanceTests` (delayed previous-account balance rejected after a switch; the new account's own result becomes concrete; stale request generation dropped; selection change re-derives spendability), `RootAccountSwitchMigrationPendingResetTests`, a breakdown "updating while masked" case in `BalancesTests`, and a failure-path case in `RootTransactionsAccountProvenanceTests`. Existing balance suites adapted to the new action shapes.

Verified locally with the `zodl-internal` scheme: build succeeded; 70 tests in 9 suites passed (balance, breakdown, Root transactions and account-switch suites), plus the Send and Swap max-button suites.


LukasKorba (approved):
utACK

PR #2100, opened 2026-09-07
[MOB-1854] Replay a deferred retryStart and tag pipelines so stale completions cannot release a newer one (stack 18/21)

Chlup:
> **Stacked PR 18/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1862-balance-provenance-stack-17`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2080 (already merged into the integration branch); description reproduced below.

## Summary

The synchronizer restart pipeline (`retryStart`) had a latch so overlapping pipelines could not call `start` twice, but a request that arrived while a pipeline was in flight was dropped **and** its migration-resume flags were consumed. When the in-flight pipeline was a broadcast-only pass (which never starts sync), sync never resumed after the migration broadcast.

- Each pipeline now carries a generation. A request that arrives while one is in flight is deferred and replayed exactly once when that pipeline finishes; its resume flags stay armed until a pipeline is actually admitted.
- Completions are tagged with their generation, so a stale pipeline finishing late can no longer release a newer pipeline's latch or re-register the synchronizer streams. Entering the background bumps the generation and clears the deferred request; the foreground path re-registers on its own.
- Cold launch still registers the streams unconditionally.

## Tests

`RootRetryStartReentrancyTests`: a dropped request is replayed once when the pipeline finishes; migration-resume flags survive a dropped request; a gate-false edge during a broadcast-only pipeline resumes sync exactly once; a pipeline finishing after the app went to the background neither clears the newer latch nor re-registers. The existing cross-background test now also asserts that the late completion does not release the newer latch, and a further case checks that backgrounding clears a deferred replay so a stale completion replays nothing. The suite's stores are built on isolated in-memory shared storage.

Verified locally with the `zodl-internal` scheme: build succeeded; the 4 focused suites (37 tests) and all 71 migration and initialization suites (553 tests) passed.


LukasKorba (approved):
utACK

PR #2101, opened 2026-09-07
[MOB-1853] Let the terminal stall rebuild wait for the guard and cancel a benchmark still in flight (stack 19/21)

Chlup:
> **Stacked PR 19/19** of the sync-contention series (umbrella #2071). Base: `michal/MOB-1848/MOB-1854-retry-start-ownership-stack-18`. Review this PR's own diff; merge the stack bottom-up with a merge commit and delete each head branch after merging so GitHub retargets the next PR to `main`.
>
> Same change as #2081 (already merged into the integration branch); description reproduced below.

## Summary

Follow-ups from a whole-branch review of the sync-contention series.

- **No double server switch after the wallet gives up syncing.** The terminal stall rebuild now cancels a server benchmark that is still running from the earlier stall attempt, and ignores a benchmark result that arrives while the rebuild is in flight, so the app cannot tear the synchronizer down twice for one stall.
- **The rebuild waits for an in-progress send instead of being skipped.** `rebuildAfterStall` acquires the guard the way the manual server switch does (waiting for the broadcast to finish, bounded by the server-switch timeout) rather than skipping when the guard is busy, so a give-up that coincides with a send still gets its one rebuild.
- The transaction-list provenance marker is cleared together with foreign rows on a failed fetch, and the stall-rebuild counter's comment now states its background-only reset.
- Two Root account-switch suites build their stores on isolated in-memory shared storage, so they can no longer interleave on the shared migration pending value.
- CHANGELOG: one new `[MOB-1853]` bullet; the two `[MOB-1862]` balance bullets are merged into one.

## Tests

`RootTerminalStallRebuildTests` (benchmark cancelled on give-up; a candidate arriving mid-rebuild is dropped; a fresh candidate applies after the rebuild finishes), `AutoServerSelectionRebuildTests` (rebuild waits for a held guard and completes after release), `RootTransactionsAccountProvenanceTests` (provenance marker cleared with foreign rows), plus the touched Root, guard and Server Setup suites.

Verified locally with the `zodl-internal` scheme on iPhone 17: build succeeded; 77 tests in 11 suites passed.


LukasKorba (approved):
utACK

PR #2102, opened 2026-09-07
Share repository instructions between Claude and Codex

Chlup:
Project instructions currently live only in `CLAUDE.md`. Move the shared guide to root `AGENTS.md` so Codex discovers it, and keep `CLAUDE.md` as a thin native import so Claude reads the same instructions.

The existing guide is preserved in full, with only its heading and introduction made agent-neutral.

Validation: exact content comparison against the original guide, import-target verification, and `git diff --check` passed. Documentation only; no app build or tests were run.


LukasKorba (approved):
LGTM

PR #2104, opened 2026-09-07
[MOB-1869] Treat an undetermined balance as undetermined in Send and outgoing Swap and Pay (stack 20/21)

Chlup:
## Summary

Send and outgoing Swap and Pay decided "insufficient funds" and enabled Review using only the spendable-mask flag, while the home balance treats the balance as still being determined whenever the wallet is syncing without a concrete balance for the selected account (for example right after a restore). The two now use the same predicate, so the forms wait the way the home screen does instead of judging a positive amount against an initial zero.

- `SendForm` and `SwapAndPay` derive "spendability being determined" from the embedded wallet-balances state's full predicate (masked, or syncing with no concrete balance for the selected account).
- Swapping another asset into ZEC keeps its exemption: it does not spend the wallet's ZEC.

## Tests

`SendFormMaxButtonTests` and `SwapAndPayMaxButtonTests`: syncing with no concrete balance and an unmasked flag reads as undetermined (no insufficient-funds verdict, Review disabled); a concrete zero ends determination and the known-zero verdict applies; the incoming-swap exemption still holds.


LukasKorba (approved):
utACK

PR #2105, opened 2026-09-07
[MOB-1848] Pin the SDK at the quiescent-stop tip (stack 21/21)

Chlup:
## Summary

Moves `.sdk-pin` to the SDK integration tip that reports a non-quiescent engine stop and refuses wallet mutations on top of it (SDK stack 14). No app source changes.


LukasKorba (approved):
utACK

PR #2110, opened 2026-09-09
[MOB-1786] Give the seed-backup prompt the top banner slot

LukasKorba:
Implements [MOB-1786](https://linear.app/zodl/issue/MOB-1786/seed-backup-prompt-should-take-priority-over-migration-and-error-state).

Three users lost funds because they never backed up their recovery phrase. The backup prompt sat at rung 6 — the bottom half of the ladder — so whenever migration or an error state held the banner slot, the prompt was not merely outranked: the walk short-circuits at the first match, so it was **never asked**.

### Stacked on #2107

Based on `MOB-1889-under-300` ([#2107](https://github.com/zodl-inc/zodl-ios/pull/2107)), so this PR shows only its own commit. It retargets to `main` automatically once #2107 merges.

### What changed

Both orderings had to move together, because the ladder decides the *initial* seat by walk position and *displacement* by rank — changing one leaves the other wrong:

- The rung is asked **first**, immediately after the account guard it already depended on, as `.evaluatePriorityWalletBackup`.
- It seats at **rank -1**, above everything including a lost connection.

`rawValue` deliberately stays 6, so the `next()` walk-down helper and its existing tests are untouched. Rung 6 remains as a pass-through, keeping the rest of the chain and its tests intact.

### What did not change

The gates are exactly as they were. A brand-new wallet cannot spend, so any transaction at all is a receive — which is why `transactions.isEmpty` is the right "has received funds" test, and why spending back down to zero must not retract the prompt. `hasUserPassedPhraseBackupTest` lives in the keychain and ends the prompt for good once the user has backed up.

The user's own **"Remind me later" is still honoured**. This ticket reorders; it does not override a choice the user made. Reminder cadence belongs to PRO-276.

### How the migration banner actually reaches production — and why it matters here

`featureFlags.migration` is `false` in every mainnet flavour and gates only the ladder's **pull** rung (`.evaluatePriorityMigration`), the foreground check, and the `.upToDate` recheck. In production the migration banner arrives by **push**: `.onAppear` arms `migrationStateStreamEffect` (no flag), the manager's state/snapshot events funnel through `.migrationStateChanged` / `.migrationReevaluationRequested` into `bannerVariant` (no flag), and `.migrationVariantUpdated` seats `priorityMigration` (no flag). The only gate on that path is inside the manager — `isIronwoodActivated()` plus funds to migrate.

So a displacement test that re-runs the ladder is not sufficient on its own. This PR also proves the **hold** direction against the production path: with backup seated, a pushed `.required` variant still fires `triggerPriority(.priorityMigration)`, and the rank guard in `.openBannerRequest` (1.5 ≥ -1) is what rejects it — `aPushedMigrationVariantCannotDisplaceASeatedWalletBackup`.

### Review-round fix: the first-receive hook (rung 6)

Field-caught by Lukas while testing this PR: a new wallet that received 1 TAZ to its transparent address got the **shielding** banner, never backup. Cause: `RootTransactions` re-enters the ladder at `.evaluatePriority6` whenever the transactions array changes — that hook is how a *new* wallet gets the prompt at all, since at init-done the head rung walked past an empty history. The first cut of this PR made rung 6 a pass-through to rung 7, so the hook landed on the shielding rung instead.

Fixed by extracting the backup decision into `walletBackupClaimsSlot(state:)`, used by both the head rung (declines continue at rung 2) and rung 6 (declines continue at rung 7, exactly as before). Rank -1 wins wherever it is asked, so a seated shielding offer is displaced. Pinned by `theFirstReceiveReasksBackupAtRungSixAndDisplacesShielding`, the empty-slot variant, and a nothing-owed walk-on test; the existing rung-6 test in the residual suite is unaffected.

### This supersedes two approved rulings

Both were pinned by tests that failed on the rank change. They have been flipped with the new reasoning recorded in the file's own convention, not quietly adjusted:

- **MOB-1749's residual dust at 1.75** (approved 2026-08-25) no longer outranks backup. It keeps 1.75 and still outranks shielding, Tor and currency conversion — `residualDisplacesTheInformationalBannersBelowIt` keeps that coverage, now against currency conversion instead of backup.
- **MOB-1853's terminal-stall Retry at 0.75** now yields to an unbacked seed holding funds. During a terminal stall the Retry is displaced until the user backs up or dismisses.

Neither moves relative to the other; only backup moves. Flagging both explicitly for @michal-fousek, since they were deliberate calls.

### Testing

Full suite green with the 5 pre-existing known issues unchanged. The existing suite was run *before* any test was touched, so the code identified what the reorder broke rather than the expectations being adjusted to fit — exactly two tests failed, both predicted.

New `SmartBannerWalletBackupPriorityTests` covers rank against every other rung, `rawValue` staying put, the ladder asking backup first, displacement of a seated migration banner and a seated stalled banner, the pushed-variant hold above, and the gates that must still hold — no transactions, and no account holding the ladder before backup is asked.

### DIFF
--- a/secant/Sources/Features/SmartBanner/SmartBannerStore.swift
+++ b/secant/Sources/Features/SmartBanner/SmartBannerStore.swift
@@ -1104,10 +1104,18 @@ struct SmartBanner {
             case .evaluatePriority5:
                 return .send(.evaluatePriority6)
 
-                // MOB-1786: rung 6 is now a pass-through. The backup decision moved to
-                // `.evaluatePriorityWalletBackup` at the head of the walk; leaving the rung in
-                // place keeps the numbering and the rest of the chain (and its tests) intact.
+                // MOB-1786: the backup rung is asked at the head of the walk now, but it is
+                // ALSO still asked here. `RootTransactions` re-enters the ladder at this rung
+                // whenever the transactions array changes -- that is the hook that shows the
+                // prompt when a new wallet's first receive lands, at which point the head has
+                // already walked past an empty history. A pass-through here (the first cut of
+                // this change) rerouted that hook straight onto the shielding rung, so a
+                // transparent first receive seated the shielding offer and backup was never
+                // asked again. Field-caught by Lukas, 2026-09-10.
             case .evaluatePriority6:
+                if walletBackupClaimsSlot(state: &state) {
+                    return .send(.triggerPriority(.priority6))
+                }
                 return .send(.evaluatePriority7)
 
                 // wallet backup — MOB-1786: asked first, seats at rank -1.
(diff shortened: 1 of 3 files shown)


Chlup (approved):
utACK

PR #2115, opened 2026-09-10
[MOB-1863] Remove the awaiting-confirmation label from sent transactions

Chlup:
Linear: MOB-1863 (part of MOB-1848).

## What

Removes the app-side change that showed a sent, unmined transaction as "Sent · awaiting confirmation" once a server had accepted it. This is the exact inverse of 81b2d0c9:

- `TransactionState` loses the `isAcceptedBySubmitter` flag; a `.sending` row reads "Sending" until it is mined, as it did before.
- The live synchronizer mapping no longer asks the SDK for a transaction's submission status when it maps a `.sending` row.
- The two localized strings (en, es) and the three label tests are removed.
- The Unreleased changelog bullet for the label is removed; nothing about it ever shipped.

## Why

The label was not a requested change. The SDK's `transactionSubmissionStatus(for:)` API (MOB-1851) is untouched and simply unused by the app now.

## Test plan

- Build for testing succeeds (the generated string accessor for the removed entry disappears with it).
- `TransactionStateTests` is back to its pre-change set, including the test that a `.sending` row shows "Sending"; the Root transaction suites pass.
- Manual: with a wallet that has a just-sent transaction, the Activity row shows "Sending" until the transaction is mined; no "awaiting confirmation" text anywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ tooling, CI, tests, and internal refactors are deliberately not listed.
 ### Changed
 
 - [MOB-1858] Sending no longer waits behind unrelated network work. Preparing a payment now runs alongside other activity instead of blocking it, and if the wallet is still busy with something else by the time it's ready to broadcast, the transaction is shown as pending and submitted automatically in the background instead of being reported as a failure.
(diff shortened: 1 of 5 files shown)


LukasKorba (approved):
ACK

PR #2116, opened 2026-09-10
[MOB-1853] Remove the stalled-sync banner and its Retry lane

Chlup:
Linear: MOB-1853 (part of MOB-1848).

## What

Removes the "Sync has stalled" smart-banner lane, which was not an approved design change:

- `SmartBanner`: the `priorityStalled` lane (case, rank 0.75, walk-down special case, arbitration cases), the `isSyncStalledTerminally` and `isLatestSyncStatusError` state, the `evaluatePriorityStalled`, `syncStalledTerminally` and `retryStalledSyncTapped` actions, the stalled content view and help sheet, and the five `smartBanner.*.syncStalled.*` strings. Two side effects the lane had introduced are undone with it: the error banner's close on a syncing tick is the plain cleanup again, and the stop tick no longer closes it.
- `Root`: only the plumbing that fed the lane — the `isSyncStalledTerminally` flag, its mark/clear helpers, the `retryTerminalStallRebuild` action and its Coordinator routing, and the clear calls at background and on progress. The give-up and rebuild-finished arms are back to their earlier bodies.
- Tests: the lane-only suite is deleted; six Root tests about the flag, notifications and Retry are removed; the seed-backup priority tests drop their stalled entry. Two Unreleased changelog bullets about the banner are removed.

## What stays

The terminal-stall rebuild after the SDK gives up, bounded per foreground; automatic server switches only when sync is idle; the policy re-read inside the guard; the background cancellation of a rebuild; MOB-1786's seed-backup banner work; the error sheet's choose-another-server row. The SDK is untouched.

## Resulting behaviour

When the SDK's recovery gives up, the app still rebuilds sync within its budget. Once the budget is spent, or a rebuild starts nothing, Root logs and does nothing else; the SDK's own error state drives the existing sync-error banner and its help sheet, exactly as before the lane. There is no user-facing Retry any more.

## Test plan

- Build for testing succeeds; the SmartBanner suites, `RootTerminalStallRebuildTests`, `AutoServerSelectionRebuildTests` and `RootAutoServerIdleGateTests` pass; the whole `zodlTests` target passes.
- `grep` over `secant`, `zodlTests` and `CHANGELOG.md` finds no reference to the lane's identifiers or string keys.
- Manual: with a wallet whose sync stalls until the SDK gives up, no "Sync has stalled" banner appears; the ordinary sync-error banner and its help sheet do.


LukasKorba (approved):
utACK

Issue #2112, opened 2026-09-10
App start can sit on the "hi" screen with no feedback — harden the launch path


LukasKorba:
Field reports of Zodl "stuck on the hi splash" (a teammate on a TestFlight build; Lukas on 3.12.1 (1) testnet saw 6–8 s of "hi" before the Face ID bubble). Investigated 2026-09-10: **not a regression from the MOB-1848 merge** — every shipped build through 3.12.0 predates that stack, and a post-merge build launches fine on device. The launch path has several connected weaknesses that together turn a slow moment into an apparent hang. One fix, one PR.

### What the code does today (`RootInitialization.swift`)

- Before init completes the destination is `.welcome` — `WelcomeView` **is** the "hi" logo. No feedback, no animation, no watchdog. Any slow or hung step is pixel-identical to a stuck splash.
- `.home` is dispatched ~0.5 s after launch, but Face ID is requested only by `SplashManager`, which exists only inside `.overlayedWithSplash` on post-welcome destinations. So if the main thread can't paint HomeView, the Face ID prompt is delayed with it.
- The launch reducers do synchronous main-thread work before their `.run`s: `walletStorage.exportWallet()` **four times** (`areKeysPresent`, `initializeSDK`, `checkBackupPhraseValidation`, `resolveMetadataEncryptionKeys` — each a `SecItemCopyMatching` on a `WhenUnlockedThisDeviceOnly` item) and `mnemonic.toSeed()` twice. Instant on a simulator; on a phone the keychain can stall behind `securityd`, worst right after unlock/reboot.
- `initializationSuccessfullyDone` is gated on `sdkSynchronizer.start()` **returning**. A start that throws is handled; a start that hangs holds "hi" forever. `engine.start/stop` are untimed continuations around synchronous Rust calls.
- `SplashModifier` constructs a `SplashManager` on every body evaluation; its `init` has side effects (starts a 60 Hz `Timer` that retains the manager), `@StateObject` discards only the wrapper, and `finished()` never invalidates its own timer. Leaked timers keep ticking after Home.

### Fix (single PR)

1. Move the keychain reads and seed derivation off the main thread — into the `.run` effects that already exist — and read the wallet once, not four times.
2. Give `WelcomeView` life: run the splash animation there and request Face ID immediately, in parallel with SDK init, instead of only after `.home`.
3. Bound `start()` in the launch path so a hung SDK start surfaces through the existing failure/retry route rather than holding the welcome screen.
[...]

PR #2117, opened 2026-09-10
[MOB-1909] Start the splash once per manager and invalidate its timer when it finishes

LukasKorba:
Refs #2112 (MOB-1909). One of the issue's four items; the rest is deliberately not in this PR — see the end.

## What
`SplashModifier` builds a `SplashManager` on every body evaluation and hands it to `@StateObject`, which keeps only the first. The initializer did the work — prepare the wipe, request Face ID, start a 60 Hz timer — so every re-render of the root view while the splash was up produced an orphan that authenticated and ticked anyway, and `finished()` never invalidated the timer whose block retained the manager for the rest of the session.

Now `init` only stores; `start()`, called from `SplashView.onAppear`, does the work once per manager; `finished()` invalidates the timer and completes once; the authentication task is kept in the existing `task` property. An orphan manager is one allocation that does nothing. The 15-minute re-auth path is unchanged: a foreground that flips `splashAppeared` still gets a fresh view identity, hence a fresh manager and a fresh prompt.

## What the code says about the "stuck on hi" reports (corrections to the issue text)
- A hung `start()` does **not** hold "hi": `.home` is dispatched 0.5 s after the wallet is read, independent of the SDK effect. It leaves Home with dead sync — a different symptom.
- The only way to see "hi" for seconds is a blocked main thread before HomeView's first paint. Which step blocks is unmeasured: two observations, no telemetry.

## Not in this PR, on purpose
- A launch timeline in the log was built and dropped: `walletLogger` is installed only under `#if DEBUG`, so every Release build (production, internal, testnet) logs nothing and the timeline could never reach a field report without changing that policy.
- Keychain/seed derivation off the main thread, hi-screen liveness, early Face ID, a UI-acting start watchdog — all wait for a reproducible case or a policy decision on logging in internal builds.

## History
The branch was force-pushed once: the timeline commit and its review-round commit were dropped, and this commit was rebuilt on `main` without the four `launchTimeline` lines the earlier version carried in `SplashView.swift`. `git range-diff` against the earlier splash commit: `1: 7f285789 ! 1: db05f624` — same patch; the only differences are the removed `launchTimeline` context lines.

## Tests
`SplashManagerTests` (6 — inert `init`, `start()` authenticates once then spins, no auth when the flag is off, failed auth shows retry without spinning, `finished()` invalidates and completes once, hidden manager never starts). A red run against the old type is not possible — it has no `start()` — so the `init` side effects are documented in the diff rather than by a failing test.

Full `zodl-internal` suite on a dedicated simulator: `** TEST SUCCEEDED **` — `━ Test run with 1862 tests in 249 suites passed after 47.139 seconds with 5 known issues.` (the 5 are the pre-existing `withKnownIssue` tests).

### DIFF
--- a/secant/Sources/UIComponents/Overlays/SplashView.swift
+++ b/secant/Sources/UIComponents/Overlays/SplashView.swift
@@ -40,16 +40,22 @@ final class SplashManager: ObservableObject {
         self.isHidden = isHidden
         self.screenSize = UIScreen.main.bounds.size
         self.completion = completion
-        
-        if !isHidden {
-            preparePoints()
-            if featureFlags.appLaunchBiometric {
-                authenticate()
-            } else {
-                Task {
-                    self.spinTheWheel()
-                }
-            }
+    }
+
+    // MOB-1909: `SplashModifier` constructs a manager on every body evaluation and `@StateObject`
+    // keeps only the first, so anything `init` did ran once per re-render — an authentication
+    // request and a 60 Hz timer per orphan. The side effects now live here, run once per manager,
+    // from `SplashView.onAppear`; an orphan is one allocation that does nothing.
+    private var hasStarted = false
+
+    @MainActor func start() {
+        guard !hasStarted, !isHidden else { return }
+        hasStarted = true
+        preparePoints()
+        if featureFlags.appLaunchBiometric {
+            authenticate()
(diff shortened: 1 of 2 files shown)


Chlup (approved):
utACK

